PR #2, opened 2017-06-19
Adding tests for FEEL constants

etirelli:
I am proposing we use tests numbers starting with 0100 and up for our extensive FEEL tests. We can leave the lower numbers to use with more adhoc tests as we've been doing.

In any case, test numbers are only really used to make sure we have unique IDs for the tests. All the reporting is done by labels.


etirelli:
@agilepro first PR for the FEEL language, for review.

PR #12, opened 2017-07-27
Configurable name for result.csv

arthware:
* Result filename is configurable by overriding
'DmnTckVendorTestSuite.getResultFileName()'
* Defaults to result.csv
* Returning null disables result file creation 

Background: We would like to integrate the tck testsuite as part of our
unit tests and test our dmn engine with different settings against the
tck.  

For that, we would like to have result file name to be customizable (so
that multiple files can be written and placed into target folder) and
also a way to disable it.

* Updated drools to release version 7.1.0.Final so that artifacts can be fetched from maven central


etirelli:
Thank you.

Issue #8, opened 2017-06-30
Nested lists in 0010-concatenate.dmn


opatrascoiu:
1. Decision literalNestedList has type `List<List<String>>` but returns a literal that is not compatible with its type. See below.

```
              <literalExpression>
			<text>[["w","x"],"y","z"]</text>
              </literalExpression>
```

We should change the decision to return 
```
              <literalExpression>
			<text>[["w","x"],["y","z"]]</text>
	      </literalExpression>
```
and the tests accordingly.

2. Types of decision concatenate1 is `List<List<String>>`. However, the type of returned literal expression is `List<String>`. I suggest to change the type of the decision to List<String>

3. Type of decisions concatenate2, concatenate3 and concatenate4 is `List<List<String>>`. However, the type of the returned results are not compatible with this type.

I suggest to change the type to` List<String>` and change the literal expression to match the type, for example flatten the second parameter.

```
		<literalExpression>
			<text>concatenate(simpleList,flatten(nestedList))</text>
		</literalExpression>

```

etirelli:
Submitted PR #22

etirelli:
Closing this ticket as the original tests were simplified and a new test was added to check for [a]=a.

PR #22, opened 2017-07-29
Issue #8: fixing usage of nested lists

etirelli:


agilepro:
we agreed friday to merge this change, and then to add another test for irregular list nesting

PR #25, opened 2017-07-31
xsi:type attributes for ambiguously typed expected test result values

garyhallmark:
I added the attribute only for cases where the value cannot be easily guessed from its lexical form.


etirelli:
Gary, this PR is conflicting because it also includes your previous commit on the oracle test results. I tried to manually merge only the xsi:type commit, but it breaks the current runner. I need a few more hours to fix it before merging.

etirelli:
manually merged on commit https://github.com/dmn-tck/tck/commit/35b67d1568c814909d0c205e91b0620f41888a67

PR #26, opened 2017-08-01
Extended TestResult object with 'lines'

arthware:
'Lines' can now be added to TestResult, so that a more detailed result
can produced and printed, showing in detail why a test failed. For
example, one line for each failed comparison of expected and actual
result.


etirelli:
This seems fine, I just would like to run the tests locally before merging to check this does not break the runner. Will do ASAP.

PR #29, opened 2017-08-11
Fixed #28 Wrong InputNode name in test file: 0001-filter-test-01.xml

### LINKED ISSUE #28: Bug: Wrong InputNode name in test file: 0001-filter-test-01.xml
The inputNode in the test file is called "Employee" whereas in the model the InputData is called "Emplyoees".

I think this is the result of a mistake during merge conflict resolution.

arthware:


etirelli:
Regression was introduced here:

https://github.com/dmn-tck/tck/commit/b95dbd7c4ec2fc369205df229f27f04f1edec37a#diff-cbd12256bd219c20d326d253df4b151aL11

Approving and merging this PR.

PR #33, opened 2017-08-14
New testcases: user defined function definition and invocation

arthware:
New testcase for simple user defined function definition and invocation testing function defined as part of literal expression and another function using a boxed definition. 
Tests positional and named parameter invocation.

* Is there any practice we should follow regarding names?
* We introduced a new label: 'FEEL Functions: user defined' -> Feel free to change or adapt the labels we chose.

We are going to add another test file which adds more complex user defined function scenarios.


etirelli:
@arthware the PR looks good, I will merge it. Regarding the labels, we use "FEEL Function Literals" for the feel function definition syntax, and "Function Definition" for the boxed function definitions. I will change your labels accordingly.

etirelli:
Manually merged with this commit: 

https://github.com/dmn-tck/tck/commit/ff9a5187360eeb930b69ca8597e5c90659f87eb7

PR #34, opened 2017-08-14
New testcases: more complex user-defined function invocation

arthware:
New testcases testing different kinds of user-defined function definitions and invocation: 

* Defining functions in a "library" for re-usage in other decisions (invocation using dot operator).
* Different invocation variants using named, positional parameters
* More complex fn invocation using expressions in parameters 
* Invocation of BKM

* Is there any practice we should follow regarding names?
* We introduced a new label: 'FEEL Functions: user defined' -> Feel free to change or adapt the labels we chose.


etirelli:
> * Is there any practice we should follow regarding names?

The practice is to use a unique 4 digits number, followed by a short string that describes the test. The names you used look fine to me, just no need to "skip". I.e., instead of 0030 and 0031, you could have used 0022 and 0023, but no problem either way.

> * We introduced a new label: 'FEEL Functions: user defined' -> Feel free to change or adapt the labels we chose.

You can use "FEEL Function Literals" when declaring functions using the feel syntax, and "Function Definition" when using boxed functions.

etirelli:
Manually merged with the following commit: https://github.com/dmn-tck/tck/commit/3971bd715ceae68e76389f1fd3b6492b3a1cb6c9

Issue #35, opened 2017-08-15
Test case 0031: semantics of instanceof are not clear in the spec


etirelli:
Test case CL3/0031 defines an item definition with a number of fields, and then declares variables of that type that do not contain those fields.

The spec, AFAIK, is not clear on the semantics of "instance of". For instance, if an item definition tCoordinate declares fields x and y. A context with a single entry x is declared as a coordinate, is this valid? I.e., is an object with only field x, but no field y, still a tCoordinate?

Some engine implementations define the semantics of "instance of" using duck-typing, due to the underspecification of types in the spec, and on those implementations, the object defined above is not a tCoordinate.

There is an open ticket with the RTF for DMN 1.2 to clearly define the semantics of "instance of" and types in general in FEEL. 

Until the RTF decides on a course of action, I suggest we avoid tests that include such ambiguity. 

Test case CL3/0031 should either be refactored to not fall into this problem or be moved to non-compliant until DMN 1.2 is published and the test adapted accordingly.

arthware:
Hi Edson, thanks for pointing this out. As our type checks are not yet reliably implemented, this hasn't caught my attention. Our runtime is not strongly typed. I fixed this ambiguity and will create another pull request.

etirelli:
Fixed by PR #38

PR #36, opened 2017-08-15
DO NOT MERGE: Issue-35 : moving test CL3/0031 to non-compliant until it is refactored

etirelli:
Since today is the deadline for test case inclusion in the August report, we should move the test case to non-compliant folder, as per the discussion in Issue #35, until we have time to discuss it. We can refactor it and move it back to include it in the test cases for September.


arthware:
I wasn't sure on that either. Of course we can refactor the test, to remove this ambiguity. 
I'll create another pull request to fix this.

arthware:
I fixed the type ambiguity in PR #38. 
 
Therefore I suggest to merge PR #38 first (which fixes the root cause) and decide afterwards if the test case still should be moved to non-compliant temporarily (e.g. merge this PR).

etirelli:
I will close this PR in favour of #38 when I merge that.

PR #38, opened 2017-08-16
Fix issue #35 fix type ambiguity

arthware:
Fix for issue #35 in  TestCases/compliance-level-3/0031-user-defined-functions. 
**NOTE:** This pull request will most probably conflict with Pull Request https://github.com/dmn-tck/tck/pull/36 from Edson, when merged after his. Therefore I suggest to merge this one first (which fixes the root cause) and decide afterwards if the test case still should be moved to non-compliant temporarily. 

#36


TestCases/compliance-level-3/0031-user-defined-functions/0031-user-defined-functions.dmn
```diff
@@ -106,7 +108,7 @@
     </dmn:context>
   </dmn:decision>
   <dmn:decision id="_yHl3UIDhEee-MeWXoLgrYg" name="fn invocation named parameters">
-    <dmn:variable id="_eApsUYDjEee-MeWXoLgrYg" name="fn invocation named paremeters" typeRef="ns1:tFnInvocationResult"/>
+    <dmn:variable id="_eApsUYDjEee-MeWXoLgrYg" name="fn invocation named paremeters" typeRef="tFnInvocationNamedResult"/>
```
etirelli: Typeref is missing namespace prefix "ns1:"

etirelli:
@arthware Thank you for providing a fix. I will check the PR soon, but just eyeballing it, it looks good. Can you please just fix the namespace prefix as per my comment? Otherwise I will fix it myself when I check it.

TestCases/compliance-level-3/0031-user-defined-functions/0031-user-defined-functions.dmn
arthware: Sorry, for that mistake :-( Unfortunately I have no time to fix it now, because I leaving for vacation just right now :-) So feel free to fix that mistake. Thank you for checking!

PR #39, opened 2017-08-16
Split test cases of FEEL functions, constants and math (for better reporting)

falko:
Currently, we have four test cases that combine many tests for different FEEL functions, arithmetics, and constant literals. If an implementation has issues in these areas, the current report does not show any details and it is really hard to even know which decision or FEEL feature caused a problem.

In order to make it easier for users to see which FEEL features are supported, we suggest a small split in the test case XML files. Each expected result of a decision should be distinguished as a separate test case and therefore gets reported separately. This also helps developers from software vendors to identify and fix issues in their tools.

This proposal keeps changes to a minimum and does not require the DMN files or the Test XML files to be split or duplicated.

Have a look at the [updated Camunda CSV file](https://github.com/falko/dmn-tck/blob/a4070f17c28b8d4a060cf1c26a64a638222f23b6/TestResults/Camunda/7.7.0/tck_results.csv) to see the effect.

### DIFF
--- a/TestCases/compliance-level-2/0101-feel-constants/0101-feel-constants-test-01.xml
+++ b/TestCases/compliance-level-2/0101-feel-constants/0101-feel-constants-test-01.xml
@@ -14,26 +14,41 @@
 				<value>0.872</value>
 			</expected>
 		</resultNode>
+	</testCase>
+	<testCase id="002">
+	    <description>Tests FEEL number constants</description>
 	    <resultNode name="Decision2" type="decision">
 	        <expected>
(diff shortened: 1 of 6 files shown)


etirelli:
@falko I like the approach you propose here, the only issue is that for this month (August) we are in the test freeze period (some vendors already provided their results file with the current test structure). I will merge this on master and the changes will be included for September results and forward. If you would like to submit a new results file with the test structure for August, please do so as a separate PR and I will merge it into the results-2017_08 branch.

PR #49, opened 2017-09-11
Adding GS results for September 2017

opatrascoiu:


etirelli:
@opatrascoiu I will merge your result files, just keep in mind that tests can still change until Sep 15th. We freeze tests on the 15th of each month, and then publish the results submitted by each vendor on the last day of each month.

PR #50, opened 2017-09-11
Adding test for conditionals

etirelli:

PR #53, opened 2017-09-14
Testcases for FEEL decimal, floor and ceiling functions

arthware:
This Pull Request adds testcases for decimal, floor and ceiling functions. 

The files are _generated_ based on our JUnit testsuite.
As we are testing FEEL expressions here, the DMN model iteself is pretty boring: Just one decision containing a boxed context with various FEEL expressions for evaluation. 

If you guys are OK with this format, we are able to provide a LOT of testcases for FEEL expressions (built-in functions, filters, semantics, ...) in the upcoming weeks. 

For now, we skipped the error cases (e.g. using feel-null for parameters). We can re-generate the files  with error cases enabled, once we are on the same page on how to deal with that. 
We think this aspect is somewhat unclear in the spec (for DMN) and might be runtime specific. Currently, our runtime does skip evaluation in case of an error (fail early), instead of continuing the evaluation with null values. Continuing the evaluation would lead to a lot of subsequent errors, leading to painful error analysis.


etirelli:
@arthware I don't think there is anything wrong with the way you are generating these tests, but it does make long term maintenance a bit harder. 
What we have been doing is changing our tests to have more granular scenarios. 
In your example, there would be a model for instance with a decision with a literal expression (instead of the context) that uses the function (for instance, decimal(value, scale), where value and scale are input values).
Then, the test case file would have multiple tests, for different pairs of value and scale. You can have multiple decisions in that model, and have each test execute one (or more) of those decisions.

Here is an example: 
https://github.com/dmn-tck/tck/pull/50/files

It also makes it easier to add new individual tests in the future (for instance, by supplying "null" values as inputs, as you mentioned), because it would only add tests to the test case, but would not change the model itself.

Would you be able to change your test generator to generate tests this way instead?

arthware:
Thanks for the your feedback Edson!

We discussed your comment today and came to the conclusion, that it would be too much effort for us to generate the asked DMN model structure out of our FEEL expression test suite. 

Our FEEL runtime tests are designed to be self-contained. That is: we test isolated FEEL expressions and assert the expected result. 

Our tests are structured like below:
${feelExpression}, ${expectedResult}, ${optionalExpectedErrorMsgArray}
 
Example: 
> { "substring("abc",null,3)", "feel-null", FeelMessages.REQUIRED_PARAMETER_IS_NULL},
> { "substring("f",1)", "f" },
> { "substring("f",1,1)", "f"},
> { "substring("foobar",6)", "r") },
> { "substring("foobar",1,6),"foobar" }
> { "substring(string:"foobar",start position : 3),"obar" }

Trying to convert these self-contained expressions into the asked DMN model structure is a non-trivial task. Especially for expressions/functions with a variable amount of arguments and for functions that accept different types.

We would need to create Decisions for every combination of function invocation (taking into account named parameter, positional parameters, optional parameters)
This would require to extract the parameters from our test expressions, generate InputData nodes with correct type information, parse the value of the parameter and set it as input parameter for the testcase.

In our structure, adding a test case is a matter of adding another row in the boxed context, add the component to the result type and add the expected result. 
The model itself is trivial and easy to understand. We also tried to increase readability, by adding a description to the model file describing what is being tested and what the expected result is.

Don't get me wrong: The efforts to increase maintainablity for the TCK tests are great! But we think, that the structure is not ideal to test simple FEEL expressions, in order to get a deep test coverage (requiring to test a lot of different combinations, types and invocation schemes).

What we could do, is to generate a dedicated Decision for every FEEL expression under test. That would satisfy the goal of having only one expression per decision (instead of the boxed context), but ends up with quite some independent Decisions in the model. I don't see much benefit here. 

To sum it up: 
We will follow the proposed DMN model structure for more complex DMN tests. 
In order to benefit from our exsiting test suite for FEEL expressions (containing over 5500 tests), we need to stick to the format like provided. Otherweise the efforts would be too high on our side, to migrate and contribute  existing tests. 

What about this deal: 
We provide tests for missing functions (and or semantics) and see how maintainability turns out in the mid-term. If it turns out that the models are not ideal from a maintenance point of view, we still can refactor them. At least we would be able to quickly increase deeper test coverage.

etirelli:
@arthware Ok, understood, lets have a quick chat today on the call with Keith, but I am ok to proceed as you suggest. 
If it is possible for you to generate individual decisions in the model instead of a single decision with a context, as you mentioned, I think it would be good, nevertheless. It simplifies adding/removing individual cases in the future.

opatrascoiu:
You should also validate the schema for xml files. Don't think you can have both value and component in <expected> it's a choice.

opatrascoiu:
dmn files fail the schema validation as well - missing namespace attribute in <definitions> at the top

arthware:
@etirelli 
* Updated like in discussed in TCK meeting 15-Sep-2017
* Added reference(s) to chapters in specification for aspects under test as knowledge source
* Added simple Readme.md which is a simple view for the model, showing all decisions
  together with the literal FEEL expression. Rationale: No screenshots
  required. Everything is generated now. Increased maintainability and
  much less effort to generate and contribute further tests.

Glad to hear some feedback :)

opatrascoiu:
I tried all of them and they seem to fine. Great work !

I have a few comments:
- all the other test folders have some sort of naming convention: lower cases with - as separator. Will be great to be consistent and keep that.
- the dmn files still fail the schema validation - missing namespace for `<dmn:definition>`. Adding namespace="..." fixes the issue. Should not be too hard to fix.

Hope this helps

arthware:
@opatrascoiu @etirelli 
Updated PR:

* XML should now finally validate
* Changed names to follow exisiting rules (dash insteand of CamelCase)

opatrascoiu:
Tried them & schema validation. Everything seems to be fine :)

etirelli:
Thank you @arthware and @opatrascoiu. Merging this in, it will be included in the October test set.

PR #63, opened 2017-09-28
FEEL expression tests for string functions

arthware:
* substring
* string length
* upper case
* lower case
* substring before
* substring after
* replace


TestCases/compliance-level-3/1103-feel-substring-function-test/1103-feel-substring-function-test-test-01.xml
```diff
@@ -0,0 +1,89 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<!-- Contributed to DMN TCK by ACTICO GmbH https://actico.com -->
+<testCases xmlns="http://www.omg.org/spec/DMN/20160719/testcase" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="">
+    <modelName>1103-feel-substring-function-test.dmn</modelName>
+    <labels>
+        <label>Compliance Level 3</label>
+        <label>Data Type : String</label>
+        <label>Literal Expression</label>
+        <label>FEEL Functions: strings</label>
+        <label>FEEL Functions: string</label>
```
etirelli: duplicate/wrong label

TestCases/compliance-level-3/1104-feel-string-length-function-test/1104-feel-string-length-function-test-test-01.xml
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<!-- Contributed to DMN TCK by ACTICO GmbH https://actico.com -->
+<testCases xmlns="http://www.omg.org/spec/DMN/20160719/testcase" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="">
+    <modelName>1104-feel-string-length-function-test.dmn</modelName>
+    <labels>
+        <label>Compliance Level 3</label>
+        <label>Data Type : String</label>
+        <label>Literal Expression</label>
+        <label>FEEL Functions: strings</label>
+        <label>FEEL Functions: string</label>
```
etirelli: wrong label

TestCases/compliance-level-3/1107-feel-substring-before-function-test/1107-feel-substring-before-function-test-test-01.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<!-- Contributed to DMN TCK by ACTICO GmbH https://actico.com -->
+<testCases xmlns="http://www.omg.org/spec/DMN/20160719/testcase" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="">
+    <modelName>1107-feel-substring-before-function-test.dmn</modelName>
+    <labels>
+        <label>Compliance Level 3</label>
+        <label>Data Type : String</label>
+        <label>Literal Expression</label>
+        <label>FEEL Functions: strings</label>
+        <label>FEEL Functions: string</label>
```
etirelli: same

TestCases/compliance-level-3/1108-feel-substring-after-function-test/1108-feel-substring-after-function-test-test-01.xml
```diff
@@ -0,0 +1,68 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<!-- Contributed to DMN TCK by ACTICO GmbH https://actico.com -->
+<testCases xmlns="http://www.omg.org/spec/DMN/20160719/testcase" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="">
+    <modelName>1108-feel-substring-after-function-test.dmn</modelName>
+    <labels>
+        <label>Compliance Level 3</label>
+        <label>Data Type : String</label>
+        <label>Literal Expression</label>
+        <label>FEEL Functions: strings</label>
+        <label>FEEL Functions: string</label>
```
etirelli: same

etirelli:
@arthware great job! If you can just remove the duplicate labels, I think this is good to merge.

arthware:
@etirelli 
I've seen this labeling in another test file, that's why I used both labels. Wasn't sure which one was right: 
See here: 
https://github.com/dmn-tck/tck/blob/master/TestCases/compliance-level-3/0002-string-functions/0002-string-functions-test-01.xml

etirelli:
@arthware that is a mistake, will fix it, thank you for pointing it out. What I've been doing is following this table as the "source of truth" for labels:

https://dmn-tck.github.io/tck/glossary.html

arthware:
@etirelli Fixed the labels. Thanks for the glossary!

opatrascoiu:
Not sure about the following replace test

replace("","", "",null)

I think the expected result should be null. See https://www.w3.org/TR/xpath-functions/ flags sections.

I've tried it with net.sf.saxon.xpath and it throws exception. According to DMN standard we should return null.

Can you double check please?

Thanks

arthware:
@etirelli any thoughts on this? The spec does not specify the behavior on what to do in case of explicit **null for optional** parameters. 

We are behaving defensive and treat such cases as if the parameter was not provided at all. Therefore:
`replace("","", "",null) = replace("","","")` 

I could remove the testcase until a decision has been made.

agilepro:
It seems obvious to treat a null as if no parameter provided, and vice versa.  However, this really should be clarified in the spec itself.

etirelli:
My understanding is the same as @arthware . Setting an optional parameter to null is the same as omitting such parameter.

```replace("","", "",null) = replace("","","")```

My suggestion is to merge the PR and we have until Oct 15th to discuss and/or change anything for which we are not in agreement.

**Comment by arthware** [comment]
@etirelli I am good with that approach. We will backport any changes to the tests back to our test-suite, so that re-exporting the tests will conform to our agreements.

**Comment by etirelli** [comment]
@arthware @opatrascoiu I merged the PR into master as is. Lets investigate the test failures and we have until the 15th to either fix or remove the offending tests.

**Comment by opatrascoiu** [comment]
The DMN spec is very clear about the semantics of replace and match: same with the semantics in the XML standard (page 133):
```
	pattern, replacement, and flags SHALL conform to the syntax and constraints specified in clause 7.6 of XQuery 1.0
	and XPath 2.0 Functions and Operators. Note that where XPath specifies an error result, FEEL specifies a null result.
```
According to XML standard, section 7.6, the effect of calling of replace/match without flags is:	
```
	The effect of calling the first version of this function (omitting the argument $flags) is the same as the effect of calling the second version with the $flags argument set to a zero-length string. 
	Flags are defined in 7.6.1.1 Flags.

```
Hence, `replace(input, pattern, replacement)` is equivalent to `replace(input, pattern, replacement, "")` 
		
I think is safe to assume that flags = "" and flags = null are equivalent (see XML spec paragraph below).
- `All these functions provide an optional parameter, $flags, to set options for the interpretation of the regular expression. The parameter accepts a xs:string, in which individual letters are used to set options. The presence of a letter within the string indicates that the option is on; its absence indicates that the option is off. Letters may appear in any order and may be repeated. If there are characters present that are not defined here as flags, then an error is raised [err:FORX0001].
`

However, `replace("", "", ""), replace("", "", "", "") and replace("", "", "", null), ` should return null as `matches("", "")` returns true (see XML spec fragment below).
- `An error is raised [err:FORX0003] if the pattern matches a zero-length string, that is, if the expression fn:matches("", $pattern, $flags) returns true. 	
`

Similarly, the test `replace("abc", "[a-z]*", '#')` should return null because [a-z]* matches the empty string. 

I don't think we should add the tests until we agree on this. We need to be 100% sure that the tests are correct, otherwise is just extra work for all of us.

I've tested all the above on an existing Java implemention of the XML standard (net.sf.saxon:Saxon-HE:9.7.0-8) and all the above are confirmed.

etirelli:
@opatrascoiu ok, so your comment above brings up a different issue: the fact that the match parameter is matching a zero-length string. If that is the case, then I agree, we should respect the XPath semantics as per the spec. 

@arthware can you please take a look? otherwise, we can remove the offending tests until we fix them.

arthware:
@opatrascoiu @etirelli 
Thanks for your detailed feedback! I will update the tests and create another pull request. 

The `replace("abc", "[a-z]*", '#')` case was included by accident. Sorry about that. We came across the empty match string case too, but I forgot to remove it. Sorry about that. As this is an error case (and we don't have a proper way to test error scenarios yet) I will remove that specific testcase. 
I willl update the other tests according to your feedback.

Some general thoughts about the process of test contribution:

In future, I think it might be better to keep a Pull Request open, that contributes test cases, until all questions are discussed and a consensus is found.

Another option would be to add a "incubator" folder next to compliance-level-3 folder and add new and incubating testcases to that folder. That way it is easier to for everybody to test new contributions, give feedback. It is also clearer for all of us, what currently is being incubated and soon will become part of the offical test suite.
Once test-cases are spec-compliant, they are moved to the "offical" part of the test-suite, once a week (maybe after the TCK weekly calls)?

The only thing that must not happen is, that provided testcases will "rot" in the incubating folder. I would suggest to limit the max time test-cases are allows to spend in the incubator. Let's say to one month. 
After that time, tests should be either dropped or accepted.

What do you think?

PR #71, opened 2017-10-16
Issue-69: Fixing as discussed in the call on Oct 13th, 2017

etirelli:


arthware:
Reviewed the change. Looks good for us. We just need to add an exception in our parser, because "time" is a reserved keyword in DMN 1.1 ;-) As a consequence this test is impossible to implement fully spec-compliant to DMN 1.1 because of this contradiction.

As this is going to change in DMN 1.2, it is not a problem for us to support, but might affect other vendors.

etirelli:
As discussed in the call, I will change this PR to move the test to the non-compliant folder.

etirelli:
Re-discussed this subject in the call of Nov 10th. All participants agreed this is not a problem anymore, so decision was to merge this PR.

PR #79, opened 2017-10-26
Extended schema to provide marker for error results

arthware:
Extended schema to add errorResult flag to ResultNode to indicate null-value as a result of an error situations


arthware:
See #80 for an example on how the schema change is used.

PR #82, opened 2017-10-27
TestCases for FEEL time() function tests

arthware:
Adding testcases for FEEL time() function including error scenarios. 
Requires #79 to be merged first.

### DIFF
--- a/TestCases/compliance-level-3/1116-feel-time-function/1116-feel-time-function-test-01.xml
+++ b/TestCases/compliance-level-3/1116-feel-time-function/1116-feel-time-function-test-01.xml
@@ -122,448 +122,448 @@
     <testCase id="017">
         <resultNode name="feel-time-function_17" type="decision">
             <expected>
-                <value xmlns:xs="http://www.w3.org/2001/XMLSchema" xsi:type="xs:time">23:59:00</value>
+                <value xmlns:xs="http://www.w3.org/2001/XMLSchema" xsi:type="xs:time">01:02:03</value>
             </expected>
         </resultNode>
     </testCase>
     <testCase id="018">
         <resultNode name="feel-time-function_18" type="decision">
             <expected>
-                <value xmlns:xs="http://www.w3.org/2001/XMLSchema" xsi:type="xs:time">01:02:03</value>
+                <value xmlns:xs="http://www.w3.org/2001/XMLSchema" xsi:type="xs:time">00:00:00</value>
             </expected>
         </resultNode>
     </testCase>
     <testCase id="019">
         <resultNode name="feel-time-function_19" type="decision">
             <expected>
-                <value xmlns:xs="http://www.w3.org/2001/XMLSchema" xsi:type="xs:time">01:02:00</value>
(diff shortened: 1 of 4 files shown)


etirelli:
@arthware still reviewing the test cases and I will comment as I go. First one:

There is a bug in some examples of the spec (already fixed for DMN 1.2). Some of the examples show the time string with the leading "T", but the time lexical format does not support the leading "T":

"A time in the semantic domain is a value of the XML Schema time datatype."
https://www.w3.org/TR/xmlschema11-2/#time

Also, hour, minute and second are always mandatory for time strings. Only the offset is optional.

DanielThanner:
@etirelli 

The list before table 58 in the spec defines the lexical format of a time string. 
In this definitions there are two possibilites: 
- (A) time datatype as specified by XML Schema Part 2 or 
- (B) extended form of local time as specified by ISO 8601, followed by "@" character, followed by time zone identifier

(A) does not allow "T" as prefix
but for (B): Excerpt from the ISO spec: "In expressions of local time, applications may put the time designator [T] immediately in front of the representations".

This means that for the case where a time zone is specified, "T" can be used (example:  "T11:22:33@Europe/Paris"). 

Since the spec uses the "T" prefix also for lexical time strings in the examples for local times and offset times, we think it is a valid use case to use "T" as a prefix, for (A) too. It would be also very surprising to the user if he could use "T" only for (B).

The behavior was changed for DMN 1.2. But for 1.1 it seems to be valid to use "T" in the lexical string.

etirelli:
@DanielThanner thank you for the additional information, I missed the use of "T" in ISO 8601 literals. 

Nevertheless, the RTF considered the use of "T" as a prefix in FEEL a "typo" in the DMN 1.1 spec, not a "change" for DMN 1.2. I am not sure if you have access, but here is the link:

http://issues.omg.org/browse/DMN12-137

The resolution is: "remove leading 'T' from all time literal examples"

So, even if your interpretation makes sense, I believe it is counter productive to add tests (or support) for it. We can raise this point it in the TCK meeting this week if you prefer.

DanielThanner:
The ISO 8601 spec allows missing seconds in the lexical representation.

In the FEEL semantic domain the seconds are mandatory. Missing seconds are mapped to 00.

Therefore we think the provided test cases are valid.

opatrascoiu:
According to the DMN 1.1 standard (see chapter 10.3.2.3.4) the semantics of the time type is the same with the one defined in the XML standard. According to section D.2.2 Lexical Mappings in the XML spec the seconds are mandatory.

opatrascoiu:
The rules in the D.2.2. define the syntax for the time literals.

DanielThanner:
I have create a new OMG ticket to clarify this issue: [http://issues.omg.org/browse/INBOX-564](OMG-Ticket)

DanielThanner:
@etirelli Please review the pull request. Ambiguous test cases were removed.

etirelli:
@DanielThanner it looks fine now, although not all tests pass for me yet because I have to fix the runner. I will merge so that we can include the results in the November publication.

PR #95, opened 2017-11-23
Fix case 57 for 1117-feel-date-and-time-function-test-01

tarilabs:
1117-feel-date-and-time-function-test-01
for
case 57 : date and time("2012-12-24")

CURRENT expected: an error
SHOULD ACTUALLY expect: 2012-12-24T00:00:00 (date and time)

Should not be an error.

Accordingly to the spec `date and time` function can receive
just the date part,
FEEL spec printed page 131 " date time string – a string value
consisting of a date string value, as specified above, optionally
followed by the character "T" followed by a time string value
as specified above "

### DIFF
--- a/TestCases/compliance-level-3/1117-feel-date-and-time-function/1117-feel-date-and-time-function-test-01.xml
+++ b/TestCases/compliance-level-3/1117-feel-date-and-time-function/1117-feel-date-and-time-function-test-01.xml
@@ -400,9 +400,9 @@
         </resultNode>
     </testCase>
     <testCase id="057">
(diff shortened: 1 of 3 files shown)


etirelli:
@DanielThanner @arthware this is correct, the time part of the date time string is optional. 

I know these tests are generated by you, so I will merge this PR as is (due to the November deadline), but if you want to change and regenerate the test files and submit a PR, feel free to do so.

DanielThanner:
@etirelli Yes you are right. The test case is wrong. 

I can't see this commit on master or on the november branch. Something wrong here?

etirelli:
@DanielThanner I just merged this into master.

PR #107, opened 2017-12-12
Regeneration of ACTICO auto generated test cases (with test case removal, id changes, ...)

DanielThanner:
Changes:
- New format of test case id. Added first 10 digits of SHA-1 hashsum as test case id suffix. This ensures that test case id is unique, even if test cases were removed.
- Added typeRef information where possible to item definitions in DMN files.
- Added category information to Readme.md
- Removed date/time/date and time/duration testcases as committed in TCK conference call

Important:
- Vendors must supply new test results because testcase ids have changed.
- Testcase id contains now alphanumeric characters.


etirelli:
@DanielThanner I am ok if we want to add the hash to the test names to ensure uniqueness, but is it really necessary? Sorry I did not join the meeting last week, but if everyone is ok with renaming tests, I think we can merge this PR, *but* we need to make sure everyone understands they have to submit new results with the new test names, because their old results will not show up on the website...

DanielThanner:
Regardless of the new hash, if already published testcases are removed we have to regenerate all testcases and they will get new testcase ids. 

In order to prevent wrong references (using the testcase id) all vendors must submit new results. This is the case with or without the hash. The hash is only to ensure that we have no invalid references.

opatrascoiu:
Having new test names is fine by me. However, having a hash in the name (test and decision) makes the name not very friendly. Is it really necessary? We could probably extend the xsd if is a hash is really needed.

etirelli:
As discussed last week on the meeting, the renames were accepted. I also reviewed the tests and they look good. 
Merging.

PR #110, opened 2018-01-12
New Test: 0034 drg scopes

DanielThanner:
Tests information requirement subgraphs as mentioned in chapter 7.1 of the DMN 1.1. specification.


etirelli:
@DanielThanner , as far as I know, all dependencies must be explicitly represented by information requirements, knowledge requirements and authority requirements in the DRD.

For instance, in your model "Decision A 2.1" depends on the input "A", but it looks like you are resolving it transitively, through the dependency of "Decision A 1" that depends on input "A". 

My understanding is that this is invalid. If "Decision A 2.1" depends on input "A", it must explicitly have an information requirement to input "A".

Can you please clarify your reasoning to allow transitive requirements? Maybe I missed something in the spec?

agilepro:
We discussed int he meeting.  Transitive requests are not according to the spec.  This needs to be fixed.  Maybe discuss with Edson to be sure?

DanielThanner:
Chapter 7 in the specification describes "Relating Decision Logic to Decision Requirements". On page 60, line 14: "Each variable that is referenced by a decision’s expression must be a variable referenced by one of the decision’s information requirements **or an information requirement in the decision's requirement subgraph**." 

Glossary page 177: "The directed graph resulting from the *transitive closure* of the
requirements of a DRG element; i.e., the sub-graph of the DRG
representing all the decision-making required by a particular element."

We think that the information requirement subgraph of a decision puts all decisions and input datas of that subgraph in scope of a decision. Transitive access therefore is possible.

etirelli:
@DanielThanner Thank you, I missed that. I will give everyone else a couple days in case someone else has anything to add, and will merge the test after that.

etirelli:
As previously discussed, I am merging this case.

DanielThanner:
The discussion at the OMG issue shows that the transitive information requirements are not in scope. Therefore I am currently changing this test case, to NOT use transitive information requirements.

etirelli:
@DanielThanner ok, sorry, I missed the meeting yesterday, and was not aware this was discussed. Would you like me to revert the PR or are you sending another on top of this?

DanielThanner:
I will create a new pull request for the changes, thanks.

etirelli:
@DanielThanner Thank you. I just read the comments on the ticket.

PR #117, opened 2018-02-02
Runner C#

bzinchenko:
This is a sample test runner cloned from Trisotech and translated to C# (approximately). It can be used as a basis for other C# runners. First experimental version.


etirelli:
I will leave this PR open for a 2-3 business days in case we get some feedback from reviewers, then I will merge!
Thanks for contributing it, really helpful!

PR #121, opened 2018-02-14
fixed 0034-drg-scopes test

kurobako:


tarilabs:
The TCK dmn model file `0034-drg-scopes.dmn` suffers from 2 issues.

1. It was missing a "no-prefix"/default `xmlns` for addressing FEEL types without a namespace, this PR propose to solve the issue by introducing as per standard practices a
```
xmlns="http://www.actico.com/spec/DMN/0.1.0/0034-drg-scopes"
```
on the top main <definitions> xml-element

2. BKMs defined in the DMN model were missing a <variable> definition accordingly to DMN spec; this PR solve the issue by introducing the necessary <variable> definition per each of the BKM element.

etirelli:
@DanielThanner can you please review this PR? In particular, for issue 2 mentioned by @tarilabs above, the schema file states the <variable> declaration is optional, but the spec, on page 50 says:

> A BusinessKnowledgeModel element also contains an InformationItem that holds the FunctionDefinition, which allows a Decision to invoke it by name. The name of that InformationItem SHALL be the same as the name of the BusinessKnowledgeModel element.

So I believe, for all intents and purposes, the variable declaration must exist or the BKM would not be invocable.

DanielThanner:
I agree that there should be variable definition (information item) for the BKM. But it should not use the typeRef. attribute The typeRef attribute für BKM information items were actually removed from all TCK tests. So this pull request must be modified before merge.

The xmlns change is okay.

kurobako:
@DanielThanner @etirelli this PR is nwo updated with suggested changes. thanks @tarilabs !

PR #141, opened 2018-04-19
Add runner for jDMN

opatrascoiu:


tarilabs:
@opatrascoiu I hope to mention an helpful reminder,
you can add end-user instruction for running and reproducing the result,
by linking a document as a value for this property: https://github.com/dmn-tck/tck/blob/3c1e4340c5844eb373072a9dab9153908b1734ac/TestResults/GoldmanSachs/1.0.0/tck_results.properties#L8
for example the link value could be indeed: https://github.com/dmn-tck/tck/blob/master/runners/dmn-tck-runner-jdmn/README.md

/cc @agilepro

Issue #151, opened 2018-05-06
Tests with nested components in decision results


bzinchenko:
We have several examples with complex nested types in inputs. Are there similar examples where decision results are complex types?

agilepro:
Matteo has supplied a test with compound output values.  This has been reviewed.  Closing this issue since it will be resolved when the pull request is integrated to the source.

PR #155, opened 2018-05-25
Propose 0035 to test for nested components in decision results

tarilabs:
Proposal solution for https://github.com/dmn-tck/tck/issues/151

/cc @bzinchenko @etirelli


agilepro:
We are waiting for May results to be published.  After that we will merge the PR and people will have some time to review the results and make sure everything is running well.

etirelli:
@agilepro just FYI, we don't need to wait to merge PRs. We can add new tests at any time we wish, it is just that only tests submitted before the 15th of every month will be considered for the results of that given month. 

In any case, I just published the website for May/2018, and will merge this PR.

PR #169, opened 2018-07-31
Adding test case 0036 for a Decision Table with variables in the input

etirelli:


DanielThanner:

I think the testcase is compliant with the DMN 1.1 specification. It runs successfully with our engine.

etirelli:
@DanielThanner thank you for reviewing. Fixed the test file name.

PR #170, opened 2018-07-31
Submitting test case 0037 for decision tables on BKM with implicit parameters

etirelli:
This test case checks for compliance on Decision Tables inside BKMs, with implicit parameters, and uses expressions on the output of the decision table.


DanielThanner:

- The BKM defines a typeref for "encapsulatedLogic" and "variable" which is invalid. We already discussed this. Please remove the typeRef attributes for this elements.
- In the test xml file, "xsd:double" is used for the "children" component. In our testrunner we use the default Java Double.toString() which produces, for example "3.0". FEEL preserves the scale value. Therefore the output/expected results should be changed to: "Alice is a mother of 2.0 children." and "Bob is a dad of 3.0 children". As an alternative you can change the type in the xml file to "xsd:integer".

etirelli:
@DanielThanner thanks for the review.

* about the typeref, my mistake. I did not notice the tool is still setting the typeref on the encapsulatedLogic. Removed it.

* about the number, I changed the model instead to use the decimal() function to remove the decimal places. This way the result of the execution of the model is not dependent on the implementation of the runner.

Fixed. Let me know if you have any further comments.

DanielThanner:
@etirelli Sorry, but there is still an issue: typeref must also be removed from encapsulatedLogic element of BKM, not only variable element.

etirelli:
@DanielThanner sorry, I removed it from the encapsulatedLogic and moved that into the decisionTable element. Please take a look and let me know if that is correct now?

DanielThanner:
@etirelli 0037 looks good now

PR #171, opened 2018-07-31
Adding test case 0038 for decition tables in BKMs with explicit parameters

etirelli:
This test checks for compliance on decision tables embedded inside a context, inside a BKM with explicit parameters. It also uses a simple expression on the DT input expression, as well as on the DT output entry.


DanielThanner:

- The BKM defines a typeref for "encapsulatedLogic" and "variable" which is invalid. We already discussed this. Please remove the typeRef attributes for this elements.
- In the test xml file, "xsd:double" is used for the "children" component. In our testrunner we use the default Java Double.toString() which produces, for example "3.0". FEEL preserves the scale value. Therefore the output/expected results should be changed to: "Alice is a mother of 2.0 children." and "Bob is a dad of 3.0 children". As an alternative you can change the type in the xml file to "xsd:integer".

etirelli:
@DanielThanner thank you for the comments. I made similar fixes here as in PR #170 . Let me know please if it is ok now.

DanielThanner:
@etirelli works fine now

PR #173, opened 2018-07-31
Adding test case 0039 to add coverage for the semantics of lists in decision tables

etirelli:


DanielThanner:

- The syntax of the 2nd rule in the decision table is invalid. A list expression is not allowed for a unary test. As mentioned in the specification on page 136, only grammar rules 17.a, 17.b and 17.c are allowed. Therefore the square brackets must be removed (as already used in the 1st rule of the decision table).
- Qualified names are allowed as a unary test, as used in the 3rd rule in the decision table. But we think that the value of the variable behind the qualified name can only be "normal" FEEL value, not a unary test or a list of unary tests. If this would be possible, someone can change the behavior of a decision table at runtime. Depending on the value that is passed for the variable "Flu Symtoms", the decision table acts differently. No static analysis of decision tables are possible any more. The test should be changed. "Flu Symtoms" should be renamed to "Flu Symtom" and only a single string is allowed to be passed for that variable.

DanielThanner:
@etirelli Update my comment about test case 0039.

DanielThanner:
I recognized the following on page 74 in the DMN 1.1 specification: "An input expression value satisfies an input entry if the value is equal to the input entry, or belongs to the list of values
indicated by the input entry (e.g., a list or a range)." Seems that list values are also valid for decision table input entries.

etirelli:
@DanielThanner thank you for your comments.

* I renamed the test file.

* you are correct about the second rule. I think that was actually an oversight on the spec, but since this was not fixed in DMN 1.2 either, I removed the rule.

* as you noticed, list variables are allowed, although it was a bit "obscure" in DMN 1.1. Bruce Silver raised this in the RTF and the text in DMN 1.2 was changed to make it more clear.

etirelli:
Let me know if my latest fixes are ok now. Thanks!

DanielThanner:
@etirelli The expected result of test case "_07726176-a1a0-4e9e-9de2-16dba51556e2" with description "select 2" is wrong. It should be "fever is in the list of Flu symptoms"

DanielThanner:
@etirelli Additional issue with decision table with hit policy collect that does not match any decision rule. 

The specification says on page 137: "decision table may have no rule hit for a set of input values. In this case, the result is given by the default output value, or null if no default output value is specified."

Therefore the expected result of last test case should not be an empty list. It should be null.

etirelli:
@DanielThanner sorry about the dumb error on the "select 2" test. I am editing the files manually and it is quite error prone, but I think it is ok now.

Regarding the hit policy, it was a bug in the spec. It was clarified that C should return an empty list and C# should return 0 in case no rule matches in a DT. Only the other policies would return null, as C+, C< and C> make no sense when no rules are matched. 

I am good just removing the test if you prefer, or we can all agree it was a bug and fix the implementation. Up to you, we can discuss on Friday's call if you prefer.

BTW, just FYI, this was fixed to make the language consistent across different metaphors. So:

* a DT with C hit policy and no matches returns an empty list the same way that a filter with no matches returns an empty list (not null).

* the count() function was also fixed, as count([]) should return 0, the same way that a C# hit policy returns 0 when no matches happen.

DanielThanner:
@etirelli Our suggestion is to remove or deactivate the test case "select none" from file 0039-dt-list-semantics-test-01.xml. If we start testing the DMN 1.2 specification, the test case can be added / activated in the future.

etirelli:
@DanielThanner ok, will remove the test for now.

etirelli:
@DanielThanner let me know if you are ok to merge this one. Thanks.

DanielThanner:
@etirelli 0039 executes now without any errors in our DMN runtime. Can be merged.

Issue #176, opened 2018-08-01
Improve the Test Case Schema with extension possibility and a name attribute for TestCase


SimonRinguette:
I'm submitting a PR for a new schema for the Test Cases.

Two improvements:

**Extensibility**: Added the possibility for vendor to have extended attributes attached to test cases. We found that this would be pretty useful in our implementation.

**Name attribute on Test Case**: In our tool, we would like to have a Human Readable name for the test cases. It's optional in the schema.

PR #177, opened 2018-08-01
New Test Case Schema #176

SimonRinguette:
That support vendor extensions and a name (optional) attribute on TestCase


SimonRinguette:
I have updated the xjb bindings (was much more simpler that i originally tought),... this can now be merged.

PR #180, opened 2018-08-09
DMN test cases for nested boxed contexts

AggarwSh:


DanielThanner:
@AggarwSh Test cases works in our ACTICO DMN runtime without any errors. DMN 1.1 compliant.

PR #194, opened 2018-11-12
Update dmn-tck-runners-drools to newer Drools version

tarilabs:
... in preparation for DMN v1.2


tarilabs:
About f255d0f updates .properties file to reflect current Drools version, no change in result.

image

PR #191, opened 2018-11-09
First conversion of the Test Cases to DMN 1.2 for external validation

SimonRinguette:
First conversion of the test cases to DMN 1.2

I am not 100% confident in the perfection of the files. This is intended for larger testing.

PR #202, opened 2018-12-02
A few test DMN model still used spurious typeRef prefixes

tarilabs:
I believe initial DMNv1.2 migration missed to remove other spurious `tns:` prefixes, I failed to notice this when browsing the history of https://github.com/dmn-tck/tck/pull/191 originally

### DIFF
--- a/TestCases/compliance-level-2/0003-input-data-string-allowed-values/0003-input-data-string-allowed-values.dmn
+++ b/TestCases/compliance-level-2/0003-input-data-string-allowed-values/0003-input-data-string-allowed-values.dmn
@@ -16,7 +16,7 @@
         </literalExpression>
     </decision>
     <inputData name="Employment Status" id="i_EmploymentStatus">
-        <variable typeRef="tck:tEmploymentStatus" name="Employment Status"/>
+        <variable typeRef="tEmploymentStatus" name="Employment Status"/>
     </inputData>
     <dmndi:DMNDI>
         <dmndi:DMNDiagram id="_0003-input-data-string-allowed-values_D1">
--- a/TestCases/compliance-level-2/0008-LX-arithmetic/0008-LX-arithmetic.dmn
+++ b/TestCases/compliance-level-2/0008-LX-arithmetic/0008-LX-arithmetic.dmn
@@ -21,7 +21,7 @@
         </literalExpression>
     </decision>
     <inputData name="loan" id="_1f4ae444-2e4e-4d26-b1f7-87a645c3f50a">
-        <variable typeRef="tns:tLoan" name="loan"/>
+        <variable typeRef="tLoan" name="loan"/>
     </inputData>
     <dmndi:DMNDI>
         <dmndi:DMNDiagram id="_1fedf2c0-0f4a-470c-bc66-a15528e8a49a_D1">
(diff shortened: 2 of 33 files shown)


tarilabs:
@StrayAlien Greg, FYI maybe this PR actually addresses what you were commenting on here https://github.com/dmn-tck/tck/pull/191#issuecomment-443468826

If not the case, can you kindly clarify original comment and I'll try to deal with this PR or eventually a separate one

StrayAlien:
@tarilabs  Hi Matteo, thanks.  In part it does yes.  There are some other namespace prefixes in there as well.  Mostly 'ns1:' as far as I can see.  I count 513 typeRefs with prefixes in the TestCases folder.

Greg.

tarilabs:
@StrayAlien Greg I did count a bit less, anyway the ones I could detect I appended a couple of commits to fix them too. 

Basically I elided:
- `ns1:` (428)
- `tns:` (49)
- `tck:` (1)

Can you kindly check on your side, using the latest commit on this PR, and let me know if you agree this cover them all? Thanks for the feedback

StrayAlien:
@tarilabs Hi Matteo.  Thanks. Yep -  all gone now from test cases.  I can run green on this PR.  There are still some in the 'non-compliant' folder which is maybe why we counted differently.  Are they to be left as is?

Greg.

tarilabs:
> There are still some in the 'non-compliant' folder 

Noticed, but I would avoid a blind-fix given they were not supposed to be touched by the automated DMNv1.2 migration. I will make as part of this PR to revert any modification to those files. Anyway they need a more specialized "care" in order to be ported back into CL2/CL3 folder.

StrayAlien:
@tarilabs cool.  thanks.  I thought it worth mentioning (still learning here!).

falko:
This was not a blind fix. I manually checked those files and what was changed was necessary if these files want to ever be compliant.

tarilabs:
@falko 
> This was not a blind fix. I manually checked those files and what was changed was necessary if these files want to ever be compliant.

wrt "blind fix" I was referring to _mine_ modifications to avoid the `xx:` prefixes in typeRef which are indeed not compliant with DMNv1.2. I believe you can agree with me on that one, as I don't have a way to test those DMN model until we agree if they are compliant or not I can (personally) opt to not modify them.

Then, I noticed the commit originating from you also changed those DMN models in the "non-compliant" directory to change the namespace (only). I believe this is a necessary but non-sufficient change, for instance the models do still have errors, like the prefixes I mentioned. Hence I reverted that change with 
aba9e45

If you want me to preserve that change kindly let me know, I will re-instantiate it no problem!
Kindly let me know, thanks in advance

tarilabs:
@falko 
> That's why I invested some time to do the things that where obvious already. Let's not waste someone else's time to redo this again.

With 5649c12 I do revert any changes performed previously with aba9e45 .
As a result in this PR the lines of the namespace changes for "non-compliant" directory of the 2 DMN files are left _unchanged_ as following merge of https://github.com/dmn-tck/tck/pull/191 .

As I originally mentioned in https://github.com/dmn-tck/tck/pull/191#issuecomment-443433533 @falko thank you again for your contributions and I want to seize the chance to assure you is clearly not my intention to waste your time, nor anyone else time.

PR #204, opened 2018-12-09
some tests

StrayAlien:
All, not looking to get this merged in as yet, but as there is a fair bit in here I'd like to get feedback and see whether any assumptions I am making about the way stuff works are valid.

I have added new tests under TestCases/strayalien to keep them separate for now.  There are about 850 or so new assertions.  The tests are all very small in scope and test one thing only - no contexts or bkms involved.  Every test is a literal expression in a decision.  

* 0050-feel-abs-function
* 0051-feel-sqrt-function
* 0052-feel-exp-function
* 0053-feel-log-function
* 0054-feel-even-function
* 0055-feel-odd-function
* 0056-feel-modulo-function
* 0057-feel-context
* 0058-feel-number-function
* 0059-feel-all-function
* 0060-feel-any-function
* 0061-feel-median-function
* 0062-feel-mode-function
* 0063-feel-stddev-function
* 0064-feel-conjunction
* 0065-feel-disjunction
* 0066-feel-negation
* 0067-feel-split-function
* 0068-feel-equality
* 0069-feel-list
* 0070-feel-instance-of
* 0071-feel-between
* 0072-feel-in
* 0073-feel-comments
* 0074-feel-properties
* 0075-feel-exponent
* 0076-feel-external-java
* 0077-feel-nan
* 0078-feel-infinity
* 0079-feel-string-function
* 0080-feel-getvalue-function
* 0081-feel-getentries-function
* 0082-feel-coercion
* 0083-feel-unicode
* 0084-feel-for-loops

A number of these are the areas I identified with no coverage at all.   There is a lot of functions being covered, but I have also taken the opportunity to get the ball rolling on some more serious stuff as well.  The areas I suspect that if we don't get the tests right then 10 vendors will do it 10 ways. 

Namely, (for now), equality, instance of, and the 'in' syntactic sugar.  There are a lot of tests in these areas as each data type gets tested in different ways. Without that you don't get to learn that (say) `java.time.Period` does not implement `Comparable`.  Nice!!

My understanding has gaps - so a) some stuff might not be correct, and b) I cannot write some tests at the moment.  Like the time and datetime equivalency for "UTC form" - I have *NO* idea what that means in practice.  :-)

The 'in' tests effectively cover >, >=, =, <, <= for all comparable types as each 'range' check there has 5 assertions for each type (before/start-endpoint/middle/end-endpoint/after).  'in' also cover what it can of non comparable types as well.  Such as testing whether a list is in a list of lists, or a context is in a list of contexts.  They can't be used as ranges, but can still be the subject of an 'in' test. 

I have made some small assumptions here and there.  Like:

* The usage of '?' says for decision tables but I'm not sure if *only* there.  I have assumed usage everywhere and asserted upon it for each type.
* I don't believe the precedence of 'in' is not specified.  Making it same as 'compare' means the expression `x in ? > 10` parses as `'(x in ?) > 10` whereas I believe we want `x in (? > 10)`. So I have given 'in' the lowest precedence to cater for it - effectively meaning what comes after 'in' will always be the rhs of the 'in' expression.   
* for an 'in' where the RHS is a infix op that results in a boolean true/false then I have assumed this is the is the result of the unary test.  
* in a context, duplicate names overwrite the previous name.
* for list functions that operate on a given type, if any elements are not that type then it is treated as a param causing an error condition.  I think the same should be asserted for all other list functions as well.  
* the stddev algorithm is the 'sample' algorithm, not 'population'.  The spec example is for 'sample'.  The sample algorithm is not defined for a single value so passing in a single val to stddev is an error.
 
For 'instance of',  have assumed the item type of a list containing non-conforming items is 'Any'.  For a list with conforming items I have assumed to *most* specific type.  For example, if a list is `[{a: "foo"}, {a:"foo", b: "bar"}]` then its type is based on `{a:"foo", b: "bar"}`.  Spec has no guidance here. 

I'd like to get onto some other grey areas as well - like String() and external java functions.  I may end up writing a lot more lists tests as well.

Feedback and comments welcome.

Greg.

PS: I found this page handy for instructions on checking out a pull request locally:

https://blog.scottlowe.org/2015/09/04/checking-out-github-pull-requests-locally/


DanielThanner:
In the past we had an aggreement that the number range from 1000..2000 is reserved for auto generated test cases of our ACTICO DMN Runtime. Therefore it would be great if you can renumber your testcases. I think you can use range 2000..3000.

Also I think we should document the "reserved" number ranges on the TCK GitHub page.

StrayAlien:
Thanks Daniel, cheers.  I see now that the ones from 1000- are ACTICO.  Being new 'round these parts' I just started at the next number figuring that was the right thing to do.  No problem.  Also, knowing that those tests are auto-generated means I won't be putting in any PRs on them.  :-)  I was looking at bolstering them with some 'standard' unhappy paths tests like no params, too many params, badly named params, boundary conditions, wrong types, singleton array coercion and so on. I'll do that in a different numbered folder.

Thanks again. I'll renumber them in the coming days.

Greg.

TestCases/strayalien/0072-feel-in/0072-feel-in.dmn
```diff
@@ -0,0 +1,2012 @@
...
+        <variable name="number_011_a"/>
+        <literalExpression>
+            <text>1 in 2</text>
+        </literalExpression>
+    </decision>
+
+    <decision name="number_013" id="_number_013">
+        <variable name="number_013"/>
+        <literalExpression>
+            <text>10 in ? = 10</text>
```
tarilabs: This model contains a lot of FEEL expression allegedly (if I understand correctly) referencing a variable called `?`, which however is not in scope as none of these Decisions have an informationRequirement referencing a DRGElement named `?`. If my understanding is correct, this implies this DMN model invalid as containing semantic errors.

TestCases/strayalien/0058-feel-number-function/0058-feel-number-function.dmn
```diff
@@ -0,0 +1,158 @@
...
+    <decision name="decision001" id="_decision001">
+        <description>Tests FEEL expression: 'number("1.000.000,01", ".", ","))' and expects result: '1000000.01 (number)'</description>
+        <question>Result of FEEL expression 'number("1.000.000,01", ".", ","))'?</question>
+        <allowedAnswers>1000000.01 (number)</allowedAnswers>
+        <variable typeRef="number" name="decision001"/>
+        <literalExpression>
+            <text>number("1.000.000,01", ".", ",")</text>
+        </literalExpression>
+    </decision>
+    <decision name="decision003" id="_decision002">
```
tarilabs: I believe a typo.
I believe this is meant to be:
```
<decision name="decision002" id="_decision002">
```

TestCases/strayalien/0072-feel-in/0072-feel-in.dmn
StrayAlien: Thanks @tarilabs.  The '?' is specified in (v1.2) grammar rule 51.c (page 131) as a 'special variable' for use with 'in' tests.  The the expression is a 'general' expression  then it may access the being-tested-for value using '?' - so, as far as I understand, it doesn't have to be an info requirement.  It's a bit like the 'item' special var for iterating.  We're supposed to magically put it in there.

Though, I am not sure whether that syntax is applicable outside of DTs.  But, the tests assert that it is.  I thought best to do more than less.

TestCases/strayalien/0058-feel-number-function/0058-feel-number-function.dmn
StrayAlien: thanks.  There will be likely be more of those.  :-) That first PR was 3.5 days effort so it all has been done quite quickly! 

I'll go through them again looking for others.  Thanks again.  I appreciated the feedback.  :-)

StrayAlien:
* 0079-feel-string-function
* 0080-feel-getvalue-function
* 0081-feel-getentries-function

added.  Original post updated.

The 'get value' and 'get entries' are referenced as 'built in' functions in the spec but not actually described in the built in functions chapter.  However .. here is some coverage.

The string() func get some coverage too.  Hopefully, nothing too nasty in there.  Duration types get 'normalised'.  Additionally, and worthy of note is that any string output that puts things in quotes like a string literal in a map or a list gets it contents 'escaped' to ensure that any quotes inside that text are escaped.

Additionally, when outputting a context, if the key name has any of "{", "}", "\"", ":", "," in it it also gets wrapped in quotes and escaped.  Seems reasonable.  For the Java folks, I am using the apache commons text  escapeJava() method.

Feedback welcome.

Greg.

TestCases/strayalien/0072-feel-in/0072-feel-in.dmn
tarilabs: I don't believe we are on the same page and risk overlapping different things. I try summarize my understanding below in the hope this helps :)

First of all, my understanding is that `?` is a valid variable name by the spec, regadless of special use cases. 

When `?` is actually used in a Decision Table it might receive a special semantic according to this clause of the DMNv1.2 specification under chapter "8.3.3 Decision Rule metamodel" :

> d) One of the expressions in the inputEntry is a boolean expressions using the special ‘?’ variable and that expression evaluates to true when the inputExpression value is assigned to ‘?’

When you are saying:

>The '?' is specified in (v1.2) grammar rule 51.c (page 131)

I believe what you are referring to here, is Table 50. Table 50 is described in the spec as:

> Note that Grammar Rules (clause 10.3.1.2) are used in decision table condition cells. These decision table syntaxes are defined in Table 50.

I am understanding the spec here is making reference that Table 50 applies to decision table syntax only. 

In conclusion my understanding is:
- if you use `?` in a FEEL expression in a Decision Table inputEntry (cell) and that FEEL expression is a Predicate (returns a boolean type), the semantic of the DMN specification implies `?` value is the value of the relevant Decision Table inputExpression (column)
- if you use `?` in a FEEL expression outside of a Decision Table inputEntry (cell) this ought to be defined as a variable in scope, hence in a DMN model it must be an informationRequirement.

I am happy to be shown where my understanding is wrong! ;)

StrayAlien: Thanks @tarilabs.  Appreciated.  That makes sense.  I really wasn't too sure whether that applied outside of DTs (hence it is actually my first note above re assumptions)

> I have made some small assumptions here and there. Like:
* The usage of '?' says for decision tables but I'm not sure if only there. I have assumed usage everywhere and asserted upon it for each type.

I am happy to remove those from the tests - I do appreciate the feedback - I'm kind of working a little blind here sometimes .  

Do have have an opinion on whether the _rest_ of the 'in' syntax should be available outside a DT? Like 'between' and the rest of them.

Many thanks,

Greg.

TestCases/strayalien/0058-feel-number-function/0058-feel-number-function.dmn
StrayAlien: @tarilabs I've fixed that typo and done a run through all the rest to make sure they line up.  Happy to 'resolve' this now if you are.

TestCases/strayalien/0072-feel-in/0072-feel-in.dmn
tarilabs: You can indeed use grammar rule 51 outside Decision Table.
For instance you can write a FEEL expression as `1 in [1, 2, 3]` is a valid FEEL expression to be used as a DMN Literal expression. It would be a Predicate. It would evaluate to true.

The way I interpret clause of the DMNv1.2 specification under chapter "8.3.3 Decision Rule metamodel"  is something like ~ _"normally in a DT cell you have a FEEL unary test. You could use instead a FEEL expression as long as you have a `?` reference in it and the FEEL expression is a Predicate"_

I am happy to be shown where my understanding is/might be wrong! ;)

StrayAlien: @tarilabs okay - great.  Those tests took a lot of effort.  Happy to keep those in.  :-)  I'll comment out the '?' tests (pending any further input here) before removing them. 

Thanks again Matteo.  It sure is helpful to get some guidance.

tarilabs: Absolutely, please don't rely uniquely on my feedback, this is also an opportunity for a shared understanding of the spec, indeed a goal of the TCK.

Meanwhile it would be helpful you place/move these test either under the CL2 or CL3 directory (and checking for the numbering it takes the next available as commented by others earlier). I would opt at this point for CL3 if that is also the target of your implementation.

tarilabs:
Thanks @DanielThanner for your https://github.com/dmn-tck/tck/pull/204#issuecomment-445715352
I've reflected that here: https://github.com/dmn-tck/tck/wiki/How-to-Create-a-Simple-Test-Case#naming

TestCases/strayalien/0079-feel-string-function/0079-feel-string-function-test-01.xml
```diff
@@ -0,0 +1,346 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<!-- Contributed to DMN TCK by StrayAlien -->
+<testCases xmlns="http://www.omg.org/spec/DMN/20160719/testcase" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation=""  xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+    <modelName>0079-feel-string-function.dmn</modelName>
```
tarilabs: I believe this "0079-feel-string-function" test case will be very difficult to agree on.

- I don't recall if date-related normalization are imposed by the spec
- being FEEL space insensitive, I doubt how we could agree on a single/unique output of a FEEL expression like `string([1, 2, 3, "foo"])`; on pragmatic terms this test case would impose it to be `[1, 2, 3, "foo"]` only, but `[ 1 , 2 , 3 , "foo" ]` would actually be a valid output too

if my understanding is correct, my personal preference would be to steer away for imposing a single/unique value check of the string format representation.

tarilabs: I'm only trying to say that the test could consider a different strategy, to avoid checking for a single/unique output, when the specification allows for _several_ possible.
The test should steer a way to impose 1 and only value, when the specification allows for several. 

For example, the test could be rewritten with a more "complex" expression, still testing the `string()` function which is what you want, but something like `replace(string([1, 2, 3, "foo"]), "[\s\u00A0]+", "")` or ~~similar, so to avoid imposing the number of spaces, and checking just the part of the string which we know should indeed be part of the output regardless of the variants allowed by the spec! ;)

TestCases/strayalien/0068-feel-equality/0068-feel-equality.dmn
```diff
@@ -0,0 +1,748 @@
...
+        </literalExpression>
+    </decision>
+
+    <decision name="singleton_001" id="_singleton_001">
+        <description>Tests FEEL expression: '[100] = 100' and expects result: 'false (boolean)'</description>
+        <question>Result of FEEL expression '[100] = 100'?</question>
+        <allowedAnswers>false (boolean)</allowedAnswers>
+        <variable typeRef="boolean" name="singleton_001"/>
+        <literalExpression>
+            <text>[100] = 100</text>
```
tarilabs: The spec says:

> For convenience of notation, a singleton list L, when used in an expression where a list is not expected, behaves as if L[1] is written.

it follows on with filtering, function invocation examples, but is not exhaustive in other example like the one proposed here.

Just based on the former definition, one could argue the other operand is a number (`100`), hence a singleton list of a number (`[100]`) would fall in the former definition; hence by "_behaving as if L[1] is written_" would lead to this expression to be actually true, while the test case is asserting it to be false.

StrayAlien:
More tests:

* 0082-feel-coercion
* 0083-feel-unicode
* 0084-feel-for-loops

0004-lending: some typeRefs have been updated - they were incorrect.  Some typeRefs on literal expression forcing the wrong type like a `typeRef="number"` on a literal expression that returns a boolean. 

0012-list-function: relaxed the typeRefs here.  Some tests has incorrect types, and as this test is not testing types, and FEEL does not support lists must varied types nicely, I have removed the typed lists with simply 'list'  

For '0082-feel-coercion' I have made a start on tests for coercion for out-of-domain values to null as params and also as return types from decision, bkm, literal function, function call and also from inputs.  As some tests had bad types and everybody was passing I suspect this may cause a few breakages.  Hopefully not.  Also testsing for singleton array to val (and vice versa) 

for '0083-feel-unicode', I have made some small 'sanity' tests around support for escaped unicode like '\uffff' both in code and in literal strings, and also for 'supplementary' unicode as well.  This involves a few tests for string functions as (for example) "string length" should be counting codepoints, not chars. The substring function is also affected and needs to be codepoint aware.  

for '0084-feel-for-loops' I have closed off the coverage gaps for using numeric ascending/descending ranges as well as using expressions for end points and negative numbers.  I know there is already a 'for loop' suite there, but that stuff looks like it comes out of a modeller so I'm not sure if I am supposed to modify it or not.  Not wishing to offend anybody, so if they should go in the other suite then cool.  

All comments welcome.

TestCases/compliance-level-3/0054-feel-even-function/0054-feel-even-function.dmn
```diff
@@ -0,0 +1,158 @@
...
+        <allowedAnswers>false (boolean)</allowedAnswers>
+        <variable typeRef="boolean" name="decision002_a"/>
+        <literalExpression>
+            <text>even(-1)</text>
+        </literalExpression>
+    </decision>
+    <decision name="decision003" id="_decision003">
+        <description>Tests FEEL expression: 'even(0)' and expects result: '0 (number)'</description>
+        <question>Result of FEEL expression 'even(0)'?</question>
+        <allowedAnswers>0 (number)</allowedAnswers>
```
tarilabs: from ` <allowedAnswers>0 (number)</allowedAnswers>`
to ` <allowedAnswers>true (boolean)</allowedAnswers>`

TestCases/compliance-level-3/0054-feel-even-function/0054-feel-even-function.dmn
```diff
@@ -0,0 +1,158 @@
...
+        <allowedAnswers>2 (number)</allowedAnswers>
+        <variable typeRef="boolean" name="decision006"/>
+        <literalExpression>
+            <text>even(number:4)</text>
+        </literalExpression>
+    </decision>
+    <decision name="decision007" id="_decision007">
+        <description>Tests FEEL expression: 'even(n:4)' and expects result: 'null (number)'</description>
+        <question>Result of FEEL expression 'even(n:4)'?</question>
+        <allowedAnswers>2 (number)</allowedAnswers>
```
tarilabs: from ` <allowedAnswers>2 (number)</allowedAnswers>`
to ` <allowedAnswers>null (boolean)</allowedAnswers>`

TestCases/compliance-level-3/0055-feel-odd-function/0055-feel-odd-function.dmn
```diff
@@ -0,0 +1,158 @@
...
+        <allowedAnswers>true (boolean)</allowedAnswers>
+        <variable typeRef="boolean" name="decision002_a"/>
+        <literalExpression>
+            <text>odd(-1)</text>
+        </literalExpression>
+    </decision>
+    <decision name="decision003" id="_decision003">
+        <description>Tests FEEL expression: 'odd(0)' and expects result: '0 (number)'</description>
+        <question>Result of FEEL expression 'odd(0)'?</question>
+        <allowedAnswers>0 (number)</allowedAnswers>
```
tarilabs: from ` <allowedAnswers>0 (number)</allowedAnswers>`
to ` <allowedAnswers>false (boolean)</allowedAnswers>`

TestCases/compliance-level-3/0055-feel-odd-function/0055-feel-odd-function.dmn
```diff
@@ -0,0 +1,158 @@
...
+        <allowedAnswers>2 (number)</allowedAnswers>
+        <variable typeRef="boolean" name="decision006"/>
+        <literalExpression>
+            <text>odd(number:4)</text>
+        </literalExpression>
+    </decision>
+    <decision name="decision007" id="_decision007">
+        <description>Tests FEEL expression: 'odd(n:4)' and expects result: 'null (number)'</description>
+        <question>Result of FEEL expression 'odd(n:4)'?</question>
+        <allowedAnswers>2 (number)</allowedAnswers>
```
tarilabs: from ` <allowedAnswers>2 (number)</allowedAnswers>`
to ` <allowedAnswers>null (boolean)</allowedAnswers>`

TestCases/compliance-level-3/0050-feel-abs-function/0050-feel-abs-function.dmn
```diff
@@ -0,0 +1,140 @@
...
+        <description>Tests FEEL expression: 'abs(1)' and expects result: '1 (number)'</description>
+        <question>Result of FEEL expression 'abs(1)'?</question>
+        <allowedAnswers>1 (number)</allowedAnswers>
+        <variable typeRef="number" name="decision001"/>
+        <literalExpression>
+            <text>abs(1)</text>
+        </literalExpression>
+    </decision>
+    <decision name="decision002" id="_decision002">
+        <description>Tests FEEL expression: 'abs(-1)' and expects result: '-1 (number)'</description>
```
tarilabs: from `<description>Tests FEEL expression: 'abs(-1)' and expects result: '-1 (number)'</description>`
to `<description>Tests FEEL expression: 'abs(-1)' and expects result: '1 (number)'</description>`

StrayAlien:
Thanks @tarilabs.  Much appreciated. I saw Keith's request in the minutes to separate out the passing CI tests into a different PR.  Any clues as to where I find the CI results for this PR?  Got a link?

TestCases/compliance-level-3/0065-feel-disjunction/0065-feel-disjunction.dmn
```diff
@@ -0,0 +1,176 @@
...
+        <allowedAnswers>false (boolean)</allowedAnswers>
+        <variable typeRef="boolean" name="decision005"/>
+        <literalExpression>
+            <text>false or false</text>
+        </literalExpression>
+    </decision>
+    <decision name="decision006_a" id="_decision006_a">
+        <description>Tests FEEL expression: 'false or null' and expects result: 'false (boolean)'</description>
+        <question>Result of FEEL expression 'false or null'?</question>
+        <allowedAnswers>false (boolean)</allowedAnswers>
```
tarilabs: from `false (boolean)`
to `null (boolean)`

StrayAlien:
@tarilabs Thanks for review and also CI notes.  Apols for late reply - not 100% on this atm.  I'll get those updates made on the tests you've mentioned and in the coming days see how much is involved in breaking up the PR.  Thanks again.

TestCases/compliance-level-3/0012-list-functions/0012-list-functions.dmn
```diff
@@ -1,19 +1,10 @@
...
-        <typeRef>string</typeRef>
-    </itemDefinition>
-    <itemDefinition isCollection="true" name="tNumList" id="tNumList">
-        <typeRef>number</typeRef>
-    </itemDefinition>
-    <itemDefinition isCollection="true" name="tListOfLists" id="tListOfLists">
-        <typeRef>tStringList</typeRef>
-    </itemDefinition>
     <inputData name="list1" id="_86f5e710-a139-4cd5-8ac4-90b4671a75b8">
-        <variable typeRef="tStringList" name="list1"/>
```
tarilabs: > 0012-list-function: relaxed the typeRefs here. Some tests has incorrect types, and as this test is not testing types, and FEEL does not support lists must varied types nicely, I have removed the typed lists with simply 'list'

I have spent some time to review (manually) CL3 `0012-list-function` but to me all is good with concerning types to me, hence I don't understand why modifications to this test case are required.
If you still believe changes to `0012-list-function` test-case are required, can you kindly describe in a bit more details commenting your proposed changes, please?

StrayAlien: Hi @tarilabs, some more info over here: https://github.com/dmn-tck/tck/issues/205

The types, I believe, may have been okay in some interpretations of the v1.1 typing system but I think are invalid in v1.2 as the type `[a]` is not the same as type `a`.  If I recall, some results in those tests produced a list, but of mixed types (some string some string array) so therefore, the actual runtime-type type of the list is 'any'.  A list of any is not type compatible with a list of string.  

I think a 1.1 impl would have treated a single string and a list of string as the same type hence the tests were okay for 1.1 - but not for 1.2.

Does that help?

tarilabs: I will need to double check that, but then assuming the above, I would suggest the best action is not to elide types from the DMN altogether, but rewrite only those Decisions in the test case which actually exhibit the issue, e.g.: appendListItem, flatten1 (this latter as it depends on appendListItem).

TestCases/compliance-level-3/0062-feel-mode-function/0062-feel-mode-function-test-01.xml
```diff
@@ -0,0 +1,165 @@
...
+    <testCase id="002">
+        <description>will return median for odd list size</description>
+        <resultNode name="decision002" type="decision">
+            <expected>
+                <list>
+                    <item>
+                        <value xsi:type="xsd:decimal">6</value>
+                    </item>
+                    <item>
+                        <value xsi:type="xsd:decimal">1</value>
```
tarilabs: Spec says:

> Returns the mode of the list of numbers. If the result contains multiple elements, they are returned in ascending order.

Hence here it should be from 
```
            <expected>
                <list>
                    <item>
                        <value xsi:type="xsd:decimal">6</value>
                    </item>
                    <item>
                        <value xsi:type="xsd:decimal">1</value>
```

to
```
            <expected>
                <list>
                    <item>
                        <value xsi:type="xsd:decimal">1</value>
                    </item>
                    <item>
                        <value xsi:type="xsd:decimal">6</value>
```

TestCases/compliance-level-3/0063-feel-stddev-function/0063-feel-stddev-function-test-01.xml
```diff
@@ -0,0 +1,146 @@
...
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="008">
+        <description>will error when a single item.  Note, the example given in the spec can only be calculated using
+            'sample' stddev algorithm, not the 'population' algorithm.  For the 'sample' algorithm, the stdev result
+            of a single item is undefined as part of the calculation is a division by the (sample size - 1).
+            Which, with a sample size of 1, results in a divide by zero error.
+        </description>
```
tarilabs: Many thanks for pointing it out Greg!
I have raised an issue on the RTF to make this explicit, but I concur the example would lead to the sample standard deviation.

StrayAlien: :-) Thanks Matteo.

TestCases/compliance-level-3/0062-feel-mode-function/0062-feel-mode-function-test-01.xml
StrayAlien: Good catch @tarilabs.  Thank you.  I'll bolster the test to have 3 result items to lessen the change of getting it right by 'accident'.

TestCases/compliance-level-3/0012-list-functions/0012-list-functions.dmn
StrayAlien: Thanks @tarilabs.  My take was as the test assertions here are for the outcome of string functions - not type assertions - then the types have no real role to play here.  If there are cases that do need to be captured regarding type assertions then they should be covered explicitly elsewhere.

To that end, the same assertions are covered (hopefully!) in the `list_*` tests in `0070-feel-instance-of`.  And the coercion to null for mismatching types in `0082-feel-coercion` .

TestCases/compliance-level-3/0050-feel-abs-function/0050-feel-abs-function.dmn
```diff
@@ -0,0 +1,140 @@
...
+        <question>Result of FEEL expression 'abs(-1)'?</question>
+        <allowedAnswers>-1 (number)</allowedAnswers>
```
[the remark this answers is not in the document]
StrayAlien: amended

TestCases/compliance-level-3/0054-feel-even-function/0054-feel-even-function.dmn
StrayAlien: @tarilabs amended.

TestCases/compliance-level-3/0054-feel-even-function/0054-feel-even-function.dmn
StrayAlien: @tarilabs amended.

TestCases/compliance-level-3/0055-feel-odd-function/0055-feel-odd-function.dmn
StrayAlien: @tarilabs amended.

TestCases/compliance-level-3/0055-feel-odd-function/0055-feel-odd-function.dmn
StrayAlien: @tarilabs amended.

TestCases/compliance-level-3/0050-feel-abs-function/0050-feel-abs-function.dmn
StrayAlien: @tarilabs amended.

TestCases/compliance-level-3/0065-feel-disjunction/0065-feel-disjunction.dmn
StrayAlien: @tarilabs amended.

TestCases/compliance-level-3/0062-feel-mode-function/0062-feel-mode-function-test-01.xml
StrayAlien: @tarilabs amended.

StrayAlien:
@agilepro @tarilabs @etirelli 

There is a request in the weekly meeting notes for me to split out this test PR into those that pass on camunda/drools and those that don't into separate PRs.

May I ask that, like Matteo is doing, that the first step is for the respective engines to undertake some investigation and begin to fix tests and question what they feel is controversial or not-as-per-spec (and thanks again Matteo).  That may reduce the scope of PR wrangling/splitting down to much less than it is now.

Camunda is failing 100+ tests because it fails to parse valid DMN 1.2 XML so I think maybe the first bit of effort should be they look into that before I invest any time splitting those out.  Does anybody know if a representative from camunda has started doing as drools is doing? Or any other engines?

With drools, there seems to be a number of failures and runtime exceptions.  Some related to invoking some of the newly-tested-function like stddev.  Some related to string tests but just sanity stuff like what if the name of a context item has the "{" or ":" char in it, or that string context values are actually enclosed in quotes.  It looks also like some java external stuff is failing (on converting a single char FEEL string to a java char type?).   

As many of the new test exercise parts of the spec that were not yet covered by the TCK it is inevitable that things will break for some engines.  That doesn't mean the test is controversial just maybe not catered for so if other engines can contribute some time that'd be cool.

TestCases/compliance-level-3/0084-feel-for-loops/0084-feel-for-loops.dmn
```diff
@@ -0,0 +1,103 @@
...
+        <variable name="decision_002"/>
+        <literalExpression>
+            <text>for i in [1,2,3], j in [4,5] return i + j</text>
+        </literalExpression>
+    </decision>
+
+    <decision name="decision_003" id="_decision_003">
+        <variable name="decision_003"/>
+        <literalExpression>
+            <text>for i in [2..4] return i</text>
```
tarilabs: In the DMNv1.2 I read:
image

hence this is not a valid test, it should be rewritten as:
```
for i in 2..4 return i
```

at least this is my interpretation of the spec v1.2

TestCases/compliance-level-3/0084-feel-for-loops/0084-feel-for-loops.dmn
```diff
@@ -0,0 +1,103 @@
...
+        </literalExpression>
+    </decision>
+
+    <decision name="decision_003_a" id="_decision_003_a">
+        <variable name="decision_003_a"/>
+        <literalExpression>
+            <text>for i in [4..2] return i</text>
```
tarilabs: analogously, this should be rewritten as:
```
for i in 4..2 return i
```

TestCases/compliance-level-3/0081-feel-getentries-function/0081-feel-getentries-function.dmn
```diff
@@ -0,0 +1,96 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<definitions namespace="http://www.montera.com.au/spec/DMN/0081-feel-getentries-function" name="0081-feel-getentries-function" id="_i9fboPUUEeesLuP4RHs4vA" xmlns="http://www.omg.org/spec/DMN/20180521/MODEL/" xmlns:di="http://www.omg.org/spec/DMN/20180521/DI/" xmlns:dmndi="http://www.omg.org/spec/DMN/20180521/DMNDI/" xmlns:dc="http://www.omg.org/spec/DMN/20180521/DC/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+    <description>FEEL built-in function 'get entries(m)' in unspecified category</description>
+
+    <decision name="decision_001" id="_decision_001">
+        <description>Tests FEEL expression: 'get entries()' and expects result: 'null'</description>
+        <question>Result of FEEL expression 'get entries()'?</question>
+        <allowedAnswers>null</allowedAnswers>
+        <variable typeRef="list" name="decision_001"/>
```
tarilabs: The spec DMNv1.2 says:
image

then further specifically for InformationItem:
image

hence I derive `list` is not allowed to be used as a built-in type in typeRef (moreover IIRC there is currently undergoing discussion to make it supports generics for next DMN version)

either this DMN model needs to specify an ItemDefinition name `list` or would be best to drop the typeRef in this case(s)

TestCases/compliance-level-3/0081-feel-getentries-function/0081-feel-getentries-function.dmn
```diff
@@ -0,0 +1,96 @@
...
+        </literalExpression>
+    </decision>
+
+    <decision name="decision_005" id="_decision_005">
+        <description>Tests FEEL expression: 'get entries(m:{a: "foo", b: "bar"})' and expects result: '[{a: "foo}, {b: "bar"}]'</description>
+        <question>Result of FEEL expression 'get entries(m:{a: "foo", b: "bar"})'?</question>
+        <allowedAnswers>[{a: "foo}, {b: "bar"}]</allowedAnswers>
+        <variable typeRef="list" name="decision_005"/>
+        <literalExpression>
+            <text>get entries(m:{a: "foo", b: "bar"})</text>
```
tarilabs: IIRC there are 2 tickets to the RTF one by yourself and one by myself, to ask to add these two functions in the Tabular section where parameters are specified. I am a bit doubtful if we really want to test by named-parameter-invocation before those RTF issues get fixed in the spec, despite from the textual description I concur parameters name are `m`, `key`

TestCases/compliance-level-3/0081-feel-getentries-function/0081-feel-getentries-function-test-01.xml
```diff
@@ -0,0 +1,114 @@
...
+            <expected>
+                <list>
+                    <item>
+                        <component name="a">
+                            <value xsi:type="xsd:string">foo</value>
+                        </component>
+                    </item>
+                    <item>
+                        <component name="b">
+                            <value xsi:type="xsd:string">bar</value>
```
tarilabs: I disagree.
the DMNv1.2 spec says:
image

based on the example, the correct result of 
```
get entries({a: "foo", b: "bar"})
```

must be
```
[{value=foo, key=a}, {value=bar, key=b}]
```

In order to satisfy the example provided by the DMNv1.2 spec itself.

TestCases/compliance-level-3/0072-feel-in/0072-feel-in.dmn
```diff
@@ -0,0 +1,2178 @@
...
+        <literalExpression>
+            <text>[1,2,3] in ? != [1,2,3]</text>
+        </literalExpression>
+    </decision>
+-->
+
+    <decision name="list_014" id="_list_014">
+        <variable name="list_014"/>
+        <literalExpression>
+            <text>[1,2,3] in ([1,2,3,4], [1,2,3])</text>
```
tarilabs: Rule 51.d as tested for instance by number_014
We apply the element, a list `[1,2,3]`, to a list of positive unary tests { `=[1,2,3,4]` , `=[1,2,3]`} .
Results are`{false, true}`.
End result, `true` (as number_014 test).

TestCases/compliance-level-3/0072-feel-in/0072-feel-in.dmn
```diff
@@ -0,0 +1,2178 @@
...
+        <variable name="list_014_a"/>
+        <literalExpression>
+            <text>[1,2,3] in ([[1,2,3,4]], [[1,2,3,5]])</text>
+        </literalExpression>
+    </decision>
+
+    <decision name="list_014_b" id="_list_014_b">
+        <variable name="list_014_b"/>
+        <literalExpression>
+            <text>[1,2,3] in ([[1,2,3,4]], [[1,2,3]])</text>
```
tarilabs: Rule 51.d as tested for instance by number_014
We apply the element, a list `[1,2,3]`, to a list of positive unary tests { `=[[1,2,3,4]]` , `=[[1,2,3]]`} .
Results are `{false, false}`.
End result, `false`.

TestCases/compliance-level-3/0067-feel-split-function/0067-feel-split-function.dmn
```diff
@@ -0,0 +1,110 @@
...
+        </literalExpression>
+    </decision>
+
+    <decision name="decision005_a" id="_decision005_a">
+        <description>Tests FEEL expression: 'split("foo","f","f")' and expects result: 'null (list)'</description>
+        <question>Result of FEEL expression 'split("foo","f","f")'?</question>
+        <allowedAnswers>null (list)</allowedAnswers>
+        <variable typeRef="tStringList" name="decision005_a"/>
+        <literalExpression>
+            <text>split("foo","f","f")</text>
```
tarilabs: I think `split()` built-in function was discussed at some point, and at least in my perspective the DMNv1.2 Beta1 spec is missing to specify a flag parameter in order to make it actually usable. I need to check what is the issue# or eventually raise one. I would propose to drop this test case based on the fact it would prevent any vendor for implementing in such a strategy, at least until clarified.

TestCases/compliance-level-3/0084-feel-for-loops/0084-feel-for-loops.dmn
StrayAlien: Thank @tarilabs .  I do recall this.  I was pretty sure I put in tests for both syntaxes actually - kind of expecting the [] one to be a discussion point.  I sure meant to anyways.  I recall reading in the spec that it was without [] and thought that odd since ranges elsewhere exist inside [].  I agree that is what the spec says.  I'll amend.  Thanks.

StrayAlien: oh, I *did* put in both - `decision_007` and `decision_008` make the assertions without [].  I think it probably best to support both (I do) but, will remove those tests.  Thanks again.

TestCases/compliance-level-3/0084-feel-for-loops/0084-feel-for-loops.dmn
StrayAlien: as per above

TestCases/compliance-level-3/0081-feel-getentries-function/0081-feel-getentries-function.dmn
StrayAlien: Thanks @tarilabs.  I raised a few 'issue' tickets for the spec and then they didn;' seem to appear in the list so I thought I must have got it wrong somehow.  Glad they made it.  :-) 

What you say seems reasonable, I'll just comment out the named param assertions until things are concrete there.  Same for

StrayAlien: .. and same goes also for assertions `decision_007` and `decision_008` in `0080-feel-getvalue-function`

TestCases/compliance-level-3/0081-feel-getentries-function/0081-feel-getentries-function-test-01.xml
StrayAlien: Thanks @tarilabs.  Great catch.  You are correct.  I'll amend.

TestCases/compliance-level-3/0067-feel-split-function/0067-feel-split-function.dmn
StrayAlien: Thanks @tarilabs.  The spec has no flags on it but that doesn't mean it is unusable or unimplementable. It works without flags for many cases as the test show.  

Might I suggest we keep the tests - they verify what the 1.2 spec says.  If flags are added in a later spec the tests can be updated, but the test verify what can be used.

TestCases/compliance-level-3/0084-feel-for-loops/0084-feel-for-loops.dmn
StrayAlien: Actually.  I remember now exactly why I put these in.  Despite what the spec says in that particular place, there is the 'syntactic sugar' section that gives addition usages on 'in'.  The [x..y] notation is one of those.

Before we lose those tests, perhaps we ought to open this question out to @agilepro and @etirelli and anyone else paying attention.  My interpretation is that it is likely valid syntax.

TestCases/compliance-level-3/0072-feel-in/0072-feel-in.dmn
StrayAlien: Thanks @tarilabs. This test and the one you mentioned below are there specifically to catch an ambiguity - or at least what I feel is one.  That ambiguity is when the target of the unary tests is a list and a list appears as a test in a unary tests.  It is an equality match?  Or, as if we were (say) testing `1` in `([1,2,3])`?  We do not compare `1` against `[1,2,3]` - we compare it against each element of the list.    

So, the test shows that a list like `[1,2,3]` tested against `([1,2,3])` is comparing `[1,2,3]` against `1`, then `2` then `3`.  None of which match.

Where this become a bit strange is that the 'syntactic sugar' section mentions has a number of different interpretations of `in`.  Maybe, when lists are the target of a unary test we should be testing each element (as shown in prev paragraph) and *also* an equality check.  Like ... run down through the syntactic sugar stuff and evaluate relevant ones.

The spec doesn't give any guidance here as far as I can tell, so I threw in a test to see if my assumption was correct. 

Does that make any sense?

TestCases/compliance-level-3/0084-feel-for-loops/0084-feel-for-loops.dmn
etirelli: @StrayAlien the 'for' loop syntax does not support ranges with [ ] . I think the reason is because in the case of 'for' it wouldn't make sense to use exclusive ranges, only inclusive. As usual, we can open a ticket to the RTF to add support for the syntax, but it is not in DMN 1.2.

StrayAlien:
@tarilabs @etirelli @agilepro - re splitting PRs etc.  Creating a single PR from the reviewed and okay'd suites sounds like a good idea.  I'll do that.  

I could also break up the other suites into a PR each.  Some work involved for me, but might be better to track comments and so on and get things moving into master in a more timely fashion.  Certainly, that is preferable to hunting through CI logs for other people's systems.

The 'controversial' suites like string and unicode can then then so merged .. or not ..

How does that sound?

agilepro:
That would be great if you can pull the tests which Matteo has identified as non controversial into a pull request.  I could approve that quickly and get them into the set of tests.

Then separate PRs grouped by issue so we can discuss and incorporate after we have agreement.  

I don't know if you know this, but we generally incorporate new tests in the first half of the month, and then give the second half of the month for companies to get their results in.  Given that the repository is versioned that does not mean you have to wait, but only that the results for this month will not include any new tests...

StrayAlien:
Thanks @agilepro,

> separate PRs grouped by issue

By suite?

TestCases/compliance-level-3/0067-feel-split-function/0067-feel-split-function.dmn
etirelli: @StrayAlien in this case, the missing flag on split() was actually my fault (yeah, this is one of the mistakes I mentioned). Without the support for flags it is not possible to split multi-line strings for instance, or ignore case, etc. It was an oversight that needs to be fixed on DMN 1.3. 
See functions matches() and replace() that have the optional "flags" parameter in place. 

In regards to the TCK, we can add a test that effectively prevents vendors from implementing the optional "flags" parameter (what you are proposing here) or we can omit such test and let vendors that want to offer such feature in advance of DMN 1.3 to do so. 

Our approach in the past in cases like this was to add positive tests for such "upcoming" features in the non-compliant folder and wait for the next DMN version to move them to the compliant one. For example, we did the same for functions any() and all() that were known to come in DMN 1.2 to replace the previously named functions or() and and(). Vendors that wished to do so, could implement the feature in advance, without penalty or prejudice.

The TCK is community driven, so we can always decide to change our minds and do different from now on. Being obviously biased (I want to enable users to do what they need), I usually defer the decision to the group (/cc @agilepro) and will abide by such.

TestCases/compliance-level-3/0072-feel-in/0072-feel-in.dmn
etirelli: @StrayAlien understood, and what @tarilabs described is the intent of the spec. The following expression is not ambiguous, because the comma in "( <value> ,  <value> )" disambiguates it:

> [1,2,3] in ([1,2,3,4], [1,2,3])

The result is described by @tarilabs .

Having said that, you raise a good point I didn't think before. The following *IS* ambiguous:

> [1, 2, 3] in ([1, 2, 3])

The ambiguity resolution rule of the language would resolve the above as:

> [1, 2, 3] in <expression> 

Where <expression> is the list [1, 2, 3]. So your conclusion is correct and the result will be false. This is quite confusing to users actually, and deserves a ticket to the RTF.

StrayAlien: May I ask you to expand further on:

> The following expression is not ambiguous, because the comma in "( <value> , <value> )" disambiguates it.

And 

> the ambiguity resolution rule

?

The only references to ambiguity I know of in the spec relate to names, not unary tests.

Regarding this example would `1 in ([1,2,3,4], [1,2,3])` test for 'in' and not '=' against the lists?  I understood when a list is an element of a unary test we are testing for inclusion in this case. Or both?

StrayAlien: > The following expression is not ambiguous, because the comma in "( , )" disambiguates it.

Thinking about it, I think you are saying it defines it as a unaryTests and not an 'expression' .. got it.  

But the 'test for inclusion vs equality' thing I'm still not 100% sure on (hence the test).

TestCases/compliance-level-3/0067-feel-split-function/0067-feel-split-function.dmn
StrayAlien: Thanks @etirelli, none of us are perfect - just look at some of my spec interpretations! :-)

In fact, to further prove it - I have only really just understood what @tarilabs  really meant.  Doh!  The test is _preventing_ 3 params.  Sorry - I just totally missed that point.  I was, after all, doing this during my lunch while stuffing my face.  :-)

The 'too many params' assertion is just a standard thing I did for all funcs I tested.  I think you are right, if it is coming in a future spec then I think it easiest just to remove that assertion for this function.

No problems.  Apologies to you and @tarilabs for not quite getting it sooner!

TestCases/compliance-level-3/0068-feel-equality/0068-feel-equality.dmn
```diff
@@ -0,0 +1,746 @@
...
+        </literalExpression>
+    </decision>
+
+    <decision name="date_003" id="_date_003">
+        <description>Tests FEEL expression: 'date("2018-12-07") = date and time("2018-12-08")' and expects result: 'false (boolean)'</description>
+        <question>Result of FEEL expression 'date("2018-12-07") = date and time("2018-12-08")'?</question>
+        <allowedAnswers>false (boolean)</allowedAnswers>
+        <variable typeRef="boolean" name="date_003"/>
+        <literalExpression>
+            <text>date("2018-12-07") = date and time("2018-12-08")</text>
```
tarilabs: This test case `date_003` but also `date_004`, `ym_duration_006`, and others which I may have missed are testing edgecase/errorcase/unspecified from the spec:
image

In my perspective these are beyond scope and better left out of testing.

opatrascoiu:
Regarding some of your tests, for example `0075-feel-exponent`. 

I think you should add attribute `errorResult="true"` for <resultNode>s that contain ill-formed  constructions (e.g. "foo" ** 4).

StrayAlien:
> Regarding some of your tests, for example `0075-feel-exponent`.
> 
> I think you should add attribute `errorResult="true"` for `<resultNode>`s that contain ill-formed constructions (e.g. "foo" ** 4).

Thanks @opatrascoiu .  Note and done.  Though the notion of an error here is a little lose.  As far as I can tell, the spec says we should log and error but the result is null - so I don't actually think we should be asserting for an error at all as how we log is outside the scope of the spec.  But, for now to keep it consistent, no probs.

StrayAlien:
> Regarding `0070-feel-instance-of` tests.
> 
> DMN 1.2 is not very clear about type expressions for complex types (e.g. list, context or function).
> 
> There is an approved proposal for that - DMN13-37.
> 
> ```
> 54. type = 
>        qualified name |
>        'list' '<' type '>' |
>        'context'  '<'  name ':' type { ',' name ':' type }  '>' | 
>        'function' '<' [ type { ', ' type } ] '>'  '->'  type
>         ;
> ```
> I think we should remove the instance of tests for the complex types for now.

Is there something about the tests which is ad odds with the new proposal.  I would have thought that means more and better tests later on - but are the ones here incorrect?

TestCases/compliance-level-3/0072-feel-in/0072-feel-in.dmn
StrayAlien: @tarilabs. test updated as such.

TestCases/compliance-level-3/0072-feel-in/0072-feel-in.dmn
StrayAlien: @tarilabs. test updated as such.

TestCases/compliance-level-3/0081-feel-getentries-function/0081-feel-getentries-function-test-01.xml
StrayAlien: @tarilabs. test updated as such.

TestCases/compliance-level-3/0081-feel-getentries-function/0081-feel-getentries-function.dmn
StrayAlien: @tarilabs. test updated as such.

StrayAlien:
All, I'll not be making an more changes on this branch.  I am breaking it up at the moment.

StrayAlien:
The suites so far given the green light are in : https://github.com/dmn-tck/tck/pull/217

All other suites are in their own PR.  Al changes done here are reflected in those other PRs now.

tarilabs:
Thank you Greg

StrayAlien:
@tarilabs - did you see the branch name?  :-)

StrayAlien:
closing PR.

PR #209, opened 2018-12-17
Drools DMN TCK runner improvements

tarilabs:
When a DMNType is not available to unmarshal TCK test expected value,
use inference heuristics based on JAXB value type.


agilepro:
OK, great!

PR #218, opened 2019-01-28
0057-feel-context

StrayAlien:


agilepro:
Got these errors in RedHat

Tests in error: 
  006(0057-feel-context-test-01): ERROR: FAILURE: 'decision006' expected='{foo+bar((!!],foo=foo}' but found='{"foo+bar((!!],foo"=foo}'
  007(0057-feel-context-test-01): ERROR: FAILURE: 'decision007' expected='{=foo}' but found='{""=foo}'

Should it be:   {foo+bar((!!],foo=foo}
or this:         {"foo+bar((!!],foo"=foo}

tarilabs:
As mentioned in the TCK meeting 2019-02-01.

This is fixed in the latest Drools version; however I question the pragmatics of what is in my view an edge-case of the DMN spec implementation for `{"": "foo"}`.

1. there is no way to access that key-value-pair until `get entries` and `get value` functions are properly addressed in the OMG RTF tickets already raised.
2. I question **imho** the pragmatic of that, being "friendly enough" and allow names to contain spaces and funny characters, when for context entries those cases not supported as native-FEEL-name can go between quotes; and further can only be extracted with just the functions mentioned in point 1 above..

StrayAlien:
@tarilabs @agilepro.  Tests are for edge cases.  :-)  In this case, yes, it would be very difficult to use a context property that has no name.  However, the spec doesn't disallow it, incoming JSON data may have it, and a test like this is really to help ensure we all tackle edges cases in the same way.  With a test like this, for sample, a vendor might feel it was an error and disallow it.
  
Whether it _should_ be allowed is a different case I believe.

The other tests for weirdo chars in names is kind of the same in spirit.  I doubt people will actually do it, but we should all handle it the same.  Man, if they did do it though they need a talking to .... :-)

opatrascoiu:
I checked the tests - they look fine to me (comply with DMN 1.2).

Don't get me started on FEEL names ... We need to fix them in DMN 2.0.

tarilabs:
I'll try to be more explicit on my concern which I tried to originally summarize in https://github.com/dmn-tck/tck/pull/218#issuecomment-459810332, point 1 :) 
...
Until https://issues.omg.org/browse/DMN13-126 is fully addressed,
 a FEEL context like `{"": "foo"}`
is in my perspective **not** usable on DMNv1.2


xunit0integration

StrayAlien:
@tarilabs.  Hah, oh man, that is FUNNY!  :-) However, sorry, I cant read an omg issues - seems I'm locked out or something: https://github.com/dmn-tck/tck/issues/245

PR #224, opened 2019-01-28
0068-feel-equality

StrayAlien:


StrayAlien:
@tarilabs "fair cop" as the Brits say.  Requested changes pushed.

TestCases/compliance-level-3/0068-feel-equality/0068-feel-equality.dmn
```diff
@@ -0,0 +1,727 @@
...
+        </literalExpression>
+    </decision>
+
+    <decision name="singleton_001" id="_singleton_001">
+        <description>Tests FEEL expression: '[100] = 100' and expects result: 'false (boolean)'</description>
+        <question>Result of FEEL expression '[100] = 100'?</question>
+        <allowedAnswers>false (boolean)</allowedAnswers>
+        <variable typeRef="boolean" name="singleton_001"/>
+        <literalExpression>
+            <text>[100] = 100</text>
```
tarilabs: type(`[100]`) is FEEL list
type(`100`) is FEEL number
This is another instance of a test for equality of two different types, and the test case expect `false` but again the spec
image

sorry :)

tarilabs: I know this is your intention, but still the test as-is I believe is conflicting with the spec snippet I pasted above, and further semantic of equality from this snippet too:
image
So I believe we need some other way to test singleton non-equality, which must be _also_ in the boundary of the spec ;) I am thinking on it but with no solution at the moment from my side

StrayAlien: I'm not sure it matters here whether they are different types or not.  In a sense the test asserts two things - singleton non-equality and different type non-equality.  However, without an explicit test like this I am not sure how we would assert singleton non-equality.  We have to assert it _somehow_ and without a test like vendors may not know they are not 1.2 compliant. 

I'm not so sure that tests need to be within the boundaries of the spec - by their very nature they test the boundaries of the spec.  Like a test for asserting too many params passed to a function, or the wrong typed params - it is the same thing in my view.

tarilabs: By the spec snippet(s) I pasted above, if I interpret it correctly and I would be happy to be corrected, an expression like `[100] = 100` could be either `false` or `null` (and an error).
Your test is enforcing for only the `false` option instead.
That is btw one of the reason we avoided in the TCK to test for errors, to avoid this kind of cases. 

I am not saying I don't want to test singleton list equality; I am saying I would prefer for a test which doesn't enforce for a single outcome when the spec (in my view) is allowing both.

tarilabs: How about (but I would need to verify it) testing for:
```
([100] = 100) != true
```
and expecting `true`?
Again, just wild guessing and I would need later to verify. Am I better able with this pragmatical example to demonstrate better what I am trying to convey?

StrayAlien: Looking at the equality section in the spec I can't see anything that says an equality check between different types is null. The check has to include type equivalence but I think it should just return false not null when types are not equal. 

Or have you seen something else in the spec?  Heaven knows stuff can be a bit spread out.

tarilabs: > Or have you seen something else in the spec?

image

StrayAlien: Thanks, but oh dear. That sucks.

tarilabs: LOL. My suggestion would be to hold (i.e.: comment out) this test and I will raise a specific issue to the RTF to ask for clarification, pointing out these 2 snippet(s) from the spec. Wdyt ?

StrayAlien: The trouble is, it means that `123 = "123"` is null, and likely `123 = null` is null also.  I really can't imagine why the spec designers would consider such a thing.  A violation of "the principle of least surprise".  

It also means a test like `([100] = 100) != true` would be comparing null to true, which would evaluate to null ... so it may as well be `([100] = 100) != "hamburger"`.

However, note your above spec snippet reference begins with 'In general' which is not exactly concrete, so I suspect this may be a reference to the date time thing: 

> "Where necessary, including the value dt function (see 10.3.2.3.6), a date value is considered to be equivalent to a date time value in which the time of day is UTC midnight (00:00:00)."

So actually, I think the tests should be changed to assert that `[100] = 100` is null, as well as `123 = "123"` is null - as *silly* as it seems IMHO.  But, I think you've shown the spec says it as such. 

But .. we should raise a concern with the RTF that the equality thing is nuts and ill-defined.  And seek clarification on what 'In general' means, as well as what "where necessary" means for date and datetime equality. 

I do think we need to keep the `[100] = 100` test in some guise to assert the non-equality of singletons - regardless of whether it is null or false.  But I think, as you've shown, it should be null.

tarilabs: I believe I just found another section of the DMNv1.2 spec where it contradicts that `[a]=a` is false, and expecting to be true. I am going to raise a specific issue and I will keep posted.

tarilabs: Ref: https://issues.omg.org/browse/INBOX-808

The spec says:
image
but
```
get entries({key1 : "value1"})[key="key1"].value = "value1"

  by substitution:

[ { key : "key1", value : "value1" } ][key="key1"].value = "value1"
[ { key : "key1", value : "value1" } ].value = "value1"
[ "value1" ] = "value1"
```

and the expecting to be true. Hence the DMNv1.2 spec seems to be contradicting this proposed test.

In conclusion, if this will be the only "controversial" test I would eventually propose to comment out this specific case only, pending for the linked OMG RTF ticket to be addressed ;)

opatrascoiu:
We should add tests for `null `as well, e.g. `123 = null`. In this case I believe the result should be false (null ). 

I think the equality operator should be treated as a function and hence apply conversions on parameters  when invoked (e.g. convert singleton list into an element and compare).

I will raise a DMN 1.2. ticket.

tarilabs:
> I will raise a DMN 1.2. ticket.

thank you @opatrascoiu can you kindly ensure to also link the issue I existing DMN ticket I already raise and mentioned above (https://issues.omg.org/browse/INBOX-808) because I believe it's indeed connected.

opatrascoiu:
Just raised https://issues.omg.org/browse/INBOX-809 to clarify once and for all what type conversions are supported and in which context they are applied.

StrayAlien:
Thanks @opatrascoiu 

Re the following:

> I think the equality operator should be treated as a function and hence apply conversions on parameters when invoked (e.g. convert singleton list into an element and compare).

That would mean `[100] = 100`.  Which was removed in 1.2.  

But, it sounds to me like both you and @tarilabs agree that equality comparisons between different base types should be `null`. If we can just confirm that I'll update the tests (and add a few more).

tarilabs:
> That would mean `[100] = 100`. Which was removed in 1.2.

@StrayAlien Greg: I believe what Octavian meant, he is suggesting for the (future?) version of the DMN specification to explicit that equality operator `=` should be treated as a function ~like `eq(T, T)` and as such it would apply the rule that type coercion if you pass as actual parameter `eq(LIST<T>, T)` that:

> Similarly, on a function f( p ) where the domain of the formal parameter p does not include lists, but on an invocation f( p a ) the actual parameter p a is a singleton list where the type of p a [1] belongs to the
domain of p, then p a [1] should be used instead

but I believe the _current_ DMN specification does not explicit this for operators in general.

opatrascoiu:
RTF tickets for the above INBOX tickets

- type conversion https://issues.omg.org/browse/DMN13-132
- dropping [a]=a and get entries example https://issues.omg.org/browse/DMN13-134

StrayAlien:
Hi @opatrascoiu, @tarilabs,

https://issues.omg.org/issues/DMN13-132 is requesting a new paragraph and https://issues.omg.org/issues/DMN13-134 is regarding the getEntries example.

Is there are clarification being sought on whether comparisons between different types is `null` or `false`?  Both of those reported issued focus on `[a] = a` but what about `false = 100`?    

I vote for `false` like the other languages I use. :-)

agilepro:
Would like to see a list of the tests from this PR which are controversial, so that we can move forward on checking in the un-controversial ones, and so we can have separate issues about each of the separate problems.

StrayAlien:
I still haven't figured if we've lodged for  clarification on the null VS false thing for equality comparisons between different types.  Anybody know? The two issues raised mentioned above do not seem to cover it explicitly. 

I am happy to raise it if we haven't already.

StrayAlien:
Amended to match `<any type> = null` to be false except for `null = null` which is true (test now included for that).   

Also `<any type> = <any other type>` to be `null` and that includes the singleton array equality check where `[100] = 100` is `null` - this is now asserted on.

I have added extra tests for each type to assert the above.

Comments welcome.

agilepro:
Discussed.  Matteo still wants to look into the problems.  Greg has clarified all the tests and responses.   

I don't know whether this is "waiting on RTF" or not.  Seems like we have the clarity.

opatrascoiu:
I've just noticed an issue that @DanielThanner raised a while back https://issues.omg.org/browse/DMN13-35. 
I am planning to work on a proposal to adrress it.

StrayAlien:
Thanks @opatrascoiu  - does that affect the legitimacy of any tests in this PR?

tarilabs:
@agilepro Acting as concurred during the TCK meeting 2019-05-11, I will proceed to immediate merge.

/cc @StrayAlien @opatrascoiu

PR #226, opened 2019-01-28
0070-feel-instance-of

StrayAlien:


StrayAlien:
@opatrascoiu re removing the tests for complex types.  The tests here for those are really only asserting what the spec 'lattice' says - like asserting a list is not an instance of a string - and what the spec currently provides.  Some of the spec examples on contexts are covered (not using the same data though).  There is some coverage on whether a context or a list is an instance of a type defined by an item definition. 

I think the type equivalency assertions are pretty important as they test whether a value is in or out of the a specified domain.  Items outside of a domain should be coerced to null (eg - I define a BKM with a given context type param and the incoming data doesn't match - no different to number vs string.  That is covered in '0082-feel-coercion' at https://github.com/dmn-tck/tck/pull/238.

It is these things that are more likely to cause us all to fail interoperability (man I hate typing that) - not stuff like (say) the string concat function.  I think the sooner we get in tests for the 'danger' spots the better we'll all be.

Things like function types can't be tested at the moment but we can get pretty good coverage on list and contexts because the spec gives us what we need and the types can be created using item definitions.   

I hope that all makes sense.  Happy for input.

Greg.

opatrascoiu:
In DMN 1.2 the `instance of` operator is defined based on the conformance relation (see table 56). The conformance relationship is based on the full type definition not on the erasure of the type. 

The boolean logic in DMN/FEEL is 3-value (true, false and null). Table 56 states when the result is true. the expected result is not obvious if the type is incorrect (e.g. typo, missing type or missing type parameters). It could be false or null.

StrayAlien:
Thanks @opatrascoiu .  Appreciated.  If I may ask, could you expand on:

> the conformance relationship is based on the full type definition not on the erasure of the type.

This is a very important aspect of FEEL so I am asking for my education, but also to clarify whether you are saying the tests do not represent the spec, or there are issues with the tests. No offence meant - but if I have it wrong let’s please make it right!

The tests cover table 56 in that they assert that the type examples given in section 10.3.2.9 “Relations between types” and also how null relates to that.  Null does not conform to any type, and everything (except null) conforms to the Any type.  

The tests cover the spec lattice with a number of simple assertion like (say) a string does not conform to date.

Function types are covered only as per the above - no params tested.

Context types gets some coverage, but only comparing a value with a simple context ItemDefinition - same for lists which has a test for a typed list and an ‘Any’ list.   This is to cover the section on Type Conformance and Type Equality.

tarilabs:
While reviewing these tests and the spec section as highlighted by Greg, I believe I may have found what could be either an issue in the spec (missing explicit "iff" rule when type operands of different types, eg a `list<context<>>` Vs `context<>`) or a typo in the example or something which may uncover other potential pitfalls, so I would hold these tests until the following RTF ticket gets clarified: https://issues.omg.org/browse/INBOX-819 imho thanks :)

opatrascoiu:
The RTF ticket raised by Matteo is here https://issues.omg.org/browse/DMN13-141. 
The examples are incorrect. I made a proposal to fix them.

Thanks Matteo for spotting them.

StrayAlien:
Thanks @opatrascoiu .  As a note, with `browse` in the URL you need to login (as an OMG member).  With `issues` in the URL as such: `https://issues.omg.org/issues/DMN13-141`.  You do not need to log in - it seems to be public.

opatrascoiu:
Here is the latest rule for type from the DMN spec

```
54. type = 
         qualified name |
         'list' '<' type '>' |
         'context'  '<'  name ':' type { ',' name ':' type }  '>' |
         'function' '<' [ type { ', ' type } ] '>'  '->'  type
```

where qualified name is either a user type (imported or not) or a primtive FEEL type (e.g. string, number).

I think the expected value should be null when there are syntactic or semantic errors. For example,  incorrect type name, missing type name (not imported ), or incorrect type expression (e.g. list<> instead of list<Any> or context<>)
 
I've raised an RTF issue to clarify the semantics https://issues.omg.org/browse/INBOX-826

opatrascoiu:
The ticket is in the DMN area under https://issues.omg.org/browse/DMN13-144. I attached a proposal, should be discussed in the near future.

agilepro:
I tried this link and it prompts for log in.   If there is a way to access without log in please try attaching it.

opatrascoiu:
https://issues.omg.org/issues/spec/DMN/1.2

agilepro:
not finalized.  Still waiting for RTF on this.   Then we need to determine if this is a "clarification" to 1.2 or a change that needs to be in 1.3.

agilepro:
Is this still being discussed in the RTF?

StrayAlien:
@agilepro not quite sure what we're waiting on.  The issue raised (above) is for some clarification on the return value of instance of for typos or missing types etc.  I'm not sure we cross those in this suite.

The original comments re same are here I think:

https://github.com/dmn-tck/tck/pull/204#pullrequestreview-196458935
https://github.com/dmn-tck/tck/pull/204#issuecomment-457989500

Maybe @opatrascoiu can chime in here but my feeling is, if that issue raises anything notable we can add extra test to cover that, but I _think _ the tests I've done here stay within what the spc says now - like not testing function params within func types etc.  That can come later.

agilepro:
There is a note from Octavian saying this was on the ballot to be discussed in February.   I don't see any big outstanding issue here.   Can we mark this "approved"?

agilepro:
What is the status of this?   Seems to be sitting a long time.   The issue in RTF was updated on May 9, so maybe still being discussed?   Would like to get a comment from Octavian on this, and Matteo has said he will look into it as well.

opatrascoiu:
I think we are in a position to move forward on this.

Based on the latest of https://issues.omg.org/issues/DMN13-144. The tests with the following pattern:

```
XXX instance of list
XXX instance of context
XXX instance of function
```
should return null as the composite types require full specification, e.g. list<Any>.

Tests with pattern like 

```
XXX instance of Any
```
should return true as all types conform to Any.

agilepro:
Discussed and Matteo will try to comment out the questionable items in this PR so that we can decide to merge the PR at the end of the month.

tarilabs:
I have applied with 7181d23 a backoffice comment out of tests subject of DMN13-144 and assimilated*  as per https://github.com/dmn-tck/tck/pull/226#issuecomment-495690112 as agreed during TCK meeting 2019-05-24

(*)accordingly also tests with the form `... instance of itemDef` where idemDef is semantically equivalent of tests mentioned in referenced comment.

From my perspective, it is okay to be merged.

agilepro:
Looks like this is ready to go   Please merge after the beginning of the month.

tarilabs:
@opatrascoiu if you have any additional reservations on some results in this test case/PR please raise a GitHub issue or comment on this PR.

opatrascoiu:
@tarilabs I checked the latest version of the pull request. The tests look fine to me. One the tests (date_007) contradicts one of the examples in the spec. I believe the example in spec is incorrect. I raised an RFT issue to fix it https://issues.omg.org/browse/INBOX-866.

tarilabs:
Thanks for the update @opatrascoiu !

PR #230, opened 2019-01-28
0074-feel-properties

StrayAlien:


tarilabs:
rechecked and good for me, thanks!

tarilabs:
[headsup for merge, please use commit message `0074-feel-properties`]

opatrascoiu:
`Table 62: List of properties per type`  contains for each temporal type a list of allowed properties.

There are tests for the incorrect use of these properties. The expected value is null, which correct. 
However, because this is a semantic error, the errorResult must be true. For example, ym_duration_003.

StrayAlien:
thanks @opatrascoiu.  I have to admit to being a little vague of what the idea of an 'error' is in the test suite granted the spec says nothing about error conditions beyond for functions returning null.  I'm am opening an issue to clarify.

PR #233, opened 2019-01-28
0077-feel-nan

StrayAlien:


opatrascoiu:
looks fine to me

tarilabs:
ok from me too, thanks!

agilepro:
@StrayAlien Greg,  we discussed this, but everyone thinks this is redundant with 0078 and so we don't want to include this test which does not add anything.   Do you agree, or is there some added value to this test which we have not considered?

StrayAlien:
@agilepro @opatrascoiu @tarilabs.  

I have just changed this assertion to `0.0 / 0.0`.  I believe they are both now valid suites.  Apols - the assertion here didn't capture what I wanted.  My fault for writing tests late at night!

Let me explain.  The Infinity/NaN tests are to catch situations is what is likely to be the most common impl languages at this time.  Java and JS/Node.  They are not specifically div-by-zero tests.

Both JS and Java report `1.0/0.0' and `-1.0/0.0` as _infinity_ - not as an exception - hence the tests in 0078.  That is to catch impls that assume an exception and will report null in that case.  No exception with those calcs so an impl might be inclined just to pass on the result.   

Same for this (now corrected) test.  Java/JS will report `0.0/0.0` as `NaN` - not an exception - so the test is to protect against the same situation as 0078 - an impl just passing a result back.

Note that Java will throw an exception for an _integer_ div by zero which most people will catch.  And also note that `BigDecimal.valueOf(0.0).divide(BigDecimal.valueOf(0.0))` will also throw an exception ... but JS will not in either circumstance.

So both these suites are really to protect against an impl just passing a result back and not specifically checking for NaN/Infinity if their numbering impl does not throw exceptions.  

We do need a div by zero test somewhere but maybe not as 'inifinity'.

PR #234, opened 2019-01-28
0078-feel-infinity

StrayAlien:


StrayAlien:
Thanks @opatrascoiu. They do yes. As many impls will be in Java these test are meant to catch those situations in java. Looking at it, I think the Nan test should actually be zero / zero.

They also differ in that one deals with integers and the other floating point. 

Arguably they could be better expressed but the intention was to catch what a majority of impls might do. 

Comments welcome. 

Greg.

PR #235, opened 2019-01-28
0079-feel-string-function

StrayAlien:


tarilabs:
Based on the explaination provided in TCK meeting about standardization and consensus concerning `string` FEEL function, I would vote to hold this test case or move in non-conformant folder

agilepro:
We would like to move string tests in the "non-conformant" folder where vendors can run the test, but the results are not reported on the web site.   @StrayAlien Can you move this?

StrayAlien:
@agilepro, @tarilabs. moved.

tarilabs:
thanks for the move, lgtm

PR #237, opened 2019-01-28
0081-feel-getentries-function

StrayAlien:


TestCases/compliance-level-3/0081-feel-getentries-function/0081-feel-getentries-function.dmn
```diff
@@ -0,0 +1,100 @@
...
+        <variable typeRef="list" name="decision_003"/>
+        <literalExpression>
+            <text>get entries(null)</text>
+        </literalExpression>
+    </decision>
+
+    <decision name="decision_004" id="_decision_004">
+        <description>Tests FEEL expression: 'get entries({a: "foo", b: "bar"})' and expects result: '[{a: "foo}, {b: "bar"}]'</description>
+        <question>Result of FEEL expression 'get entries({a: "foo", b: "bar"})'?</question>
+        <allowedAnswers>[{a: "foo}, {b: "bar"}]</allowedAnswers>
```
tarilabs: This comment is unconsistent with the expected result.
Suggested change
From `<allowedAnswers>[{a: "foo}, {b: "bar"}]</allowedAnswers>`
To `<allowedAnswers>[{value=foo, key=a}, {value=bar, key=b}]</allowedAnswers>`

opatrascoiu:
TypeRef `list` is not defined in
  
            `<variable typeRef="list" name="decision_004"/>`

List types have to be defined via <itemDefinitions>. List types in FEEL have one parameter (e.g. list<string>).

StrayAlien:
@opatrascoiu @tarilabs. 'list' type removed from all tests, but as per comment: https://github.com/dmn-tck/tck/pull/238#issuecomment-462084211, the lack of ability to use 'any' as a type means I would not be able to create an itemDefinition for the return value of getEntries()

agilepro:
Still waiting for RTF vote on Matteo's proposal

agilepro:
balloted.  If approved we can check in.

StrayAlien:
@tarilabs - the text given in https://issues.omg.org/issues/DMN13-134 assumes the `[a] = a` is `false` - given that we now have agreed it is actually `null` and not `false` does the amended example still hold true? (I have not yet implemented the odd 'nested lists' thing)

cc @opatrascoiu

opatrascoiu:
looks good to me

tarilabs:
Based on https://github.com/dmn-tck/tck/pull/237#issuecomment-480343447 https://github.com/dmn-tck/tck/pull/237#pullrequestreview-229946354 https://github.com/dmn-tck/tck/pull/237#issuecomment-486732107 will merge.

PR #238, opened 2019-01-28
0082-feel-coercion

StrayAlien:


opatrascoiu:
TypRef `list `is note defined via `<itemDefinition>`

`<variable`` typeRef="list" name="decision_002"/>`

StrayAlien:
Thanks @opatrascoiu. Darn!! Scouring the spec I do believe you're right. I'd love to be able to say otherwise. It seems like a spec oversight to me. List, context and range should be basic types I believe. 

I'll amend. Thanks. Your comment re same also noted in other PR.

tarilabs:
from my side `list` was allowed as a built-in FEEL type.
Otherwise I agree the spec is a bit unclear, for instance in this paragraph:
image
is clearly missing to enlist FEEL built-in `date`

StrayAlien:
Well spotted there @tarilabs.  That was the paragraph that clinched it for me - no lists.  But sir, you have an eagle eye - I didn't spot the missing date. I guess for now I'll update the tests to remove lists.

StrayAlien:
@tarilabs @opatrascoiu.  'list' thing removed (well, one modified to be a `tStringList` and the other commented out.

However, according to above spec fragment, it seems we cannot use `any` as a type - even in an item definition.  So we're left with no way to say that a list may contain mixed element types.   Looking at the XSD it seems an item definition _must_ have a typeRef if it does not contain components .... but we can't use `any` ....

Or have I missed something?

opatrascoiu:
I think one should be able to use `Any `in typeRefs, just like any primitive FEEL type.

The list and context types, can be described via `<itemDefinition>`. However, there is no way of describing the function types in a similar way. 

We should raise an RTF ticket to fix this if we agree on it.

StrayAlien:
@tarilabs. Added new test `decisionService_001` to assert same as for other types.

StrayAlien:
added new tests for decision services covering direct and indirect invocation for null and singleton list coercion of params.

agilepro:
The RTF is still discussing this?

StrayAlien:
@agilepro, not sure about RTF, maybe @opatrascoiu or @tarilabs can chime in.

I do need to modify or lose some tests here - I have recently seen that the spec says "If an instance of Expression that defines the output of a Decision element includes a typeRef, the
referenced type SHALL be the same as the type of the containing Decision element."

It does _not_ say the same for BKM or Decision Service though.

So some small tidy up required here.  I'll get that done this week.

TestCases/compliance-level-3/0082-feel-coercion/0082-feel-coercion.dmn
```diff
@@ -0,0 +1,505 @@
...
+        <variable typeRef="tNameAndAge" name="decision_005"/>
+        <literalExpression>
+            <text>{name: "foo"}</text>
+        </literalExpression>
+    </decision>
+
+    <decision name="decision_006" id="_decision_006">
+        <variable typeRef="tStringList" name="decision_006"/>
+        <literalExpression>
+            <text>"foo"</text>
```
tarilabs: This test seems to me to rely on DMN13-132 hence I wonder if we can truly say it is DMNv1.2

opatrascoiu:
I've been having the same feeling since we started to look at DMN13 tickets ...

It feels like working on a DMN 1.3 snapshot.

Lets discuss in the next meeting.

StrayAlien:
actually, looking at it, I think it should probably be nil/error.   As far as I can tell, the box/unbox array thing only happens on function/invokable call.  I would argue that this test is not correct (yes - I know, I wrote it) and am happy to amend.

Re DMN13-132 - this has come up a number of times and I still don't understand the relevance of the ticket: "type conversions are too spread out in Ch 10" and the call-to-action is about consolidating them in the spec - I don't see that it is clarifying anything regarding types.    The fn() part of it is a reworking of this I believe:

image

Or have I missed something?

At any rate - happy to amend this test (there may be some others too) but given what I understand now I think it should be nil/error.

agilepro:
Discussed this for a while in the meeting.  Some of these tests suggest results that are different from what DROOLS produces.  The spec is ambiguous on these points.  I have argued that we need not wait for the spec to be clarified:  we only need to come to agreement on what an answer should be within the TCK in order to write a test for that.    Our job is not to anticipate what the RTF will say, but only to do our best job of interpreting the spec as it is.  If the RTF later changes the spec for 1.3 we will simply change the tests to correspond to the clarification.

Matteo has offered to go through this list of tests, and indicate the tests that he has confidence are correct.  There are ~40 tests in here and most of them are fine. Let's get those checked in.

The tests with low confidence should be pulled out and resubmitted with a separate request.  Matteo makes a good point that just because DROOLS gets a particular answer does not mean that he has high confidence that it is the right answer.  He would like to spend some time considering what the right and proper interpretation of the spec should be for the edge cases that Greg has exposed.

tarilabs:
I spent some time checking this PR -as always, thanks Greg for your efforts.

I believe there are a number of issues, I'll try to summarise below

### Decision Service tests

There are a number of test cases which relies on Decision Service, and which are not consistent with the other proposed PRs we currently have open to agree on a basis for Decision Service testing from the TCK API/infra point of view.

ACTION My proposal is to defer this tests at least until said PRs are merged, and we have at least the basics of Decision Service we agree on.
This applies to test cases named with prefix `decisionService_`.
I can do by backoffice if we agree.

### typeRef inconsistencies

There are a number of test cases which inconsistently have the `typeRef` not on the variable element but the decision logic element. I would prefer to have it all on the variable element

ACTION My proposal is to update tests to have the typeRef on the variable element.
This applies to test cases named with prefix `literal_` and `invoke_` if I checked correctly.
I can do by backoffice if we agree.

### Expected result is a wrapped element

A number of tests cases expect the result to auto-wrap as a singleton list.
For instance a Decision results in `123`, while because variable typeRef is a list of number, expected result is `[123]`.
I am not sure where DMNv1.2 specify this.
I am open to be shown where this is mandated by the spec.

ACTION I cannot understand the result of these cases, and to me unless shown otherwise on the DMNv1.2 spec, I propose to **_erase them_**.
This applies to test cases
- `decision_006` 
- `decision_bkm_004`
- `invoke_003`
- `literal_003`

I can do by backoffice if we agree.

### Result of `invoke_001` is wrong.

The test is an Invocation, but semantically is assimilated to an analogous function invocation of:
```
bkm_001({name: "foo"})

where bkm_001 is a BKM analogous to:

function bkm_001( nameAndAge : tNameAndAge) nameAndAge != null

so by subsitution we have:
bkm_001({name: "foo"})
bkm_001(null) // as value {name: "foo"} does not conform to type
null != null
false
```

BKM does not define explicitly the result type (`Any`) hence `invoke_001` test case result should be false, while original PR proposes it's expecting error, ref:
https://github.com/dmn-tck/tck/pull/238/files#diff-e310d677e25aabf0ef8d45d63fd39204R204

ACTION Fix `invoke_001` test case result should not be expecting error, and expected result `false`.
I can do by backoffice if we agree.

### Result of `invoke_005` is wrong.

The test is an Invocation, but semantically is assimilated to an analogous function invocation of:
```
bkm_005([10])

where bkm_005 is a BKM analogous to:

function bkm_005( arg ) [arg]

so by subsitution we have:
bkm_005([10])
[[10]]
null   // BKM variable's typeRef is number.
```

BKM do define explicitly the result type (`number`) hence the result just out of the BKM invocation is null because `[[10]]` does not conform to `number`.

ACTION Fix `invoke_005` test case result should be expecting error (and `null` as test case value result)
I can do by backoffice if we agree.

tarilabs:
I've implemented by backoffice what reviewed, discussed and agreed during TCK meeting 2019-06-07, proceeding to merge.

PR #239, opened 2019-01-28
0083-feel-unicode

StrayAlien:


TestCases/compliance-level-3/0083-feel-unicode/0083-feel-unicode.dmn
```diff
@@ -0,0 +1,133 @@
...
+        <literalExpression>
+            <!-- horse + grinning face emoji -->
+            <text>contains("\ud83d\udc0e\uD83D\uDE00", "😀")</text>
+        </literalExpression>
+    </decision>
+
+    <decision name="decision_006" id="_decision_006">
+        <variable name="decision_006"/>
+        <literalExpression>
+            <text>{🐎: "bar"}</text>
```
tarilabs: FEEL grammar rule for `61. key = name | string literal ;` taking the `name` path;
which in turn is `27. name = name start , { name part | additional name symbols } ;`
which in turn is `30. name start char = ...`
I do not believe the horse-emoji can be used in here (without making it within a string literal double quote start/end)

agilepro: Rule 30 on page 110 says:

name start char = "?" | [A-Z] | "_" | [a-z] | [\uC0-\uD6] | [\uD8-\uF6] | [\uF8-\u2FF] | [\u370-\u37D] | [\u37F-\u1FFF] | [\u200C-\u200D] | [\u2070-\u218F] | [\u2C00-\u2FEF] | [\u3001-\uD7FF] | [\uF900-\uFDCF] | [\uFDF0-\uFFFD] | [\u10000-\uEFFFF] ;

The character is 1F40E which clearly falls into the last range mentioned in rule 30.

Seems to me that this is explicitly allowed by the spec unless I am misunderstanding something about the spec.

StrayAlien: Thanks @agilepro , @tarilabs.  I just did the same calc, plus doubled checked the lexer code and I believe it is a valid start of name char.

agilepro: Worth mentioning here that this is not effected by the \uXXXX issue because in this case the character is specified in UTF-8.  There are no ambiguities or outstanding questions with UTF-8 directly in the source text.   In this case, the character is clearly specified in the text, and the spec says it is allowed.   Matteo, please look at it again.

tarilabs: will do.
On other hand this test suite is affected by the problem with the `string length` which is still not addressed ;)

agilepro:
Octavian says there is a ticket at RTF for the string length.   I don't know the ID.  There was some discussion about being able to link directly to the ticket.   Can someone add the link here?

StrayAlien:
@agilepro @tarilabs @opatrascoiu - if I changed the string length test to be based on utf16 length and not codePoint counts would it clear this fr merging?  I am happy to do so - I am not religious about it at all.

agilepro:
once again, all the raised issues have been addressed.   String length has been changed to be in accordance with the proposal to RTF, which is in accordance with all the existing implementations.   If the RTF decides to do something different, not just the tests will have to be changed, but all the implementations.   That being the case, I don't see why we can't move forward on adopting these tests.

agilepro:
Apparently discussion in RTF about "business rationale".

There is no “business rationale” behind the string length.   The proposal is simply define it to match the way that Java, JavaScript, and all the known implementations do it.

A purist can certainly claim that “code-points” are the fundamental building block, and those are what *should* be counted.  

Another purist would insist that it should count the number of “graphemes”.   A grapheme is a unit that looks to the user like a single block.  For example, å can be made with two code points:  an A with a COMBINING RING.  Emojis can be combined with as many as 7 code points used to make a single grapheme.  so string length("å") == 2 if you do code points, but it can also be done in a single code point.

My point, of course, is that there is nothing more "intuitive" about code points than UTF-16 code values.   Neither really reflect what the user sees, so it make more sense to conform to the "de jure" standard of counting UTF-16 code values.

StrayAlien:
@agilepro - I haven't changed the tests as yet - I will do though - like I said, not religious about it, as long as we all do the same thing. I'll amend in the coming days.

StrayAlien:
@agilepro - tests changed. Length is now as per Java.

opatrascoiu:
Hi Keith,
Aplogies for the delay. Here is the feedback from the OMG RTF discussion:
 1. We decided not to maintain SFEEL.
2. Shouldn't the FEEL change be:
"\'" | "\"" | "\ \" | "\n" | "\r" | "\t" | "\u", hexadecimal digit, 5 * [hexadecimal digit] ?
3. I don't think that changing string length to sometimes return 2 for a string containing 1 character is a step forward for business users.
I think we should specify somewhere the default encoding (e.g. UTF-16 vs UTF-32). Once we decide about that, string length will be natural (2 for UTF-16 or 1 for UTF-32 for the horse character). It looks like some members of the RTF team prefer UTF-32.
I will not be able to join the TCK call today. 
Kind regards,
Octavian
On Tuesday, 23 April 2019, 18:53:18 BST, Keith Swenson <notifications@github.com> wrote:
 
 
 
Apparently discussion in RTF about "business rationale".

There is no “business rationale” behind the string length. The proposal is simply define it to match the way that Java, JavaScript, and all the known implementations do it.

A purist can certainly claim that “code-points” are the fundamental building block, and those are what should be counted.

Another purist would insist that it should count the number of “graphemes”. A grapheme is a unit that looks to the user like a single block. For example, å can be made with two code points: an A with a COMBINING RING. Emojis can be combined with as many as 7 code points used to make a single grapheme. so string length("å") == 2 if you do code points, but it can also be done in a single code point.

My point, of course, is that there is nothing more "intuitive" about code points than UTF-16 code values. Neither really reflect what the user sees, so it make more sense to conform to the "de jure" standard of counting UTF-16 code values.

agilepro:
Octavian,

I think we agree that:

  string length("\uD83D\uDC0E") == 2

don't we?   I am not proposing a "change" to anything.  I think we should go ahead and commit these tests.   Later, if the RTF changes the meaning of the spec, we can change our tests.   My proposal is to leave everything as it is today.

Can you outline the reasons that you think we should continue to hold up this commit for the RTF to change the definition?  We should go ahead and commit today, and if they change the spec later, we can change our tests to suit.  Isn't that the more logical way to proceed?

opatrascoiu:
Latest comments from RTF:
https://issues.omg.org/issues/DMN13-127

```
1. typo involving FFF->FFFF, change grammar rule hex digit -> hexadecimal digit, etc. (see above), other simple bugs?
2. change defn of string length to depend on unicode encoding (resistance to this proposal)
We can fix obvious bugs but maybe not change string length

```
The idea is to fix the typos/defects in the grammar & string length to return the number of Unicode characters.

PR #247, opened 2019-02-08
Add JDK 11 related dependencies to Drools runner

baldimir:
@tarilabs @etirelli could you please review? This enables running Drools runner with jdk11+.


tarilabs:
Clarified this for now is just to adapt Drools runner to run on JDK11+. If other vendors or the full tck later migrates it could incorporate these changes, when needed/suitable.

PR #251, opened 2019-02-12
Test has invalid type "tFnInvocationResult"

StrayAlien:
Test had typeRef="tFnInvocationResult" in two places.  tFnInvocationResult in not defined in the model.  I had a flag to relax this stuff as I was working but, truth be told, it should not pass and is not a valid model.

The test is not a test for types so I have just removed the typeRefs.

### DIFF
--- a/TestCases/compliance-level-3/0031-user-defined-functions/0031-user-defined-functions.dmn
+++ b/TestCases/compliance-level-3/0031-user-defined-functions/0031-user-defined-functions.dmn
@@ -86,7 +86,7 @@
         <informationRequirement id="_1080c15e-bc84-4473-99f3-1b8ea7d12e61">
             <requiredInput href="#_5eBhQH6PEeePe9Zmt-encA"/>
         </informationRequirement>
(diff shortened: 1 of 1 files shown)


agilepro:
We discussed, but we are not clear why this is considered invalid.   Can you explain?

tarilabs:
(this PR is OK for me)

StrayAlien:
Hi @agilepro, The test has a typeRef that refers to a user-defined type `tFnInvocationResult` that is not defined anywhere in the model.  I would have thought this would make it an invalid model?

I guess (like I was during development) others are relaxing errors for missing types, but, tightening a few bolts, this crops up. 

cc @tarilabs

agilepro:
@StrayAlien after more discussion we figured out that a custom type is referenced, but the definition is not present.  I think we all agreed that this is a problem that we really don't want in the tests.  Matteo has approved.  Octavian wanted to look at it again, and will probably approve the change.  I will merge it when he does.   Thanks again for raising this issue -- and fixing it.

StrayAlien:
Thanks @agilepro

opatrascoiu:
Looks fine to me.

PR #252, opened 2019-02-17
0085-decision-services

StrayAlien:
A few tests around Decision Services.  Not really sure what kind of tests here but it seems not too many.  Just covering outputDecisions and invoke-ability really.   Spec calls for function semantics with decision services but it has no formal params properties, so I have just asserted that it can be called as a function with no params.

I have also added another test into `0082-fell-coercion` to decision table 'null coercion'. 

Can anybody think of any more tests we could do here?

All comments welcome.

### DIFF
--- a/TestCases/compliance-level-3/0085-decision-services/0085-decision-services-test-01.xml
+++ b/TestCases/compliance-level-3/0085-decision-services/0085-decision-services-test-01.xml
@@ -1,0 +1,47 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<!-- Contributed to DMN TCK by StrayAlien -->
+<testCases xmlns="http://www.omg.org/spec/DMN/20160719/testcase" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+
+    <modelName>0085-decision-services.dmn</modelName>
+    <labels>
+        <label>Compliance Level 3</label>
+        <label>Decision Services</label>
+    </labels>
+
+    <testCase id="001">
+        <description>Will output decision result</description>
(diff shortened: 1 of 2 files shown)


opatrascoiu:
I think the DMN model is incorrect, the encapsulatedDecision property is not populated correctly  (see Figure 5.10 and 5.11 from DMN 1.2 spec)

StrayAlien:
Hi @opatrascoiu - thanks - I was wondering if anybody was ever going to have a look here!  

To confirm, I am figuring you mean: 

```
    <decisionService name="decisionService_002" id="_decisionService_002">
        <outputDecision href="#_decision_001"/>
        <outputDecision href="#_decision_003"/>
        <encapsulatedDecision href="#_decision_002"/>
    </decisionService>
```
does not require `#_decision_002` as an encapsulatedDecision?

opatrascoiu:
Chapter 5.3.3 Decision services states that:

```
The service must encapsulate not just the output decisions but also any decisions in the DRG directly
or indirectly required by the output decisions which are not provided in the input decisions (the “encapsulated decisions”).

```
hence definition should be

```
<decisionService name="decisionService_002" id="_decisionService_002">
        <outputDecision href="#_decision_001"/>
        <outputDecision href="#_decision_003"/>
        <encapsulatedDecision href="#_decision_001"/>
        <encapsulatedDecision href="#_decision_002"/>
        <encapsulatedDecision href="#_decision_003"/>
    </decisionService

``` 

Tables 5.10 and 5.11 contain 2 examples for the above rule - output decision must be in the encapsulated collection.

StrayAlien:
Thanks @opatrascoiu, noted.  My misinterpretation.  Granted I think I have not implemented DS correctly as yet (discussed here: https://github.com/dmn-tck/tck/issues/259), I'll close this PR and have another go at it when done.  :-)

PR #257, opened 2019-03-01
0004-lending

StrayAlien:
Small correction of some types from "string" to what they actually represent in the model, "boolean", "number" etc.

This was originally in the big 'consolidated' PR but got lost when we split them up.

https://github.com/dmn-tck/tck/pull/204#issuecomment-448089266

### DIFF
--- a/TestCases/compliance-level-3/0004-lending/0004-lending.dmn
+++ b/TestCases/compliance-level-3/0004-lending/0004-lending.dmn
@@ -616,7 +616,7 @@
                     </inputExpression>
                 </input>
                 <input id="_1d794320-6e6c-46bf-b86d-de073822d6f6" label="Pre-bureauAffordability">
-                    <inputExpression typeRef="string">
+                    <inputExpression typeRef="boolean">
                         <text>Pre-bureauAffordability</text>
                     </inputExpression>
                 </input>
@@ -1274,7 +1274,7 @@
                     </inputExpression>
                 </input>
                 <input id="_fc1055cf-ab08-44a0-91bc-2d6539df00c5" label="Post-bureauAffordability">
-                    <inputExpression typeRef="string">
+                    <inputExpression typeRef="boolean">
                         <text>Post-bureauAffordability</text>
                     </inputExpression>
                 </input>
--- a/TestCases/compliance-level-3/0004-lending/0004-lending.dmn
+++ b/TestCases/compliance-level-3/0004-lending/0004-lending.dmn
@@ -734,12 +734,12 @@
             <formalParameter typeRef="number" name="ApplicationRiskScore"/>
             <decisionTable hitPolicy="UNIQUE" outputLabel="Pre-bureauRiskCategoryTable" preferredOrientation="Rule-as-Row">
                 <input id="_6f7fbf13-3fa6-44f5-afe4-7c1e688e76a9" label="ExistingCustomer">
-                    <inputExpression typeRef="string">
+                    <inputExpression typeRef="boolean">
                         <text>ExistingCustomer</text>
                     </inputExpression>
                 </input>
                 <input id="_89d36bd5-03c2-4c6f-bf47-faa7d28e13a1" label="ApplicationRiskScore">
-                    <inputExpression typeRef="string">
+                    <inputExpression typeRef="number">
                         <text>ApplicationRiskScore</text>
                     </inputExpression>
                 </input>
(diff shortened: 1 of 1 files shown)


opatrascoiu:
Looks fine to me.

agilepro:
Matteo would like to review this.  Trying to find the time.

TestCases/compliance-level-3/0004-lending/0004-lending.dmn
```diff
@@ -616,7 +616,7 @@
                     </inputExpression>
                 </input>
                 <input id="_1d794320-6e6c-46bf-b86d-de073822d6f6" label="Pre-bureauAffordability">
-                    <inputExpression typeRef="string">
+                    <inputExpression typeRef="boolean">
```
tarilabs: With BKM parameters defined as:
```
            <formalParameter typeRef="tRiskCategory" name="Pre-bureauRiskCategory"/>
            <formalParameter typeRef="boolean" name="Pre-bureauAffordability"/>	            
            <formalParameter typeRef="number" name="Age"/>	            
```
I agree this second table input is a `boolean`. Technically also previous parameter ("Pre-bureauRiskCategory") could be amended to reflect from `string` to `tRiskCategory`, wdyt?

TestCases/compliance-level-3/0004-lending/0004-lending.dmn
```diff
@@ -734,12 +734,12 @@
...
             <decisionTable hitPolicy="UNIQUE" outputLabel="Pre-bureauRiskCategoryTable" preferredOrientation="Rule-as-Row">
                 <input id="_6f7fbf13-3fa6-44f5-afe4-7c1e688e76a9" label="ExistingCustomer">
-                    <inputExpression typeRef="string">
+                    <inputExpression typeRef="boolean">
                         <text>ExistingCustomer</text>
                     </inputExpression>
                 </input>
                 <input id="_89d36bd5-03c2-4c6f-bf47-faa7d28e13a1" label="ApplicationRiskScore">
-                    <inputExpression typeRef="string">
+                    <inputExpression typeRef="number">
```
tarilabs: LGTM as BKM parameters are defined as:
```
            <formalParameter typeRef="boolean" name="ExistingCustomer"/>
            <formalParameter typeRef="number" name="ApplicationRiskScore"/>	            
```
so indeed boolean, number.

TestCases/compliance-level-3/0004-lending/0004-lending.dmn
```diff
@@ -843,17 +843,17 @@
...
                 </input>
                 <input id="_fba27bc9-7678-4457-893e-6bc96fc69626" label="ApplicationRiskScore">
-                    <inputExpression typeRef="string">
+                    <inputExpression typeRef="number">
                         <text>ApplicationRiskScore</text>
                     </inputExpression>
                 </input>
                 <input id="_b54042e3-1fe7-4fb3-90d6-ec528047fafb" label="CreditScore">
-                    <inputExpression typeRef="string">
+                    <inputExpression typeRef="number">
```
tarilabs: LGTM the BKM params are defined as:
```
            <formalParameter typeRef="boolean" name="ExistingCustomer"/>
            <formalParameter typeRef="number" name="ApplicationRiskScore"/>	            
            <formalParameter typeRef="number" name="CreditScore"/>	            
```
and as the decision table is using them directly, so `boolean`, `number`, `number`

TestCases/compliance-level-3/0004-lending/0004-lending.dmn
```diff
@@ -1284,7 +1284,7 @@
                     </inputExpression>
                 </input>
                 <input id="_f692dbba-dc13-4076-a5ba-275041c0b944" label="CreditScore">
-                    <inputExpression typeRef="string">
+                    <inputExpression typeRef="number">
```
tarilabs: Analogous comment as above, also for parameter "Post-bureauRiskCategory" used in the decision table

StrayAlien:
Thanks @tarilabs - very interesting observation re string type vs a string type with allowedValues.  

The PR for type coercion, has no assertions for when the types are compatible but a value falls outside the allowed values.   Which I think it should do.

So say, if type `tFoo` is a string type with allowedValues of `"A", "B", "C"` and I have a BKM with a formal param of type `tFoo` that I pass the value `"D"` to ... then it should null-coerce (with `errorResult=true` for the test.

I hadn't considered that.

tarilabs:
My comment was more about the fact a Decision such as Pre-bureauRiskCategory type is not really a FEEL `string`, but precisely defined type `tRiskCategory` in the model, that is indeed a fixed set of enumerated `string`. As this PR deals with correcting the types, I thought it was worthy to check in the details ;)

StrayAlien:
Heya - I do understand what you meant. I guess, strictly speaking, `tRiskCategory` is a narrowing of `string` and therefore can be passed to anything accepting a `string` - so I think the test is still correct .. though maybe it could be made _more_ correct I guess.  :-)

I can fix this up with another PR (after merge) if you like, but I'll also add a test or two to the type-coercion PR.  That is a good catch and an omission on my part.

agilepro:
OK, all approved to merge after the beginning of the month.

PR #258, opened 2019-03-01
0012-list-functions

StrayAlien:
Some type fixes to align with v1.2 types (another v1.1 'singleton array' thing).

The summary: the list results are of mixed types and not pure string lists as the dmn specifies.  Likely was okay with 'singleton array' code but, truth be told, they are not string lists anymore.

The purpose of the test suite is testing lists, not types, so, I have simply removed the offending types.   

This was originally in the big 'consolidated' PR but got lost when we split them up.

Mentioned here: https://github.com/dmn-tck/tck/pull/204#issuecomment-448089266

And some discussion here: https://github.com/dmn-tck/tck/pull/204#issuecomment-452887856

### DIFF
--- a/TestCases/compliance-level-3/0012-list-functions/0012-list-functions-test-01.xml
+++ b/TestCases/compliance-level-3/0012-list-functions/0012-list-functions-test-01.xml
@@ -443,18 +443,8 @@
 				</item>
 			</list>
 		</inputNode>
-		<inputNode name="list2">
-			<list>
-				<item>
-					<value>x</value>
-				</item>
-				<item>
-					<value>y</value>
-				</item>
-				<item>
-					<value>z</value>
-				</item>
-			</list>
+		<inputNode name="string1">
+			<value>x</value>
 		</inputNode>
 		<resultNode name="appendListItem" type="decision">
 			<expected>
@@ -469,17 +459,7 @@
 						<value>c</value>
 					</item>
 					<item>
-						<list>
-							<item>
-								<value>x</value>
-							</item>
-							<item>
-								<value>y</value>
-							</item>
-							<item>
-								<value>z</value>
-							</item>
-						</list>
+						<value>x</value>
 					</item>
 				</list>
 			</expected>
--- a/TestCases/compliance-level-3/0012-list-functions/0012-list-functions.dmn
+++ b/TestCases/compliance-level-3/0012-list-functions/0012-list-functions.dmn
@@ -61,7 +61,7 @@
         </literalExpression>
     </decision>
     <inputData name="numList" id="_b894ee22-f795-46f7-b713-77d69d07ae22">
-        <variable name="numList"/>
+        <variable typeRef="tNumList" name="numList"/>
     </inputData>
     <decision name="sum1" id="_d6b2c4a0-a147-44a1-a9f4-3575f1bb5695">
         <variable typeRef="number" name="sum1"/>
@@ -106,7 +106,7 @@
         </literalExpression>
     </decision>
     <decision name="sublist12" id="_95e0ad53-c08f-46af-baa0-9c36d69002f5">
-        <variable name="sublist12"/>
+        <variable typeRef="tStringList" name="sublist12"/>
         <informationRequirement id="a29945fa-45fe-4464-8d07-3e37c37f7069">
             <requiredInput href="#_86f5e710-a139-4cd5-8ac4-90b4671a75b8"/>
         </informationRequirement>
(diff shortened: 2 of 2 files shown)


opatrascoiu:
When the test was written there was no type lattice or the concept of Any.

I think removing the types is not the best fix. For example, the type for input nodes list1 and list2 is correct, list of strings. 

If a decision returns a list with mixed values, the type of the output variable of that decision should be list<Any>; Any is a valid typename, just like number or string. Can you point to any such decision please?

StrayAlien:
Hi @opatrascoiu, 

Thanks.  Whether the types are valid or not, I'd support removing the types here as the _purpose_ of the test is to test list functions not list types - list types are covered in other tests.  If you can remove something from a test and the test assertions still mean the same thing afterwards, then what you removed was not part of the test.

IMO the less a unit test does to prove its assertions the better.  Things are just clearer then.   This discussion being an example of such I guess.

Re the 'list' type.  Over in https://github.com/dmn-tck/tck/pull/238 I removed `decision_002` because of the comments about what `typeRef` can reference.  `list` is not supported and neither is `Any`.  Despite what we might like it to, the docs (albeit missing 'date'!) do not list Any, list, null, context, or range.  The test was removed because, according to those comments, it is not possible to define an ItemDefinition that supports a list of any.  Re-quoting Matteo's screen shot from that PR:

> and the typeRef attribute would reference the type or element definition in the imported schema. If the type language is FEEL the built-in types are the FEEL built-in data types: number, string, boolean, days and time duration, years and months duration, time, and date and time.

That is what I thought the outcome of that exchange was - neither `Any` nor `list` can be used in a typeRef - hence the test being removed.

You did mention over there you believed 'Any' could be used as a typeRef, but `list` could not.  Is there something in the spec or otherwise to support that ? (no offense - just asking - I would _love_ those types to just be available in typeRef, I think @tarilabs may also! :-) ).

Personally, I think all the built-in types should be available for use in a typeRef, with `list` being list<any> and `context` holding any property.  It just makes more sense that way.

If you / others feel strongly that types are important to these list function tests then the input types as tStringList could go back in without dealing with spec ambiguities for sure.

agilepro:
We believe this will be clarified when RTF makes a clarification on types and Any then we can fix this one correctly.   Marking as waiting on RTF.  type literals issue.

StrayAlien:
Hi @agilepro, thanks.  These tests are for list functions, not the Any or list types.  Might I suggest we close this off as the tests have not changed, nor has the value or meaning of the tests - when we get clarification on Any etc, that can be asserted on in the correct suites.  There are already places where assertions for those can be housed.

agilepro:
@StrayAlien do you mean to pull back / remove this PR?  Please do.   Other option is to leave this here for the future.  Which do you want to do?

StrayAlien:
Hi @agilepro. I mean keep these changes as is and merge them.  The RTF clarification for types is certainly very important, but not to these particular tests.  The types here were incorrect but these tests assert on list functions and thus their assertions are not related to type.  

When we get RTF clarification on Any (etc) then we can make those assertions in their associated suites.

agilepro:
Greg,

We spent quite a while discussing this.  Both Octavian and Matteo (also Simon and me) agree that the typerefs included in the tests are valid and correct (except maybe one, see below).  Since these are valid constructs, we want to keep these tests as they are.   

It may be equally valid to have tests like these without the type references, leveraging Any, however those should be new tests, and not replacing the old ones.

They recognize one element of unclarity and that is the "append" function.  This test is written to take a list of strings, and append onto it, as we new element, a list.   The result is a list where the elements are sometimes strings, and sometimes a list of strings.  

The conclusion was that this test should be changed.  A new data item should be created which is a string value, and the test should append this string value onto the list of strings, and the result will still be a list of strings.   This makes the test uncontroversial by avoiding a list with mixed type elements.

If you would like to propose the modified tests as new tests, I think everyone is OK with that, but we would like to keep the current tests with the one small change mentioned above.

StrayAlien:
I've not taken any action as yet here - but it sounds like you're saying the tests where right - except for the when they were wrong :-) :-)

That's the reason why I submitted this PR - that append makes a list with some of the elements as strings and some as lists - which is not compatible with the tstringlist type there.

The _types_ themselves are, of course, valid types, but the data being produced does not match the type.  Hence this PR.    

I think we've probably made enough of a mountain out of this proverbial molehill - I'll amend the test.

StrayAlien:
@agilepro @tarilabs - amended as per request.  Hopefully I have it right.  I have amended the append() test to add a string to a list of strings.  The `flatten1` test relied on the append test returning a list of (string +  list of string) so I have amended that test to retain the original assertion - otherwise it would have been flattening an already 'flattened' list - which test very much.

No offence to original author(s) but I think this suite needs breaking up into separate suites each for a particular list function covering the usual function scenarios like good/bad params, too little, too many params, badly named params etc.

tarilabs:
Thanks for the updated PR. I believe the currently applied changes make it less controversial.
We can discuss ItemDefinition properties separately.
There can always be additional test, again separately ;)

tarilabs:
@opatrascoiu for me we could merge this one, could you kindly double-check and let me know?

Also if any other has a feedback, welcome

agilepro:
Discussed in the meeting today.   Matteo thinks this is good to go.  Octavian, please let us know if you want to check this.  Let Matteo know if he should merge this or not.

PR #260, opened 2019-03-08
Drools version update, result refresh to include new tests

tarilabs:
/cc @etirelli


agilepro:
leave these to merge by Edson

PR #261, opened 2019-03-08
Drools runner update: infer current version based on Maven pom

tarilabs:
The previous work (and current) behaviour of Drools TCK runner just update the "last refreshed" date and erases "previous version" results directories based on lexicographical ordering of the version.

With this update, Drools version is inferred from the Maven pom and reflected both in the TCK properties file content update and in the directory --moves the directory as the "previous version" TCK properties file contains other description which are to be preserved.

This update does not address the problem of lexicographical ordering of the version which still need to be manually addressed only in those rare occurrences :)

@etirelli @baldimir


agilepro:
leave these to merge by Edson

PR #262, opened 2019-03-09
0074-feel-properties

opatrascoiu:
Added missing error flags and fixed the zone IDs

### DIFF
--- a/TestCases/compliance-level-3/0074-feel-properties/0074-feel-properties-test-01.xml
+++ b/TestCases/compliance-level-3/0074-feel-properties/0074-feel-properties-test-01.xml
@@ -282,7 +282,7 @@
 
     <testCase id="ym_duration_003">
         <description>ym duration - days is null</description>
-        <resultNode name="ym_duration_003" type="decision">
+        <resultNode name="ym_duration_003" type="decision" errorResult="true">
             <expected>
                 <value xsi:nil="true"/>
             </expected>
@@ -291,7 +291,7 @@
 
     <testCase id="ym_duration_004">
         <description>ym duration - hours is null</description>
-        <resultNode name="ym_duration_004" type="decision">
+        <resultNode name="ym_duration_004" type="decision" errorResult="true">
             <expected>
                 <value xsi:nil="true"/>
             </expected>
@@ -300,7 +300,7 @@
 
     <testCase id="ym_duration_005">
         <description>ym duration - minutes is null</description>
-        <resultNode name="ym_duration_005" type="decision">
+        <resultNode name="ym_duration_005" type="decision" errorResult="true">
             <expected>
                 <value xsi:nil="true"/>
             </expected>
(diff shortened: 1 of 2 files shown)


StrayAlien:
Worth noting re `@Etc/Utc` vs `@Etc/UTC` that the dmn spec refers to iana which is not case-sensitive as far as I know.   https://data.iana.org/time-zones/theory.html.   In "Names of timezones":

> Do not use names that differ only in case. Although the reference implementation is case-sensitive, some other implementations are not, and they would mishandle names differing only in case.

In this instance, for example, Java happily accepts both zones, though other languages/temporal libs may not.  Perhaps before we merge this we need to clarify if dmn zones are expected to be case-sensitive?

opatrascoiu:
@StrayAlien I'm afraid Java8 doesn't support Etc/Utc. It throws 

java.time.zone.ZoneRulesException: Unknown time-zone ID: Etc/Utc

On top of that UTC is an acronym, hence Utc is incorrect from the grammatical point of view.

StrayAlien:
Thanks @opatrascoiu.  Re Java: Out of curiosity I did some digging and debugging yesterday to see why I'm good for case insensitive zones and funnily, I _totally_ forgot I had this set:

https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatterBuilder.html#parseCaseInsensitive--

Re the link: yes, they are uppercase, but there is no requirement for impls to be case sensitive was the point I was making.

Re zone case sensitivity on the whole, let's get this in as `UTC` (as I do agree, it is 'nicer' that way), and I'll open an issue here to clarify whether case sensitivity applies.

agilepro:
Matteo t review.   If OK please merge it.

opatrascoiu:
@tarilabs I thought the same until I came across ZoneId.of() :)

StrayAlien:
Thanks @tarilabs.  I mentioned the inbox # over here: https://github.com/dmn-tck/tck/issues/264.  it is INBOX-836 https://issues.omg.org/issues/lists/unclassified#issue-44836

tarilabs:
I am getting notified of commit appended after approvals.
I notice b3a7eb7 and I believe @opatrascoiu you meant to raise that as a separate PR to update jDMN results?

opatrascoiu:
That's correct. I forgot I had this one in play.

agilepro:
Some discussion on how to straighten out the checkins to make this right.

tarilabs:
I will cherry-pick only the relevant commit, which is 954c080 to be merged already.

tarilabs:
I updated this PR with only the relevant commit as agreed during 2019-03-22 DMN TCK meeting and reported in https://github.com/dmn-tck/tck/pull/262#issuecomment-475703172

tarilabs:
I merged this PR after it was updated by cherry-picking only the relevant commit:
```
@opatrascoiu Added missing flag and fixed zone ID 954c080
```
as agreed during 2019-03-22 DMN TCK meeting and reported in #262 (comment)

tarilabs:
Just for completeness, as this was merged [15..end of the month], for Drools this was not a problem either way, and double-check by the latest master CI check:
https://travis-ci.org/dmn-tck/tck/builds/510029645
image

PR #265, opened 2019-03-11
0085-decision-services

StrayAlien:
Re-done DS tests!  I really didn't quite 'get it' last time I submitted tests.  Hopefully this time I do.  :-)   

I have tested for both 'direct' invocation (calling it directly from the test suite like it was a 'microservice'), as well as 'indirect' (which is calling it from another DRG element - in this case, decisions).   The 'direct' tests are fairly separate in case this causes issues.  

... but, I think it is really worth asserting that a DS can be called as a root level element like a decision.  Hopefully that causes no issues with people's tests runners - mine required no alteration so hopefully others are the same.

I have asserted on similar stuff to normal BKM/functions, like correct params, no params, named params, badly-name params, too many params, incorrect type params. etc.

I have also asserted on param _order_ as this comes from the DS inputs, not from formal params.  Positional and named params both get tests to assert params are interprested in the correct order as per "Semantics of Decision Services".

Param types also gets asserted as the formal params for a DS get their types from input decision/data.

I have added two tests to assert no 'bleed through' occurs for DS params and their 'global' counterparts.  That is, even though a DS is invoked with (say) a input decision value, the actual decision can still be used outside of a DS normally.  And vice versa.  These tests are really to assert that a DS 'scopes' its input decisions/data params and they have no influence on the actual decisions/data.  

As this stuff can be confusing to follow through in XML I have kept the tests as simple as I can and completely isolated from each other. 

Comments welcome.  If you can think of any more tests that might be relevant, let me know.

Greg.

PS, thanks @opatrascoiu and @tarilabs for the assistance in getting my head around it. The clincher was realising that `encapsulatedDecisions` represents every decision in the _entire_ DS call tree.

PPS.  We really ought to consider a 'DMN validation' tool to verify correctness beyond just schema conformance.

### DIFF
--- a/TestCases/compliance-level-3/0085-decision-services/0085-decision-services-test-01.xml
+++ b/TestCases/compliance-level-3/0085-decision-services/0085-decision-services-test-01.xml
@@ -1,0 +1,259 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<!-- Contributed to DMN TCK by StrayAlien -->
+<testCases xmlns="http://www.omg.org/spec/DMN/20160719/testcase" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+
+    <modelName>0085-decision-services.dmn</modelName>
+    <labels>
+        <label>Compliance Level 3</label>
+        <label>Decision Services</label>
+    </labels>
+
+    <testCase id="001">
+        <description>Direct invocation: with no params</description>
+        <resultNode name="decisionService_001" type="decisionService">
+            <expected>
+                <component name="decision_001">
+                    <value xsi:type="xsd:string">foo</value>
+                </component>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="002">
+        <description>Direct invocation: with an input decision</description>
+        <!-- we expect the input param value to 'override' the actual impl of 'decision_002_input' -->
+        <inputNode name="decision_002_input">
+            <value xsi:type="xsd:string">baz</value>
+        </inputNode>
+        <resultNode name="decisionService_002" type="decisionService">
+            <expected>
+                <component name="decision_002">
+                    <value xsi:type="xsd:string">foo baz</value>
+                </component>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <!-- note: we can't test invoking decision service directly passing a input data param as the param
+    and the global input data will actually have the same value - so a little difficult to assert where input data value
+    came from - param or global input -->
+
+    <testCase id="002_a">
+        <description>Direct invocation:  with an input decision but supplying no param</description>
+        <!-- requires decision_002_input but we're not providing it-->
+        <resultNode name="decisionService_002" type="decisionService" errorResult="true">
+            <expected>
+                <value xsi:nil="true"/>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="002_b">
+        <description>Direct invocation: with an input decision - but supplying a null value</description>
+        <inputNode name="decision_002_input">
+            <value xsi:nil="true"/>
+        </inputNode>
+        <!-- errorResult is true because of the string + null operation -->
+        <resultNode name="decisionService_002" type="decisionService" errorResult="true">
+            <expected>
+                <component name="decision_002">
+                    <value xsi:nil="true"/>
+                </component>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="002_c">
+        <description>Direct invocation: with an input decision - but supplying an incorrect param type</description>
+        <!-- decision_002_input is a string type but we're providing a number so we expect it to be null-coerced -->
+        <inputNode name="decision_002_input">
+            <value xsi:type="xsd:decimal">1234</value>
+        </inputNode>
+        <resultNode name="decisionService_002" type="decisionService" errorResult="true">
+            <expected>
+                <component name="decision_002">
+                    <value xsi:nil="true"/>
+                </component>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="003">
+        <description>Direct invocation: with two input decisions and an input data</description>
+        <!-- positive test -->
+        <inputNode name="decision_003_input_1">
+            <value xsi:type="xsd:string">B</value>
+        </inputNode>
+        <inputNode name="decision_003_input_2">
+            <value xsi:type="xsd:string">C</value>
+        </inputNode>
+        <inputNode name="inputData_003">
+            <value xsi:type="xsd:string">D</value>
+        </inputNode>
+        <resultNode name="decisionService_003" type="decisionService">
+            <expected>
+                <component name="decision_003">
+                    <value xsi:type="xsd:string">A B C D</value>
+                </component>
+            </expected>
+        </resultNode>
+    </testCase>
+
+
+    <testCase id="004">
+        <description>with no params</description>
+        <!-- positive test -->
+        <resultNode name="decision_004_1" type="decision">
+            <expected>
+                <component name="decision_004_2">
+                    <value xsi:type="xsd:string">foo</value>
+                </component>
+            </expected>
+        </resultNode>
+    </testCase>
+
+
+    <testCase id="005">
+        <description>passing a param when DS has no params</description>
+        <resultNode name="decision_005_1" type="decision" errorResult="true">
+            <expected>
+                <value xsi:nil="true"/>
+            </expected>
+        </resultNode>
+    </testCase>
+
+
+    <testCase id="006">
+        <description>passing a single input decision param</description>
+        <!-- positive test -->
+        <resultNode name="decision_006_1" type="decision">
+            <expected>
+                <component name="decision_006_2">
+                    <value xsi:type="xsd:string">foo bar</value>
+                </component>
+            </expected>
+        </resultNode>
+    </testCase>
+
+
+    <testCase id="007">
+        <description>passing a single input decision param with incorrect type</description>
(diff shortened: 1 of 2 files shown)


agilepro:
Looks like this needs discussion.

opatrascoiu:
I've started to look at them. Here are my comments so far:
- missing typeRef for inputData_003. The typRefs are optional but a good model should contain the typeRefs at least for the inputs.
- tests 001-003 (the direct calls) AFAIK decision services are invocables, hence the associated variable is not bound to the result of the invocation but to the definition of the function -  they should be invoked via XXX(...) to obtain. The decisions are different, the associated variable is bound to the output of the decision not to definition of the decision. We might need a different type of node to assert expected values for BKMs & DSs.

Will look at the other tests later.

StrayAlien:
Thank @opatrascoiu, @agilepro   Appreciated. Hopefully they are not too confusing - there is a chance of that.

Re typeRefs, I have added that one, but on the whole, for tests, I just try to keep the test as _lean_ as possible.  I absolutely do agree a good model has typeRefs all over, but IMO a test model is not necessarily a good model - it just has to serve the test and be clear what it is trying to achieve. 

Re the direct and indirect thing.  I figured this might spark some conversation which is cool.

My thinking goes like this:

If a DS is called externally, it'll not be invoked like `foo_ds()`, it'll likely be invoked passing the input data and the target DRG element id (or name?).  Yes, RPC interface could implemented, but I doubt the spec writers are thinking about that.  I am thinking http POST to a DMN microservice here.

So, to support this a DMN runtime will likely need to be able to take a 'target' DRG and input data and _itself_ invoke the DS passing the correct params in the correct order. 

To that end, the 'direct' tests are simply saying that: 'here is some data and the target DRG I want invoked - you do it'.  Test runner impls can do what they want and marshall/coerce/arrange params as they see fit.

Which is exactly what the test runners assume now.  Target DRG and some input data.  

Keep in mind that the test XML is just an abstract way of providing some input data and a target and defining expected outcomes - it is not DMN, and therefore does not have to support DMN invocation semantics.

So in that sense, I don't think we need anything special in the test runner or test syntax.  I think it is just the same as we do it now which is 'given a target DRG, here is some input data and, please DMN runtime, invoke it and give me the result'.

I don't think it need to be any more complicated than that - but sure happy to hear people's thoughts.

Greg.

StrayAlien:
As a note: the coercion tests PR has also been updated with new tests to cover null coercion and singleton list coercion for decision services params.  https://github.com/dmn-tck/tck/pull/238

opatrascoiu:
Not sure how much value the direct calls of DS add from the testing perspective, as we already have tests for DS invocation (see tests when invocation is inside decisions) 

I just had a look at the XSD schema for Nodes. The resultNode has an attribute called type that is meant to describe the DMN element type. So far on our tests we have only "decision".

That means that we don't need to change the schema if we decide to support resultNodes for decision services. We probably need to agree on the operational semantics (what the runner should do) of these nodes: resultNodes with type="decisionService". For example, we could have:
1. lookup DS definition
2. evaluate arguments
3. build a function invocation
4. evaluate function

Lets discuss.

StrayAlien:
Hi @opatrascoiu.  Yes -  I have used that 'type' on the result node here for 'decisionService' - it seem to fit nicely. 

I'm happy to get input here, I did the 'direct' thing as a 'throw it out there and see what people think' kind of thing.  I think the value of the 'direct' tests is in testing the invocation a DRG element _outside_ general model execution - in a way, like a microservice may .. and that is the point of DSs.

As it happens, I didn't have to change anything in our test runner to support this as it doesn't assume what type of DRG element is being invoked - it just gathers info from the test being executed and hands that off to the DMN runtime 'front door' where it is dealt with as a 'data plus DRG target' invocation.   But just looking now at some of the runners in the TCK repo they look like they hard code that a test is only for a Decision.

But, happy to be guided here.  If others do not feel the same I'll move them into my own non-TCK test suite.  No problems.

Thanks again,

Greg.

tarilabs:
Before I'll dig into the test further, I believe there are a few problem with the DMN model per-se.

Decision Service is an Invocable, and similar to BKM it must have a variable.
Decision Service nodes from 'decisionService_001' to 'decisionService_014' are missing the variable declaration

StrayAlien:
Thanks @tarilabs . Actually, according to the 1.2 XSD they are optional and so I have always assumed they may not be there - regardless of what the spec says:

```
	<xsd:complexType name="tInvocable">
		<xsd:complexContent>
			<xsd:extension base="tDRGElement">
				<xsd:sequence>
					<xsd:element name="variable" type="tInformationItem" minOccurs="0"/>
				</xsd:sequence>
			</xsd:extension>
		</xsd:complexContent>
	</xsd:complexType>
```
But strictly according to the spec you are correct. I'll amend.  Thanks.

StrayAlien:
Hey all,

(@agilepro @opatrascoiu @tarilabs )

From the weekly meeting notes, folks may have some questions re tests here.  Feel free to contact me directly if you like.  Keith has my email or you can fish it from git.

I’ve tried to keep these DS tests small and discrete with some notes each, but this, and imports, can be _very_ confusing so no stress. 

Taking a peek into the failing RH and camunda CI jobs - it seems camunda is failing basically everything - maybe they are not yet supporting 1.2 models? 

The RH CI does show some errors here - some of which seems runner related - the others seems to be (maybe) returning nulls as DS context value results.  But the CI also seems to show a lot of other failing tests (if I read correctly) - stuff like: 

```
3:12:57.229 [main] INFO  o.o.d.t.runner.drools.DroolsTCKTest.executeTest:205 - Executing test '0076-feel-external-java-test-01 / literal_007_a'
03:12:57.230 [main] ERROR o.k.d.f.r.functions.BaseFEELFunction.invokeReflectively:152 - Error trying to call function <anonymous>.
java.lang.IllegalArgumentException: Unable to coerce parameter c1. Expected char but found class java.lang.String
```
That test suite is already in master - so maybe the latest RH code is not running in CI?

At any rate, again, feel free to contact me.

agilepro:
These are still failing in RedHat.  Seems like the tests are legit.   Matteo would like some time to discover what the source of the failure is.  I might be a runner problem.

StrayAlien:
@agilepro @tarilabs - I still have some more tests to do for decision services .. more weirdness of paths and inputs etc .. but I am happy to hold off on more tests if it looks like these will get merged.

tarilabs:
https://github.com/dmn-tck/tck/pull/265#issuecomment-480345580
> I might be a runner problem.

I stand by the opinion it is indeed a runner problem, specifically in the TCK test file XSD "type" attribute for `resultNode` is a (free text) `string`:
https://github.com/dmn-tck/tck/blob/8b543ff6f6ee22be8cb7054f19d151c53e70b246/TestCases/testCases.xsd#L42
and not the `testCaseType` enumeration:
https://github.com/dmn-tck/tck/blob/8b543ff6f6ee22be8cb7054f19d151c53e70b246/TestCases/testCases.xsd#L4
which enlist for instance `decision`, `bkm` (and naturally to be augmented to a value for `decisionService`, for instance).

Additionally, imho, is the `testCase` which should be of type "decision service" (not resultNode).
This way, each inputNode is the input parameter of the Decision Service, resultNode are the components of the Decision Service's output decisions.

I'll try to use this PR as a baseline to show with further commits my pragmatics change proposal, but for now I wanted to backup with facts the reported quote which pointed at my name.

agilepro:
Matteo is interested in working on this.  The TCK API needs to be altered to be fully enabled for decision services.  Might require an XSD change.  Not a big change.  We need something that clearly indicates a decision service.   The decision service is a subset of the decision model where a particular result comes back from a particular (subset) of the input.

tarilabs:
> The TCK API needs to be altered to be fully enabled for decision services. Might require an XSD change.

Indeed only small XSD change to make clear is invoking a decision service; hopefully there is no API change needed, but to be confirmed.

tarilabs:
### Changes to TCK XSD

I have raised a PR for TCK XSD Schema changes to support Decision Services, here: https://github.com/dmn-tck/tck/pull/308

I have raised a PR for @StrayAlien to eventually accept, which will align this original PR from @StrayAlien , to the aforementioned changes: https://github.com/StrayAlien/tck/pull/1

### Wrong test result

I do NOT concur with the result of test case `010`, from the spec "10.3.2.13.4 Positional and named parameters":

> If named, unsupplied parameters are bound to null.

hence expression:
```
decision_010_1 : decisionService_010(foo: "bar")
decision_010_1 : decisionService_010(null)
decision_010_1 : {
        decision_010_2: "foo " + null
}

final result:

decision_010_1 : {
        decision_010_2: null
}
```

and not just null.

I have NOT modified @StrayAlien PR to resolve on this disagreement.

tarilabs:
There is also a repetition of outputDecision as encapsulatedDecision which I believe is not necessary, throughout the decision service defined --I mean you can, if you really want by the way this model was designed, but imho is not necessary the encapsulatedDecision _when is already an outputDecision_; you can just have said decision as an outputDecision.

agilepro:
Please move forward with what you have proposed here.

tarilabs:
> Please move forward with what you have proposed here.

As @StrayAlien did not merge PR https://github.com/StrayAlien/tck/pull/1 I am proceeded by aligning to TCK XSD schema change, by backoffice as decision in TCK meeting.

TestCases/compliance-level-3/0085-decision-services/0085-decision-services-test-01.xml
```diff
@@ -0,0 +1,249 @@
...
+            </expected>
+        </resultNode>
+    </testCase>
+
+
+    <testCase id="010">
+        <description>passing badly named param</description>
+        <resultNode name="decision_010_1" type="decision" errorResult="true">
+            <expected>
+                <value xsi:nil="true"/>
```
tarilabs: Ref: https://github.com/dmn-tck/tck/pull/265#issuecomment-493759855

```suggestion
                <component name="decision_010_2">
                    <value xsi:nil="true"/>
                </component>
```

PR #266, opened 2019-03-19
Update Trisotech Results to version 6.1.22

SimonRinguette:
I don't see new test cases, i've only updated the version/date


tarilabs:
I believe only new tests are "properties" as per commit https://github.com/dmn-tck/tck/commit/e670929e39b4b9e5edbaa69ea12e8d65f734799c and you had them included already with the deferred Feb refresh ;)

PR #271, opened 2019-03-23
0087-chapter-11-example

StrayAlien:
The chapter 11 example from the 1.2 spec.  @tarilabs gave the thumbs up for the idea to include it in the TCK.  Also thanks to same for info on the correctness of the test.  The spec example feeds incorrect data into the model.

This is as per the spec with a few small changes:

* typeRefs now have had "feel:" removed from them.
* The two decision services now have `variable` child elements.  They were missing.
* The `import` element now has a `locationURI` that specifies the name of the dmn being imported.  In this case it is called 'financial.dmn'.

Comments welcome.

Greg.


agilepro:
Are there any issues on this?   If not, we should merge.

StrayAlien:
@agilepro it has an import element. That is a bit in flux at the moment.

tarilabs:
Reopening.

I believe there is merit in the original idea to put the DMN specification CH11 example under test.

I believe the pragmatical obstacles were: using DMN Import, using DMN decision service in the tests.

Commit 1a604ae propose a pragmatical change where the use of those DMN feature is deferred, and simply amend the original file to use directly decisions (instead of decision services) and avoid the need of DMN Import (which can be avoided by simply using a orphaned BKM in the original file).

For me with 1a604ae this PR is good to be merged.
Once DMN Import, DMN decision service usage in TCK is stable we can always come back and augment this test case too.

StrayAlien:
Hi @tarilabs - No real issues here but I wasn't aware that decision services were an obstacle - just the import - which is why I removed the test.  

If bringing the test back any reason to remove the decision services?  There has been no discussion on the decision services PR noting anything in particular and its not waiting on RTF.  Last comment there was the tests looked legit.  Or is it more to do with the runner than the tests?

As a side note - it'd be great if we could move these types of 'test' to their own folder as per: https://github.com/dmn-tck/tck/issues/293.  It is not really a 'test'.

StrayAlien:
just saw your comment over on decision services PR.  Fair enough.

tarilabs:
>  It is not really a 'test'.

I don't concur with this conclusion for the reason @StrayAlien always invite people to submit "real" DMN models. imho it doesn't get any better than putting under the test suite the one and only example from the DMN specification itself! ;)

I appreciate however some of the reasoning behind the linked issue, so for this specific case I thought worthy to add fine-grainer test to facilitate implementor with 194a070

StrayAlien:
Hey @tarilabs, don't get me wrong. I think there is great value in 'whole' models. For sure.  I submitted this PR initially!

I guess it is more about the focus of tests. BTW, I didn't anticipate any action on it here right now! Just noting.

I think this is a worthy inclusion in the suite as a pretty complex example. Hey it even has a decision in it with no logic. Who would have though that could be so! That will become its own test eventually. 😁

PR #272, opened 2019-03-23
0086-imports

StrayAlien:
All, some initial tests for imports covering much of the basics - with some notable exceptions that I will raise an issue on - namely on importing 'InputData'.

Apols in advance for the lots-of-words here.

In reading these tests, or indeed, trying to get to grips with the imports spec there is one key factor in it all and it is important I think to understand this from the start:

> the spec only defines uniqueness to _within_ a given model, not _across_ models.  

Such that:

* id's are unique to a model only.  So imported models can have elements with the same ids as the importing model or other sibling imported models or their own imports. 
* names are unique to model only, so same applies as with ids.

And _very_ importantly I feel:

> an import element namespace and name are only relevant _inside_ the model it is defined in.
  
That is, say, a model `A.dmn` can have an import element like the following to import (say) model `B.dmn`:

`<import namespace="http://foo" name="i" locationURI="B.dmn" importType="...etc"/>`

And the `namespace` and `name` are what model A uses to refer to stuff in model B. 

**however**, model `B.dmn` is permitted to have the following to import (say) `C.dmn`:

`<import namespace="http://foo" name="i" locationURI="C.dmn" importType="...etc"/>`

And the `namespace` and `name` are what model B uses to refer to stuff in model C.

But note that namespace/name they may also be the _same_ amongst models - so they are also 'model local'.  The spec says an import namespace - _should_ be globally unique, but there is no way to guarantee it of course.  Actually, the spec says in "12.3.1 Document Structure":

> "Each file SHALL declare a “namespace” that MAY differ between multiple files of one model."

So, treating the namespace of a dmn as unique across models is not a great idea.

Also important to note is the `locationURI` property on the import element is optional - but, for the sake of these tests, imports use a locationURI.

So, taking all the above into account, these tests use duplicate name in models and also duplicate import namespace and names.  This is not to complicate things, but it purposeful in order to help flesh out issues with how a runtime may identify and use drg elements.

The thing that is not tested here as yet is how to deal with InputData in imported models, and the InputData from their imported models, and their imports in turn and so on.   I'll raise an issue to discuss.

All comments welcome.

### DIFF
--- a/TestCases/compliance-level-3/0086-imports/0086-imports-02.dmn
+++ b/TestCases/compliance-level-3/0086-imports/0086-imports-02.dmn
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<definitions namespace="http://www.montera.com.au/spec/DMN/0086-imports"
+             name="0085-decision-services"
+             id="_i9fboPUUEeesLuP4RHs4vA"
+             xmlns="http://www.omg.org/spec/DMN/20180521/MODEL/"
+             xmlns:di="http://www.omg.org/spec/DMN/20180521/DI/"
+             xmlns:dmndi="http://www.omg.org/spec/DMN/20180521/DMNDI/"
+             xmlns:dc="http://www.omg.org/spec/DMN/20180521/DC/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+             xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+
+    <description>Imports 01</description>
+
+    <import name="i"
+            namespace="http://montera.com.au/imports"
+            locationURI="02/import1.dmn"
+            importType="http://www.omg.org/spec/DMN/20180521/MODEL/"/>
+
+    <decision name="decision_001" id="_decision_001">
+        <variable name="decision_001" typeRef="string"/>
+        <informationRequirement>
+            <requiredDecision href="http://montera.com.au/imports#_decision_001"/>
+        </informationRequirement>
+        <literalExpression>
+            <text>"0 -> " + i.decision_001</text>
+        </literalExpression>
+    </decision>
+
+</definitions>
--- a/TestCases/compliance-level-3/0086-imports/0086-imports-test-01.xml
+++ b/TestCases/compliance-level-3/0086-imports/0086-imports-test-01.xml
@@ -1,0 +1,115 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<!-- Contributed to DMN TCK by StrayAlien -->
+<testCases xmlns="http://www.omg.org/spec/DMN/20160719/testcase"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+
+    <modelName>0086-imports.dmn</modelName>
+    <labels>
+        <label>Compliance Level 3</label>
+        <label>Imports</label>
+    </labels>
+
+    <testCase id="001">
+        <description>Decision can reference an imported decision</description>
+        <resultNode name="decision_001" type="decision">
+            <expected>
+                <value xsi:type="xsd:string">decision_001 in import1</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="002">
+        <description>Decision can reference an imported BKM</description>
+        <resultNode name="decision_002" type="decision">
+            <expected>
+                <value xsi:type="xsd:string">foo</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+<!--    <testCase id="003">
+        <description>Reference an imported inputData</description>
+        <inputNode name="import1.inputData_001">
+            <value xsi:type="xsd:string">foo</value>
+        </inputNode>
+        <resultNode name="decision_003" type="decision">
+            <expected>
+                <value xsi:type="xsd:string">foo</value>
+            </expected>
+        </resultNode>
+    </testCase>-->
+
+    <testCase id="004">
+        <description>Decision can reference an imported decisionService</description>
+        <!-- note that the imported DS uses a outputDecision ref for an id that exists both in the
+        imported AND importing model.  We expect the DS to use the 'local' decision  -->
+        <resultNode name="decision_004" type="decision">
+            <expected>
+                <component name="decision_001">
+                    <value xsi:type="xsd:string">decision_001 in import1</value>
+                </component>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="005">
+        <description>Decision can reference an imported decisionService passing param by name/positional</description>
+        <resultNode name="decision_005" type="decision">
+            <expected>
+                <component name="i.decision_002">
+                    <value xsi:type="xsd:string">foo</value>
+                </component>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="006">
+        <description>BKM can reference an imported BKM</description>
+        <resultNode name="decision_006" type="decision">
+            <expected>
+                <value xsi:type="xsd:string">foo</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="007">
+        <description>BKM can reference an imported Decision Service</description>
+        <resultNode name="decision_007" type="decision">
+            <expected>
+                <value xsi:type="xsd:string">decision_001 in import1</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="008">
+        <description>Decision can reference an imported typeRef</description>
+        <!-- imported type is a string list and the decision returns a string - so we're
+        expecting coercion -->
+        <resultNode name="decision_008" type="decision">
+            <expected>
+                <list>
+                    <item>
+                        <value xsi:type="xsd:string">foo</value>
+                    </item>
+                </list>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="009">
+        <description>Item definition can reference an imported typeRef</description>
+        <!-- imported type is a string list and the decision returns a string - so we're
+        expecting coercion -->
+        <resultNode name="decision_009" type="decision">
+            <expected>
+                <list>
+                    <item>
+                        <value xsi:type="xsd:string">foo</value>
+                    </item>
+                </list>
+            </expected>
+        </resultNode>
+    </testCase>
+
+</testCases>
--- a/TestCases/compliance-level-3/0086-imports/0086-imports-test-02.xml
+++ b/TestCases/compliance-level-3/0086-imports/0086-imports-test-02.xml
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<!-- Contributed to DMN TCK by StrayAlien -->
+<testCases xmlns="http://www.omg.org/spec/DMN/20160719/testcase"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+
+    <modelName>0086-imports-02.dmn</modelName>
+    <labels>
+        <label>Compliance Level 3</label>
+        <label>Imports</label>
+    </labels>
+
+    <testCase id="001">
+        <description>Deeply reference imported decisions</description>
+        <resultNode name="decision_001" type="decision">
+            <expected>
+                <value xsi:type="xsd:string">0 -> i1 -> i2 -> i3</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+
+</testCases>
--- a/TestCases/compliance-level-3/0086-imports/0086-imports.dmn
+++ b/TestCases/compliance-level-3/0086-imports/0086-imports.dmn
@@ -1,0 +1,179 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<definitions namespace="http://www.montera.com.au/spec/DMN/0086-imports"
+             name="0085-decision-services"
+             id="_i9fboPUUEeesLuP4RHs4vA"
+             xmlns="http://www.omg.org/spec/DMN/20180521/MODEL/"
+             xmlns:di="http://www.omg.org/spec/DMN/20180521/DI/"
+             xmlns:dmndi="http://www.omg.org/spec/DMN/20180521/DMNDI/"
+             xmlns:dc="http://www.omg.org/spec/DMN/20180521/DC/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+             xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+
+    <description>Imports 01</description>
+
+    <import name="i"
+            namespace="http://montera.com.au/imports"
+            locationURI="01/import1.dmn"
+            importType="http://www.omg.org/spec/DMN/20180521/MODEL/"/>
+
+    <itemDefinition name="fooType" id="_fooType">
+        <typeRef>i.tStringList</typeRef>
+    </itemDefinition>
+
+    <!-- A decision using an imported decision -->
+
+    <decision name="decision_001" id="_decision_001">
+        <variable name="decision_001" typeRef="string"/>
+        <informationRequirement>
+            <requiredDecision href="http://montera.com.au/imports#_decision_001"/>
+        </informationRequirement>
+        <literalExpression>
+            <text>i.decision_001</text>
+        </literalExpression>
+    </decision>
+
+    <!-- *************************** -->
+
+    <!-- A decision using an imported BKM -->
+
+    <decision name="decision_002" id="_decision_002">
+        <variable name="decision_002" typeRef="string"/>
+        <knowledgeRequirement>
+            <requiredKnowledge href="http://montera.com.au/imports#_bkm_001"/>
+        </knowledgeRequirement>
+        <literalExpression>
+            <text>i.bkm_001("foo")</text>
+        </literalExpression>
+    </decision>
+
+    <!-- *************************** -->
+
+    <!-- A decision using an imported InputData -->
+
+<!--
+    <decision name="decision_003" id="_decision_003">
+        <variable name="decision_003" typeRef="string"/>
+        <informationRequirement>
+            <requiredInput href="http://montera.com.au/imports#_inputData_001"/>
+        </informationRequirement>
+        <literalExpression>
+            <text>i.inputData_001</text>
+        </literalExpression>
+    </decision>
+-->
+
+    <!-- *************************** -->
+
+    <!-- A decision using an imported Decision Service -->
+
+    <decision name="decision_004" id="_decision_004">
+        <variable name="decision_004"/>
+        <knowledgeRequirement>
+            <requiredKnowledge href="http://montera.com.au/imports#_decisionService_001"/>
+        </knowledgeRequirement>
+        <literalExpression>
+            <text>i.decisionService_001()</text>
+        </literalExpression>
+    </decision>
+
+    <!-- *************************** -->
+
+    <!-- A DS using an imported decision pass a param by name -->
+
+    <decisionService name="decisionService_005" id="_decisionService_005">
+        <variable name="decisionService_005"/>
+        <outputDecision href="http://montera.com.au/imports#_decision_002"/>
+        <encapsulatedDecision href="http://montera.com.au/imports#_decision_002"/>
+        <inputDecision href="http://montera.com.au/imports#_decision_003"/>
+    </decisionService>
+
+    <decision name="decision_005" id="_decision_005">
+        <variable name="decision_005"/>
+        <knowledgeRequirement>
+            <requiredKnowledge href="#_decisionService_005"/>
+        </knowledgeRequirement>
+        <literalExpression>
+            <text>decisionService_005(i.decision_003: "foo")</text>
+        </literalExpression>
+    </decision>
+
+    <!-- *************************** -->
+
+    <!-- a BKM that using an imported BKM -->
+
+    <businessKnowledgeModel name="bkm_006" id="_bkm_006">
+        <variable name="bkm_006" typeRef="string"/>
+        <encapsulatedLogic>
+            <formalParameter typeRef="string" name="val"/>
+            <literalExpression>
+                <text>i.bkm_001(val)</text>
+            </literalExpression>
+        </encapsulatedLogic>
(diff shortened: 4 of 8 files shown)


StrayAlien:
No tests as yet for InputData.  I have opened issue : https://github.com/dmn-tck/tck/issues/274 to discuss.

tarilabs:
Greg, as always thanks for these tests.

While I agree with most of your statements, as indeed extracted from the DMN spec, I believe there are some impacts which go _beyond_ what specified in the DMN spec if we mix several model (files) within the same name/namespace in a single TCK test case _folder_.

Just to mention one, the DMN spec doesn't go into the detail on how to call a specific DMN model for evaluation, at API level, and if you mix several model (files) having the same name/namespace then how could you distinguish which one specifically you want to evaluate?
I believe in your PR, you are taking the assumption the model **file** as indicated in the TCK test case xml file, but the DMN spec doesn't specify a file for API is required. So this assumption I believe is beyond the spec.
Similarly, I believe in your PR you are taking the assumption the optional attribute `locationURI` could be used by an implementation to locate the required import-ed models. However as you very well know, this is an optional attribute and relying on this for the TCK test infra is going beyond the spec. Moreover an URI is not an URL so there is nothing in the DMN spec which could help us in relying on this for actually fully _locating_ it.

On other hand, I believe a pragmatical approach is to contain within a given TCK test case directory only the models under test and its necessary imports, and keep namespace unique within such models.
This way, we don't have to rely on the optional `locationURI` to identify which import is being dragged in.
This also allows for properly augmenting the TCK API with providing the runners the additional model `URL`s, and keep a consistent approach where we supply to vendor engine the model under test, as a URL, and additional models which are required, again as a collection of URLs.
Indeed this goes under the assumption that additional model are supplied within the same directory, but I believe it is consistent with the requirement that the model under test is under said directory.
Ref https://github.com/dmn-tck/tck/pull/291

We can use your import test by separating them into consistent TCK test directories --by consistent I mean by avoiding having two models with the same namespace/name as this would be impacted by the problems of "beyond spec" which I tried to highlight above.

What do you think?

StrayAlien:
Hi @tarilabs,

Thanks for the comments.  The spec is so vague here and has no 'model loading api' such that we're going to run into 'it's not in the spec' issues which ever way we turn.  So, ouch!  But, I think it is about choosing the path that offers most benefit and is more akin to the 'decision as a service' philosophy at play here.  Let me explain (and hopefully I have understood your comments!):

We're all aiming for the same thing here so please take my comments as constructive.

Correct, the spec does say the location URI is optional, so, I agree, the truth is we should not rely on it.

But equally, any *other* mechanism employed is _also_ not in the spec and therefore cannot be relied on either.  So we're a bit stuck. 

If I understand (please correct me if I am wrong) you are promoting a means of locating imported models that is also not in the spec - by somehow the test runner locating the models and handing a list of URLs to a runtime? Maybe including namespace or something in the URL?  

Any way we go here we're doing something 'outside' the spec. 

It is important to keep in mind that a test runner is just a test runner - each vendor may have their own way to load tests and run them (as I do).  Any discussion about the test xml and how import models are located really must exclude the way any particular test runner operates.  Given a model a runtime should run it.  And a runtime, in a microservice world, likely will not have access to a single folder where everything it needs lives.

The spec does not say that a runtime model has to share or otherwise the namespace with models it imports or that the 'name' attribute on import elements is unique among a group of models or that an absent locationURI means that all models are in the same 'folder'. In fact the notion that a model's imported models exist in the same 'folder' together implies they have been 'authored' together maybe by the same tool and deployed together - and this, I feel, is _exactly_ the opposite to the standardised DMN 'decision as a microservice' goal this spec is about.  It sounds more like 2000's J2EE 'bundle, jar, deploy' monolithic thinking than lightweight, fast moving ,interconnected, interoperable, multi-vendor decision models. 

The spec doesn't say that a runtime will be handed a list of URLs of models it resolves via namespace, or something.

In a microservice word, what is a 'folder'? In an enterprise with separate libraries of DMN models from 5 different DMN vendors across 3 different cloud providers what is a 'folder' and how to you enforce namespaces are unique when you import another vendor's model?  I don't think it is reality.

I do believe the locationURI approach is, perhaps, the lesser of the 'non spec' evils. It is the only approach that has has no 'magic' in it and is consistent and simple enough to implement.

The spec offers nothing that says _anything_ is unique across models.  It only defines uniqueness _within_ a model.  So, by implication, uniqueness has to be 'external' to a model.  The truth is, the only reliable unique thing we could possibly have is the 'location' of a model and given that element ids are unique to a model, that also serves to uniquely identify an element within any group of models.  The global uniqueness of an element is not the namespace/id, nor the import name/id, it can only be the location/id combination.  That is the _only_ possible thing that can define uniqueness.  How vendor chooses to implement 'location' in their runtime is their concern, but even if you choose to use 'namespace' then you're still treating it as a 'location' identifier - albeit not a unique one - a flawed one.

> the DMN spec doesn't go into the detail on how to call a specific DMN model for evaluation, at API level, and if you mix several model (files) having the same name/namespace then how could you distinguish which one specifically you want to evaluate?

This is life without a locationURI - it all gets confusing then and goes off into the realms of magic and _assumptions_ about uniqueness relying on not-guaranteed-to-be-unique identifiers. I think the absence of locationURI on an import should be considered an edge case meant to cater for some vendor-specific circumstances - not the other way around.  I think usage of location URI is the _mainstream_ use case. 

So, I understand that locationURI is optional - but we have to create tests for this stuff somehow.  Every possible solution here is outside the spec, but I think using locationURI for the tests is the best and most beneficial way to do it.

The tests here do not define a URI scheme, they are just a relative path - so we're not enforcing http:// or ftp:// or file:// - just a relative path. 

Given a URI to a model, a vendor can resolve paths to imports and imports of import (regardless of the test runner)  It is how the world works these days.

Sorry for the long reply - but I did think about it very carefully.    

All comments welcome.

Greg.

tarilabs:
> If I understand (please correct me if I am wrong) you are promoting a means of locating imported models that is also not in the spec - by somehow the test runner locating the models and handing a list of URLs to a runtime? Maybe including namespace or something in the URL? Any way we go here we're doing something 'outside' the spec.

Sorry I disagree.

This is the part were we specified how the TCK infrastructure works.

We have a folder structure and people can rely on that if they want to supply results, manually.

We _also_ have an API, and we have agreed to provide an URL (not an URI) of the model to execute, and the TCK test case file/data. I am proposing to extend this API, by providing directly the additional impored model URLs (again, not URI). To avoid inconsistencies this requires for model namespace/name to be unique without a given TCK folder --this is a requirement too if people are not planning to use the API of the runner and simply rely on the TCK folders, as locationURI in my view is not enough to discriminate otherwise.

I hope I expressed better this time?

StrayAlien:
Thanks @tarilabs,

Cool - I understand the 'pass URLS to the runtime (runner)' API thing.  Thanks for the clarification.

Though, we're maybe still in the same place I think.  In that - "any way we go we're doing something outside the spec".

As an effort to not use locationURI I guess this change also involves removing the locationURI from the PRs tests imports elements and then having the namespace become an import model 'identifier' (hence the request for uniqueness) - then matching something (the namespace) _inside_ another model file?  If so, that is then simply overloading the import namespace to now be the identifier as well as the namespace.  Without a location, _something_ must link the given URL with an import ...

My reading of the spec is that the import namespace attribute is not an external identifier of any sort at all.  It is a model _local_ identifier and only has relevance inside the model that has defined the import.  The spec does not say it is a reference to anything external and indeed it actually says it cannot be:

> "Each file SHALL declare a “namespace” that MAY differ between multiple files of one model."

Unfortunately, "MAY", not "SHALL".  In a multi-vendor scenario - one simply cannot enforce "SHALL", so it is correct.

The PR tests do not assume an import namespace is an identifier for anything other than the current model referencing things for itself.  I believe that is the correct usage.  The tests do this on purpose so that basically, if you get the import tests running you can be pretty confident you're handling imports and namespaces and element ids correctly.  That is exactly why the tests are structured as such.

Using the import namespace element as anything other than a model-local means to specify an import href I think is outside the spec and we should not expect vendors to support that.  

Re:

>  To avoid inconsistencies this requires for model namespace/name to be unique without a given TCK folder

(within?)

and 

> this is a requirement too if people are not planning to use the API of the runner and simply rely on the TCK folders, as locationURI in my view is not enough to discriminate otherwise.

I can attest that the locations are enough to discriminate and namespaces do not have to be unique. Our test runner required no modifications to run the test with imports - and duplicate namespaces and identifiers are handled.

SIDEBAR: I stumbled and fell a number times involving all manner of complexity until I actually just implemented the spec and permitted duplicates - and the impl is the simplest of them all with the greatest flexibility.  

Absolutely this is not an easy part of the spec, and somehow with the vagueness of the spec here "we're damned" no matter how we go about it.  But, I believe the current tests are within the spec - consider them tests for using locationURI if you like - but using import namespace as an external identifier I believe is just against the spec and likely a little dangerous.

As always, happy for comments, and please understand, none of this is 'said in anger', just working towards a common goal here.

Greg.

StrayAlien:
Hey @tarilabs,

Just combing the spec looking at all relevant occurrences of 'namespace'.  Trying to make some concrete sense of it all.  It seems to me quite some confusion arises because of the overloading of the term 'namespace'. 

I think we need to get some stuff clarified as things are all over the place (IMO).

Like, we are given that a Definitions has a 'namespace' but it doesn't say it has to be unique, nor are we given what it is to be used for.  It does not give us specifically that the namespace is related to visibility, nor that it is to be used for imports.

> An instance of Definitions has a namespace, which is a String. The namespace identifies the default target namespace for the elements in the Definitions and follows the convention established by XML Schema.

My feeling is that maybe the definitions namespace is like a 'default import' such that if a reference within the model uses that namespace then it just means 'in this model' - which would be the same as not using it at all.  So, it seems a bit useless in that context. Unsure.  

We're given that an import has a 'namespace' which is to be used in references, but that 'namespace' equally can be used for non-DMN imports like XML schemas and so on.  We're not given that the 'namespace' for a DMN import relates to the 'namespace' of the imported model in any way.

We're given that an import namespace 'should be globally unique', but 'should' is not SHALL and 'globally' is pretty ill-defined as well.

We're also given that Definition namespaces within a set or related model files "MAY differ between multiple files of one model", which implies they may also be the same.   Funnily, "MAY differ between multiple files of one model" makes it almost feel like them being the same is the _default_ expectation - it sounds like a kind of warning that they may differ.

Then, additionally, in '12.3.2 References within the DMN XSD' we get an example like:

```
<requiredDecision
href=”http://www.example.org/Definitions01.xml#prebureauriskDec01”/>
```

Which it says is valid ... and does not use an import.  But it'll have no prefix 'name' so cannot be used in an expression like `<name>.prebureauriskDec01`. Hmm.

But, an ItemDefinition MUST have an import when coming from an external document ...

> An ItemDefinition element MAY have a typeRef, which is a string that references, as a qualified name, either an ItemDefinition in the current instance of Definitions or a built-in type in the specified typeLanguage or a type defined in an imported DMN, XSD, or other document. In the latter case, the external document SHALL be imported in the Definitions element that contains the instance of ItemDefinition, using an Import element specifying both the namespace value and its name when used a qualifier

It really is all over the place!!!

So, maybe we just need to take a deep breath and seek clarification with the spec authors exactly what we're supposed to do with all this?

Stuff like:

- what is the purpose of the Definitions namespace?
- does import element namespace relate to a Definitions namespace?
-- are they the same thing?
-- if so, how are duplicates Definitions namespaces with a set of model files handled?
- can an external file be used _without_ an import, as per the example in 2.3.2'?
-- if so, (it has no import name) how are its elements referenced in an expression?

Thoughts?

Greg.

tarilabs:
> I think we need to get some stuff clarified as things are all over the place (IMO).

I always try to help to clarify things at least as per my understanding.
On this subject however I believe I have a different perspective, which _avoids_ most of the ambiguities you are raising, and of course I'm not implying my perspective is necessarily correct, but at least I find it is working. So I am not sure how much of further help, beside proposing a https://github.com/dmn-tck/tck/pull/291 which workaround the inconsistencies altogether, I can do.

I'll try to feedback on some of the new points you raised.


> [about example in 12.3.2] Which it says is valid ... and does not use an import. But it'll have no prefix 'name' so cannot be used in an expression like <name>.prebureauriskDec01. Hmm.

That to me clearly a snippet of a bigger DMN file.
From my perspetive such file must contain an Import.
The import will have `namespace="http://www.example.org/Definitions01.xml"` for that to be valid.
Additionally, if you want to use that in an expression in the way you have indicated, 
image,
the same import must also correspond with the attribute for name being `name="<name>"`

.

> what is the purpose of the Definitions namespace?

for instance when you import, you want to point to the other model Definitions' namespace

> does import element namespace relate to a Definitions namespace?
> -- are they the same thing?

my perspective is they correspond with the same _value_ when you import a DMN model from another DMN model with the import

> -- if so, how are duplicates Definitions namespaces with a set of model files handled?

my perspective if you "bootstrap" an engine with 2 DMN model having the same namespace, it is inconsistent --by the spec.
Please notice this is also compatible with what the spec are saying for the namespace, imho: like, they are not expecting when you produce a DMN model serialization, for the namespace to be unique in the whole Universe now and forever.

At the same time however, imho, they are not mandating that an engine can successfully bootstrap an execution if you supplied two models with the same namespace.

An example I believe is always worth to try explain what I mean.
Say you have two models
```
<definitions namespace="abc" ...>
<decision name="decision01"><literalExpression>1</literalExpression></decision>
```
```
<definitions namespace="abc" ...>
<decision name="decision01"><literalExpression>9</literalExpression></decision>
```

What is the result of evaluating decision01 for the model having namespace "abc".
It is undeterministic.
From this example you might be tempted to say to the engine: evaluate the first model, or evaluate the first file, but the DMN spec doesn't deal with this, similarly it doesn't deal how the API should be designed, beside hinting at stuff for decision services etc.

If you consider instead, when you bootstrap, you are only putting in models with unique namespace, you solve the above inconsistency.

Just for transparency, in Drools we have a slightly more sophisticated way to distinguish to model which may also happen to share the same namespace; however as I view it from the DMN spec only, safe route is to have unique in namespace when you want to import.

> can an external file be used without an import, as per the example in 2.3.2'?
> -- if so, (it has no import name) how are its elements referenced in an expression?

As I tried to explain above, I don't believe you can use an external file without an import.
In other words, I simply don't share the perspective you can reference an (imported) ItemDefinition or DRGElement without having defined, indeed, an `<import/>` for it.

StrayAlien:
Hi @tarilabs,

Thanks for the comments.  I mean no offence (always very hard to get across in these responses), but, respectfully, I believe they are assumptions, not facts.

I don't believe the PR https://github.com/dmn-tck/tck/pull/291 should be accepted as I don't believe it is spec compliant.  I think the tests in this PR are representative - albeit using the 'location' aspects of the spec.

That an import namespace is related to a model namespace is not supported in the spec anywhere - in fact, I believe the spec gives us exactly the opposite.  It also says that DMN model namespaces are _not_ unique.  I'll keep coming back to the "MAY differ" quote - not sure why we're ignoring that as it is so very important in this discussion.

Consider these two spec facts:

1. "Each file SHALL declare a “namespace” that MAY differ between multiple files of one model."  
2. "The (import element) namespace value should be globally unique"

Read them twice and add them up. Clearly the "should be unique" namespace on an import element _cannot_ be related to the "MAY differ" namespace - one should be unique and the other may be duplicated.  They just cannot be the same thing - nor, I believe, are they _intended_ to be the same thing. The just share an (overloaded) name.

I understand it may be pragmatic for an implementation to treat DMN namespaces as unique, and an import namespace as a reference to it, and it may make life easier - but it is not spec compliant.

The solution to the example you have given above is very simple if you accept DMN namespace and import namespace are not related:

> An example I believe is always worth to try explain what I mean.
> Say you have two models
> 
> ```
> <definitions namespace="abc" ...>
> <decision name="decision01"><literalExpression>1</literalExpression></decision>
> ```
> 
> ```
> <definitions namespace="abc" ...>
> <decision name="decision01"><literalExpression>9</literalExpression></decision>
> ```
> 
> What is the result of evaluating decision01 for the model having namespace "abc".
> It is undeterministic.

It is _entirely_ deterministic!  The above example is only problematic when you assume import namespace relates to model namespace. Solving this is _exactly_ what the namespace on an import element is for. Those two models can both be imported into another model each with a different ('should be global') namespace and then that namespace used in references _within_ that importing model. Solved - and I think this is how it is supposed to be.  No magic or sophisticated disambiguation required.  Very simple.

This is how we've implemented it and it works just fine.  The tests in this PR verify same.  The above example is effectively covered by these PR tests to show that it can and does work.  

I believe the PR at https://github.com/dmn-tck/tck/pull/291 is violating the spec in the following ways:

* it assumes that model namespaces are unique with a set of related models
* it assumes that an import namespace relates to a model namespace
* it assumes that an import namespace is also an external reference to something (in this case, a namespace in an external file)    

I do think some things need clarifying about the spec, though IMO, the above namespace/import/uniqueness stuff is all pretty clear.  If you feel the opposite then we should seek clarification.  I went back and read the `http://www.example.org/Definitions01.xml` it does mention the import element. Sorry!!!! So maybe it is not so 'all over the place' if you accept the namespace/import/uniqueness stuff I've outlined above.

Greg.

etirelli:
@StrayAlien a lot being discussed here, so I just want to relay some information directly from the RTF:

> what is the purpose of the Definitions namespace?

To create a unique identifier for a model.

> does import element namespace relate to a Definitions namespace?
> -- are they the same thing?

Yes, they are the same thing, and that is how a model is referenced. By it's namespace. 

-- if so, how are duplicates Definitions namespaces with a set of model files handled?

No two models should have the same namespace. I know, it is not how I would have done, but this answer comes straight from Gary that authored this on the spec.

> can an external file be used without an import, as per the example in 2.3.2'?
> -- if so, (it has no import name) how are its elements referenced in an expression?

No, in DMN all references must be explicit. If one needs to reference any part of an external model, that external model must be imported.

-------

Like you I also had similar questions because it is not clear in the spec. So I asked the RTF, and more directly Gary, as he is the author of that part, and he confirmed that the intent of the spec is that a <definitions> namespace should always be unique. The text of the spec might need change to fix ambiguities. 

Please note that in Drools, we have extra logic to treat non-unique namespaces that was implemented before Gary gave us the answer, but for all purposes, from the spec/RTF perspective, no two models can have the same namespace.

So, on your questions:

> * it assumes that model namespaces are unique with a set of related models

According to Gary, it is supposed to me.

> * it assumes that an import namespace relates to a model namespace

It is supposed to be. In other words, the definitions namespace is the "unique identifier" of a model. Importing a model requires importing its namespace.

> * it assumes that an import namespace is also an external reference to something (in this case, a namespace in an external file)

It is. 

From page 49 of the spec:

> An instance of Import has a namespace, which is a URI that identifies the namespace of the imported element, and also a name, inherited from NamedElement, which is a string that serves as a prefix in namespace-qualified names, such as typeRefs specifying imported ItemDefinitions and expressions referencing imported InformationItems. The namespace value should be globally unique, but the import name, which is typically a short business-friendly name,must be distinct from the names of other imports, decisions, input data, business knowledge models, decision services, and item definitions within the importing model only.

Hope this helps.

StrayAlien:
Hi  @etirelli 

Thanks for posting.  Appreciated.

If the author meant those things he sure didn't get as far to putting them into the spec.

If light of your comments, what does this mean then?

> "Each file SHALL declare a “namespace” that MAY differ between multiple files of one model."

It seems to specify exactly the opposite - and a lot hinges on it.

etirelli:
> "Each file SHALL declare a “namespace” that MAY differ between multiple files of one model."

Hehe,  :) when I pressed Gary about why they just simply say explicitly that each namespace is unique, I did not exactly understand what he said, but it goes along the lines that in XML, you can actually compose a document from multiple files. So in theory, a DMN model that is persisted in XML, can actually be persisted as multiple files, because XML itself supports importing models one into another (not to confuse with the DMN import that is something else).

I personally have no idea when/how/why anyone would create a single model from multiple files, and even worse, how that would work if namespaces were different.

StrayAlien:
@etirelli , @tarilabs ,

Thanks - I can't help but say I'm a bit bummed by it all - but such is life. < standing up and dusting off >.

I kind of feel like it is an opportunity lost.  The separation of concerns between an import namespace being a 'local' ref to whatever it is that is being imported I feel is a better design for a mixed vendor environment, and especially a microservice one.

Like, one has to ask oneself, what does the import namespace mean when importing an XML schema? Or what about a JSON schema? - no namespaces there.  They _sure_ aren't DMN model namespaces - so, immediately, there is (modern term) cognitive friction.  It is simply not consistent - you have to think too much about it.  Like, if I am importing a JSON schema, the namespace has to be model-local ... so why not for a DMN model? "because it is a DMN model, not a JSON schema"  Huh?

Additionally, one might ask, what to do when (say) importing a model via http (using locationURI) and the import element namespace does _not_ match the namespace in the http-fetched model.  What then?  Error?  To me it only make sense to use the import element namespace - so, in that case, it _does_ become a model-local reference anyways.

It only seems to make sense to me to treat them separately.  Then, everything becomes consistent.   But, such is life.  (said smiling) Might be an example of: https://en.wiktionary.org/wiki/a_camel_is_a_horse_designed_by_a_committee. :-)

I really (_really_!) think the 1.3 spec must be updated with these things explicitly - otherwise, it seems it carries maybe a little too little real information and maybe a little too much dis-information.  I really don't it is possible to arrive concretely at the connection between namespaces etc with what is in the spec as it is.  

Let's not start with an 'import' that only optionally says what is being imported!   

At any rate - @tarilabs, rather than PRs on PRs, as a suggestion, perhaps it might make more sense for me to cancel this PR and maybe you resubmit a version you feel happy with? - along with your amendments to your runner?  It might make life simpler in the end.

I'll keep these tests aside as they as valid tests for our runtime.

Greg.

etirelli:
@StrayAlien 

Please don't shoot the messenger. :) I do agree with you, I don't think the import feature is fully fleshed out. This feature was impossible to implement in DMN 1.1 because it was missing the "name" attribute, that was added in DMN 1.2 and I believe the (few) vendors that have an implementation for it now, only support importing DMN model (I might be wrong).

But here in the TCK, all we can do is prove something is not working, open a ticket to the RTF and handle it there. Submitting a full proposal with the ticket is actually the best way to ensure it gets fixed/changed for upcoming releases.

For DMN 1.2 however, importing a model really depends on 2 attributes: "namespace" of the imported model, and the "name" local alias to reference elements of said model. The locationURI attribute is optional, and is used by some tools, but I don't believe its semantics are well defined enough for interchange between vendors. It has its own problems, like what happens when moving a model from a dev/stage environment to production, where the locationURI would change?

Again, I am on your page in regards to this feature needing improvement in the spec. But what I describe is its current state in DMN 1.2.

StrayAlien:
Thanks @etirelli - not shooting any messengers!  :-)

Hopefully, we can get the spec updated to be more concrete.  Thanks again for the insight.  Appreciated.

Greg.

PR #275, opened 2019-03-30
non-compliant: 0019-flight-rebooking

StrayAlien:
small update here to remove 'feel:' from typeRefs.  One typeref changed from string to number.  `Passenger1.Miles` is represented elsewhere in the model as a number.

```xml
		        <input id="b_Passenger_Priority_dt_i_P1_Miles" label="Passenger1.Miles">
		            <inputExpression typeRef="number">
		                <text>Passenger1.Miles</text>
		            </inputExpression>
		        </input>
```

### DIFF
--- a/TestCases/non-compliant/0019-flight-rebooking/0019-flight-rebooking.dmn
+++ b/TestCases/non-compliant/0019-flight-rebooking/0019-flight-rebooking.dmn
@@ -8,62 +8,62 @@
 	
 	<itemDefinition id="_tFlight" name="tFlight">
 		<itemComponent id="_tFlight_Flight" name="Flight Number">
-			<typeRef>feel:string</typeRef>
+			<typeRef>string</typeRef>
 		</itemComponent>
 		<itemComponent id="_tFlight_From" name="From">
-			<typeRef>feel:string</typeRef>
+			<typeRef>string</typeRef>
 		</itemComponent>
 		<itemComponent id="_tFlight_To" name="To">
-			<typeRef>feel:string</typeRef>
+			<typeRef>string</typeRef>
 		</itemComponent>
 		<itemComponent id="_tFlight_Dep" name="Departure">
-			<typeRef>feel:dateTime</typeRef>
+			<typeRef>dateTime</typeRef>
 		</itemComponent>
 		<itemComponent id="_tFlight_Arr" name="Arrival">
-			<typeRef>feel:dateTime</typeRef>
+			<typeRef>dateTime</typeRef>
 		</itemComponent>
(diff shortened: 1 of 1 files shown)


agilepro:
Matteo asks the question: why is this in the non-compliant folder?   He also want to check to see if all the changes have been made to make this 1.2 compliant.

tarilabs:
@StrayAlien could you also double-check for the whitespace changes which looks have been commited in, please?

tarilabs:
could it be the line-ending?

StrayAlien:
re line endings - strange - my local diff shows just my changes. I'll check it out. Thanks.

Re why this is in non-compliant?  Recursion I believe.

My personal feeling is that is should be non-compliant still because BKM recursion is a grey area of the spec.  Despite the fact that I want it there and that I like it, I would say that it is not actually supported by the spec.  This is also why my recent PR on recursion tests is in the non-compliant folder.

(btw, this is unlike self-referencing ItemDefinitions which I do think is supported by the spec) 

But I think we should get recursion in the spec!.  :-)

I'll open an issue with the OMG and create issue here so we can track it.

StrayAlien:
Re line endings - sorry, I'm not sure what is going on there.  I checked out the previous version of the file and pushed it so as to revery the changes and git still thought I had made changes in some way.  I'm on windows with git autocrlf set to true - as it has been for all other commits.

The reverted file seems to have crlf on all line endings as you would expect.  I have reapplied all my changes to the revert files and no joy.

So, apologies for the 'noise' but the changes are really just removing 'feel:' and 'kie:' from types and the change mentioned above to change from string to number:

```xml
<input id="b_Passenger_Priority_dt_i_P1_Miles" label="Passenger1.Miles">
    <inputExpression typeRef="number">
        <text>Passenger1.Miles</text>
    </inputExpression>
</input>
```    

Apols again ...

StrayAlien:
anything wrong with this PR?

tarilabs:
> anything wrong with this PR?

as mentioned in previous comments, the original commits introduced "whitespace" changes for the whole file which impacted for the full file change, and personally I would avoid unless strictly necessary. 

I appended some commits which should perform as you proposals, but without this issue. When the PR will be squash+merged, as this was raised by you, the resulting commit on master will attribute to you.

StrayAlien:
Thanks @tarilabs - I coudn't get to the bottom of that!  All indications here were just those lines were changed.  Even looking in a hex editor.  I can only suspect something weirdo about CR/LF conversions somewhere.

Thanks again, and once more, apologies for the hassle.  Not intended.

PR #279, opened 2019-04-01
non-compliant: 0088-recursion

StrayAlien:
All, though the rebooking model in non-compliant exercises recursion, I thought it best to put in some specific tests.

The recursion here is in two flavours.  

One is a basic fibonacci test.  

The other is a little more interesting: it is recursive to depth-first traverse a tree structure, but also defines a typeRef that is 'self referencing' to create a tree type has children of the same type as itself.

The spec does not provide a 'well-formed' definition for ItemDefinition nor contains any references that I could find to it being acyclic so I am pretty sure a self-referencing structure is legal.

Comments welcome.

### DIFF
--- a/TestCases/non-compliant/0088-recursion/0088-recursion-test-01.xml
+++ b/TestCases/non-compliant/0088-recursion/0088-recursion-test-01.xml
@@ -1,0 +1,89 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<!-- Contributed to DMN TCK by StrayAlien -->
+<testCases xmlns="http://www.omg.org/spec/DMN/20160719/testcase" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation=""  xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+
+    <modelName>0088-recursion.dmn</modelName>
+    <labels>
+        <label>Recursion</label>
+    </labels>
+
+    <testCase id="001">
+        <description>fibonacci sequence fn(10)</description>
+        <resultNode name="fibonacci_001" type="decision">
+            <expected>
+                <value xsi:type="xsd:decimal">55</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+
+    <testCase id="002">
+        <description>input data tree traversal using self referencing datatype</description>
+        <inputNode name="treeData">
+            <component name="children">
+                <list>
+                    <item>
+                        <component name="children">
+                            <list>
+                                <item>
+                                    <component name="children">
+                                        <list></list>
+                                    </component>
+                                    <component name="value">
+                                        <value xsi:type="xsd:decimal">4</value>
+                                    </component>
+                                </item>
+                                <item>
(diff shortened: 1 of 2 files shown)


StrayAlien:
Hi Matteo, 

Thanks.  As I read it, ItemDefinitions are not part of the requirements subgraph of DRG elements.  The definition well-formed for each of the DRG elements only includes other DRG elements.  I could be wrong, but I can't see how an ItemDefiniton, or a reference to one could cause a cycle between DRG elements.  

The fact it doesn't mention well-formedness for ItemDefinitions led me to the opposite conclusion which was that it omitted to permit such structures!

opatrascoiu:
I agree with Greg on this one. There is no restriction on the spec about recursive ItemDefinitions.

However, the equivalent / conforms to relationships do not cover that, hence it needs to be specified (e.g. name equivalence vs structural equivalence). 

Once we agree on the approach we can submit a proposal for DMN 1.3.

StrayAlien:
Hi @opatrascoiu,

Thanks, just my opinion here, but I am pretty sure the current spec equivalence / conformance does cover it - and quite naturally so.   

The spec just refers to equivalence / conformance of list elements and context components and so on.  That still holds true for a tree structure.

Actually, if I were writing a spec for a language that permitted self-referencing data-types, I reckon it'd be pretty close to what the spec says right now.  It is implied.

I sure think the spec supports it more naturally than a 'self referencing BKM that creates a cycle in an explicitly spec defined acyclic graph without even a reference to itself in its requirements list'.  :-)  

.. or have I missed something ? .. happy for advice.

tarilabs:
> EDIT: to support the above, ItemDefinitions cannot reference DRG elements so there is no possibility to create a cycle whereby a DRG element refers to a type and then somehow that type references a DRG element that creates a cycle back to the same DRG element. I don't believe it is possible.

say you have an Employee ItemDefintion. And you have a "supervisor" field which is a circular reference. in a DRGElement such a Decision, you could have normal employees do reference to another Decision for supervisor. Now take a consultant, which you want to express the supervisor is itself. You would have such Decision for the consultant, an informationrequirement on itself because you want to define the supervisor is itself. Can you point any pitfal in my example @StrayAlien please?

etirelli:
I briefly discussed this with @tarilabs, but I think @StrayAlien is correct. Item definitions are allowed to self reference, based on the current spec. It is not clear to me the full impact of that choice, but that's what we are here for... :)

StrayAlien:
Thanks @etirelli @tarilabs @opatrascoiu.

Cool. I think it seems reasonable to be able to expect a tree or graph of data.   @tarilabs - responded over on your raised issue.

I think we leave these tests here in non-compliant until recursion is also supported formally in the spec.

fyi: https://github.com/dmn-tck/tck/issues/284

@opatrascoiu would you mind expanding on:

> However, the equivalent / conforms to relationships do not cover that, hence it needs to be specified (e.g. name equivalence vs structural equivalence).

Or have we established otherwise?

StrayAlien:
Thanks @tarilabs - I totally agree it should be in non-compliant until recursion is official.  Cool by me.

Issue #274, opened 2019-03-25
Using InputData with imports, and imports of imports


StrayAlien:
All, imports can be pretty confusing, so bear with me here.

This is not so much as issue - but more of a ‘let’s agree on how we do this’.  It is about how to supply input data to a model when the inputData is an import, and import of an import, or deeper …. and so on. And we don’t have anything guaranteed to be unique.

For TCK test model execution we have to supply the required input data as input nodes.

Say we have three models - A, B, C.  A imports B as `foo`, and B imports C also as `foo`.  The import name property is model-local.  We could have used different names.  Let’s say all three models have an InputData element with id `fooInput`. 

We need to supply three things called `fooInput` to the test, and they all mean something different.

Does the test input data look something like this (as JSON for clarity) ?:

```json
{
  "fooInput": "for model A",
  "foo": {
    "fooInput": "for model B",
    "foo": {
      "fooInput": "for model C"
    }
  }
}
```

Get my drift?  When A references `fooInput` it means the data in its own model.  When A references `foo.fooInput`, it means the data for model B, and when B references `fooInput` it means its own data, and when B references `foo.fooInput` it means model C and so on.

As nothing in the spec _guarantees_ uniqueness between models somehow we have to distinguish between what input data is meant for what model.

The only real thing we could have possibly used to make things unique would be import locationURI#elementId …. But locationURI is optional ….

Hopefully that makes some small sense.

Happy for comments.

Greg.

PS.  I did have a look at the RH import unit tests and they seem to assume that an import `name` is globally unique. If I read that right, I don't actually think that is correct.   But I am not familiar with that code base so I could be wrong.

agilepro:
There was a discussion on this in the meeting today.  Simon explained how this problem was solved by the implementations.  (I didn't exactly follow it.)   I might be nice to write this down because the agreement was that this had to be figured out, and it was not in the spec anywhere.

StrayAlien:
@agilepro @SimonRinguette @tarilabs 

I'll summarise.  Yes - the spec is very light here on what is a complex thing.  Wait until we start testing decision services with import inputdata - but that will be another separate issue I raise for discussion.

The discussion agreed the input data in my previous post here should look like this:

```json
{
  "fooInput": "for model A",
  "foo.fooInput": "for model B",
  "foo.foo.fooInput": "for model C"
}
```
(personally I prefer the nested data model - easier to group data and less effort to process etc.)

[...]

PR #280, opened 2019-04-01
0086a-imports-input

StrayAlien:
An import input data test to cover what is discussed over here: https://github.com/dmn-tck/tck/issues/274.

This suite is pretty compact - it is actually only one test.  It has three models A, B and C.  A imports B as 'i', and B imports C as 'i'.  A _also_ imports C directly as 'x'.

A has a decision that uses a local inputData and also the a decision/inputData from B.  That decision in B also uses a 'local' inputData and the a decision/inputData in C.  The decision in C uses a 'local' inputData.

The decision in A _also_ uses the decision/inputData and decision from C directly.

The inputNode to the test is structured to support that.

Phew.


agilepro:
Matteo to double check and merge if it all looks OK.

StrayAlien:
@tarilabs - you were saying you prefer 'context' objects rather than dot-path notation for the data - is that how Red Hat implements it .. or do you do dot-notation?

(I prefer an object  - it is easy and unambiguous in terms of naming)

Issue #285, opened 2019-04-07
error in issue DMN13-133 ?


StrayAlien:
@tarilabs,

Re https://issues.omg.org/issues/DMN13-133

Should the last line in the corrected example be:

`[1, [2.1, 2.2], 3, 4, 5]` and not `[1, [2.1, 2.1], 3, 4, 5]` as given in the link above?

(note the 2.2 vs 2.1 in the second element)

tarilabs:
Thanks for reporting Greg, but in the already-voted proposal that was already fixed.

StrayAlien:
@tarilabs - thanks mate.  Excellent. Time to put in a test to assert it.  :-)

PR #286, opened 2019-04-08
0090-feel-paths

StrayAlien:
Initial commit to cover feel paths on nested lists.  Only two tests covering the 'nested lists' example in the spec with the v1.2 corrections requested here: https://issues.omg.org/issues/DMN13-133 and with amendments noted here: https://github.com/dmn-tck/tck/issues/285.

Hopefully not too controversial.

EDIT: And thanks @tarilabs for the correction on the spec example.  Man, I stared at that thing for _ages_ trying to figure out what was supposed to happen - it totally fried my brain.  It didn't occur to me it was v1.1 singleton array type equality.  Doh!  After the correction things were pretty straight forward.  Thanks again.


StrayAlien:
@tarilabs, @opatrascoiu, @agilepro  - hopefully this one has no mysteries - it is just the updated example from Matteo's spec correction (thanks again!).  Two very small tests showing the two equality statements from the example. 

Should be okay to merge?  Any issues?

PR #289, opened 2019-04-26
Align to DMN13-125 Disambiguation for Modulo / Remainder function

tarilabs:
I am proposing the following alignments, and new tests, following DMN13-125. 

In short: align modulo function results for negative/non-integer with Spreadsheet-like system.

My perspective is that DMN13-125 clarifies behavior, even on DMNv1.2, for an unclear part of the spec. Hence I am raising this test cases now, in order to align with DMN13-125 clarifications.

### DIFF
--- a/TestCases/compliance-level-3/0056-feel-modulo-function/0056-feel-modulo-function-test-01.xml
+++ b/TestCases/compliance-level-3/0056-feel-modulo-function/0056-feel-modulo-function-test-01.xml
@@ -21,7 +21,7 @@
         <description>Assert on negative divisor</description>
         <resultNode name="decision001_a" type="decision">
             <expected>
-                <value xsi:type="xsd:decimal">2</value>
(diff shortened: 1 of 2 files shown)


etirelli:
Here is the text that is going into the spec:

-----
modulo( dividend, divisor )

dividend and divisor are numbers, where divisor must not be 0 (zero). Returns the remainder of the division of dividend by divisor. In case either dividend or divisor is negative, the result has the same sign of the divisor. The modulo function can be expressed as follows:

modulo(dividend, divisor) = dividend - divisor * floor( dividend/divisor )

modulo(12,5)= 2
modulo(-12,5)= 3
modulo(12,-5)= -3
modulo(-12,-5)= -2
modulo(10.1, 4.5)= 1.1
modulo(-10.1, 4.5)= 3.4
modulo(10.1, -4.5)= -3.4
modulo(-10.1, -4.5)= -1.1
--------

PR #291, opened 2019-04-28
TCK infrastructure API changes to support multiple models, for DMN import features

tarilabs:
Proposal for TCK infrastructure API changes to support multiple models.

This is to provide similarly to model under test URL, a collection of additional URL located in the same directory of the test case.
Please notice there is no change proposed to TCK xml file, in order to keep things simple.
I will comment on the limitations of other PRs for similar scope in the relevant PR page.

Adding at the same time test case `0086-import` to demonstrate.

### DIFF
--- a/runners/dmn-tck-runner/src/main/java/org/omg/dmn/tck/runner/junit4/DmnTckVendorTestSuite.java
+++ b/runners/dmn-tck-runner/src/main/java/org/omg/dmn/tck/runner/junit4/DmnTckVendorTestSuite.java
@@ -14,11 +14,12 @@
 
 package org.omg.dmn.tck.runner.junit4;
 
+import java.net.URL;
+import java.util.Collection;
+import java.util.List;
+
 import org.junit.runner.Description;
 import org.omg.dmn.tck.marshaller._20160719.TestCases;
-
-import java.net.URL;
-import java.util.List;
 
 /**
  * An interface to allow the TCK runner integration
@@ -64,9 +65,25 @@ public interface DmnTckVendorTestSuite {
      *                call.
      * @param testCases the parsed content of a test file.
      *
-     * @param modelURI the resolved URI to the DMN model file.
+     * @param modelURL the resolved URL to the DMN model file.
+     * @deprecated To support DMN features involving multiple models, see {@link #beforeTestCases(TestSuiteContext, TestCases, URL, List)}
      */
+    @Deprecated
     void beforeTestCases(TestSuiteContext context, TestCases testCases, URL modelURL );
+
+    /**
+     * A callback to give vendors an opportunity for initialization
+     * before running each test file.
+     *
+     * @param context the context created by the <code>createContext()</code>
+     *                call.
+     * @param testCases the parsed content of a test file.
+     *
+     * @param modelURL the resolved URL to the DMN model file.
+     */
+    default void beforeTestCases(TestSuiteContext context, TestCases testCases, URL modelURL, Collection<? extends URL> additionalModels) {
+        beforeTestCases(context, testCases, modelURL);
+    }
 
     /**
      * A callback to give vendors an opportunity to prepare for a
--- a/runners/dmn-tck-runner/src/main/java/org/omg/dmn/tck/runner/junit4/DmnTckRunner.java
+++ b/runners/dmn-tck-runner/src/main/java/org/omg/dmn/tck/runner/junit4/DmnTckRunner.java
@@ -14,6 +14,26 @@
 
 package org.omg.dmn.tck.runner.junit4;
 
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.stream.Collectors;
+
+import javax.xml.bind.JAXBException;
+
 import org.junit.runner.Description;
 import org.junit.runner.notification.Failure;
 import org.junit.runner.notification.RunNotifier;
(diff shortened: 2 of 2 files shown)


runners/dmn-tck-runner/src/main/java/org/omg/dmn/tck/runner/junit4/DmnTckVendorTestSuite.java
```diff
@@ -64,10 +65,26 @@
...
+     * A callback to give vendors an opportunity for initialization
+     * before running each test file.
+     *
+     * @param context the context created by the <code>createContext()</code>
+     *                call.
+     * @param testCases the parsed content of a test file.
+     *
+     * @param modelURL the resolved URL to the DMN model file.
+     */
+    default void beforeTestCases(TestSuiteContext context, TestCases testCases, URL modelURL, Collection<? extends URL> additionalModels) {
```
tarilabs: In order to maintain backward compatibility, a default implementation calling back to previous method is provided. Previous method this way is supported OOTB for unchanged runners. Old runners will still see old callback, however naturally without any explicit change they won't be able to make use of the new parameter.
New runners can make use of this new method.

TestCases/compliance-level-3/0086-import/0086-import.dmn
```diff
@@ -0,0 +1,108 @@
...
+                      xmlns:feel="http://www.omg.org/spec/DMN/20180521/FEEL/" 
+                      xmlns:tc="http://www.omg.org/spec/DMN/20160719/testcase"
+                      xmlns:include1="http://www.trisotech.com/definitions/_f27bb64b-6fc7-4e1f-9848-11ba35e0df36" 
+                      xmlns:trisofeed="http://trisotech.com/feed" 
+                      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
+                      xmlns:xsd="http://www.w3.org/2001/XMLSchema" 
+                      xmlns="http://www.omg.org/spec/DMN/20180521/MODEL/" id="_f27bb64b-6fc7-4e1f-9848-11ba35e0df36" 
+                      namespace="http://www.trisotech.com/definitions/_c3e08836-7973-4e4d-af2b-d46b23725c13"          exporter="DMN Modeler" exporterVersion="6.2.1" 
+                      name="Import BKM and have a Decision Ctx with DT" triso:logoChoice="Default">
+    <import namespace="http://www.trisotech.com/definitions/_f27bb64b-6fc7-4e1f-9848-11ba35e0df36"
```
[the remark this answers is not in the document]
tarilabs: Yes, it is indeed a replacement for the #272 PR test, as I mentioned here https://github.com/dmn-tck/tck/pull/272#issuecomment-487385943 and here https://github.com/dmn-tck/tck/pull/272#issuecomment-488036941

No, I have no rush to have this merged until there is a consensus.

> but still - let's discuss first

This is indeed a PR and not a direct push to master

TestCases/compliance-level-3/0086-import/0086-import.dmn
```diff
@@ -0,0 +1,108 @@
...
+                     name="myimport"
+                     importType="http://www.omg.org/spec/DMN/20180521/MODEL/"
+                     />
+    <inputData id="_9df2ca89-d100-4ba3-9a44-6a71cae5c001" name="A Person">
+        <variable name="A Person" id="_9e1f6cbc-584f-41f6-8748-97f579a3df43" typeRef="myimport.tPerson"/>
```
tarilabs: typeRef uses import's name, as per DMN spec.

TestCases/compliance-level-3/0086-import/0086-import.dmn
```diff
@@ -0,0 +1,108 @@
...
+    </inputData>
+    <decision id="_2d131943-c513-416b-acc6-6efe8fe01ba4" name="A Decision Ctx with DT" triso:useOutputTypeAsAnswer="false">
+        <variable name="A Decision Ctx with DT" id="_1a9b6949-afac-4c9e-afcd-178d9f720f29" typeRef="Any"/>
+        <informationRequirement id="_01a9f8c0-6333-45cf-a693-e2e67b23fa13">
+            <requiredInput href="#_9df2ca89-d100-4ba3-9a44-6a71cae5c001"/>
+        </informationRequirement>
+        <knowledgeRequirement id="_73d5099c-5a54-4a04-a9b5-80e1957ad8e9">
+            <requiredKnowledge href="http://www.trisotech.com/definitions/_f27bb64b-6fc7-4e1f-9848-11ba35e0df36#_32543811-b499-4608-b784-6c6f294b1c58"/>
```
tarilabs: href uses import's namespace, as per DMN spec.

TestCases/compliance-level-3/0086-import/0086-import.dmn
tarilabs: `typeRef="myimport.tPerson"`

TestCases/compliance-level-3/0086-import/0086-import.dmn
tarilabs: `href="http://www.trisotech.com/definitions/_f27bb64b-6fc7-4e1f-9848-11ba35e0df36#_32543811-b499-4608-b784-6c6f294b1c58"`

etirelli:
@tarilabs can you please split the tests from here into a separate PR, so that we can merge the infrastructure changes without braking compatibility for other vendors. 

Once we reach an agreement on the tests, we can merge the tests.

StrayAlien:
@etirelli @tarilabs - perhaps it might be best if I cancel the other PRs that involve imports - rather than PRs on PRs etc - it may get confusing.  If the tests in those PRs  (0086-imports, 0086a-imports-input, 0087-chapter-11-example) could prove useful as a starting point then maybe get a copy of them and modify?

Thoughts?

Greg.

tarilabs:
> rather than PRs on PRs

This was now mentioned twice but I am afraid I am not understanding it; this PR targets master, not another PR(branch). What am I missing?

StrayAlien:
Oh roger that - yes, correct. I saw the modifications 0086-imports thing (being in PR) and had a brain fade.  

I'll remove the other PRs that have imports.  Those named above.

tarilabs:
> I'll remove the other PRs that have imports

imho you can move them into non-compliant because they have duplicate namespaces.
If you are now sharing instead the view namespace is to be unique, you can amend and reflect in the same original PR.

Anyhow I am being requested to split this PR as well into infrastrucutrue API changes and tests, anyway

StrayAlien:
Heya @tarilabs, "no worries" as we say here.  :-) 

I think it easier to remove them for now and pending what you commit with the above I'll modify them later and maybe resubmit.

Thanks as always,

Greg.

tarilabs:
> @tarilabs can you please split the tests from here into a separate PR, so that we can merge the infrastructure changes without braking compatibility for other vendors.

Split done.
I have no idea to which _braking compatibility_ is being referred: I did provide a TCK API change with default method purposefully to avoid introducing bc problems?

agilepro:
OK to merge starting tomorrow

PR #295, opened 2019-05-09
Align 0083-feel-unicode to assumption made in PR #239

tarilabs:
that is, assumption defined in this comment:
https://github.com/dmn-tck/tck/pull/239#issuecomment-489116665

### DIFF
--- a/TestCases/compliance-level-3/0083-feel-unicode/0083-feel-unicode.dmn
+++ b/TestCases/compliance-level-3/0083-feel-unicode/0083-feel-unicode.dmn
@@ -107,24 +107,63 @@
     <decision name="substring_001" id="_substring_001">
         <variable name="substring_001"/>
         <literalExpression>
+        <!-- *assuming* string length("\uD83D\uDC0E") is 2 as per https://github.com/dmn-tck/tck/pull/239#issuecomment-489116665
+             in other words string length over the FEEL string for "🐎" results in 2,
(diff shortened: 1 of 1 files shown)


tarilabs:
Based on the meeting 2019-05-10 I will actually comment out these tests.

tarilabs:
I don't have any more reasons to propose this consistency alignment, based on the fact during TCK meeting 2019-05-11 decision was actually to comment out these tests.

@StrayAlien you may still want to take a look to what was my original proposal for tests alignments, based on https://github.com/dmn-tck/tck/pull/239#issuecomment-489116665 I am not sure if you are forking these test for your internal CI, you may want to consider my code comments. Hope possibly valuable.

I believe there will be a discussion again about these specific test cases, once https://issues.omg.org/issues/DMN13-127 is clarified.

Superseeded by https://github.com/dmn-tck/tck/pull/298 closing this PR.

StrayAlien:
Thanks @tarilabs  - I saw your excellent ascii-art inline comments.  Thanks - I wasn't at the meeting (and I don't think the minutes have been released as yet) so don't really have much context except that maybe the string length thing is 'controversial' still! 

`((string length = cans of worms) = true)`

tarilabs:
Correct, with #298 these 3 tests have been commented out as another vendor perspective did not agree with this PR of alignment, and was a valid/supported perspective to flag them as controversial.

PR #301, opened 2019-05-12
Augment 0090-feel-paths with spec example from DMN13-133

tarilabs:
The actual balloted/approved proposal for DMN13-133 involves a different example which will end up in the new release of the spec. Augmenting the original test case with the actual example.

please note @StrayAlien original test do remain valid nevertheless, thanks again.


agilepro:
Looks like OK to merge this.

PR #308, opened 2019-05-19
Proposal of TCK XSD Schema change to support Decision Service tests.

tarilabs:


tarilabs:
Comment in another discussion which I believe sensible and I used it for reference:

> [...] if you have a type=bkm, the inputNodes are the input parameters of the BKM (if any) and there can be only one result node which is of the type returned by the bkm.
> 
> If we extrapolate this to decision services, the type of the test case would be decisionService and the inputs nodes would match the inputdecision/inputdata of the service and the resultnode the outputdecision of the service to checke against.
> 
> Yes I guess that it would need a decision service name attribute. We also need to change the XSD enumeration to include decision service.

SimonRinguette:
Since we have not yet started to create BKM test cases, I would propose that if we go in the direction of adding a decisionServiceName, we instead make it invocableName to cover BKM invocation. (invocable is the superclass of both decision service and bkm in DMN 1.2).

This way, the mechanism for invocation of either a BKM or a DS would be the same:

type="decisionService/bkm"
invocableName="ds/bkm name
inputNodes: ds/bkm input requirements
expected: ds/bkm expected result

tarilabs:
Excellent point, I forgot about it; thanks @SimonRinguette

agilepro:
Discussed.  Octavian thinks it probably looks fine.

opatrascoiu:
Looked at it, like @SimonRinguette idea.

PR #328, opened 2019-07-31
Test case demonstrating InputData with imports, and imports of imports

### LINKED ISSUE #274: Using InputData with imports, and imports of imports

tarilabs:
Test case proposal to resolve #274

### DIFF
--- a/TestCases/compliance-level-3/0089-nested-inputdata-imports/0089-nested-inputdata-imports-test-01.xml
+++ b/TestCases/compliance-level-3/0089-nested-inputdata-imports/0089-nested-inputdata-imports-test-01.xml
@@ -1,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<testCases
+  xmlns="http://www.omg.org/spec/DMN/20160719/testcase"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+
+  <modelName>0089-nested-inputdata-imports.dmn</modelName>
+  <labels>
+    <label>Compliance Level 3</label>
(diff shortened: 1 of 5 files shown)


tarilabs:
As discussed in the DMN TCK meeting 2019-08-02 there is consensus this PR answer the original issue; I am leaving this PR open for further comments, if any.
I will proceed to merge it on Friday Aug 9th.

PR #355, opened 2020-01-09
usage of 'in' in identifier names - for/some/every.

StrayAlien:
The 'in' identifier is a kind of special case around the for/some/every loops.  It serves as a terminator when identifying the loop iteration variable name.  This PR's tests are to assert that, despite marking the end of the iteration var name, the 'in' identifier serves no special purpose in the loop and can be embedded in the list name or even start the name.

This is not so much proving something from the spec - more just providing tests that should fail if an impl treats 'in' too simplistically for loops, or treats 'in' as a general name terminator and not just for the loop iterator name case.

Comments welcome.    

Greg.


TestCases/compliance-level-3/0016-some-every/0016-some-every.dmn
```diff
@@ -102,6 +102,61 @@
...
+        <variable name="decision_014a"/>
+        <informationRequirement>
+            <requiredDecision href="#_valuesInAList"/>
+        </informationRequirement>
+        <literalExpression>
+            <text>some i in values in a list satisfies even(i)</text>
+        </literalExpression>
+    </decision>
+
+    <decision name="in list values" id="_inListValues">
```
tarilabs: from the DMN spec:

> A name start (grammar rule 26) SHALL NOT be a literal terminal symbol.

and again according to DMN spec FEEL grammar `in` is a literal terminal symbol.

This makes this model invalid.

TestCases/compliance-level-3/0084-feel-for-loops/0084-feel-for-loops.dmn
```diff
@@ -72,4 +72,38 @@
...
+        <variable name="decision_014"/>
+        <informationRequirement>
+            <requiredDecision href="#_daysInWeekend"/>
+        </informationRequirement>
+        <literalExpression>
+            <text>for i in days in weekend return if i = "sunday" then true else false</text>
+        </literalExpression>
+    </decision>
+
+    <decision name="in a list" id="_inAList">
```
tarilabs: from the DMN spec:

> A name start (grammar rule 26) SHALL NOT be a literal terminal symbol.

and again according to DMN spec FEEL grammar `in` is a literal terminal symbol.

This makes this model invalid.

StrayAlien:
Thanks @tarilabs - good catch.  I have removed those tests. 

Which makes we wonder (2 things actually) ... :-)

... do you consider 'not' to be a terminal?  Like, it looks valid to have a var named 'not null'.

Secondly, as the above clarification relates only to FEEL, named elements may violate this and the model will still be valid as long as FEEL is not involved (so the decision named "in a list" in the model is valid - but its usage in a FEEL expression is not - if you get what I mean) .   For example:

```
<decision name="return to bed options">
    <variable name="return to bed options"/>
    <context>
        <contextEntry>
            <variable name="in the mood for work"/>
            <literalExpression>
                <text>true</text>
            </literalExpression>
        </contextEntry>
        <contextEntry>
            <variable name="if hungry"/>
            <literalExpression>
                <text>false</text>
            </literalExpression>
        </contextEntry>
        <contextEntry>
            <variable name="not yet awake"/>
            <literalExpression>
                <text>true</text>
            </literalExpression>
        </contextEntry>
    </context>
</decision>
```

return/if/in/not would be invalid in FEEL, but, IMO, this is still a valid model.  Not every DMN impl will use only FEEL expressions.

Thoughts?

Thanks again,
Greg.

agilepro:
OK for merging after beginning of July

StrayAlien:
This guy is approved, can we get it merged please?

opatrascoiu:
The dmn files (0016 and 0020) have to change to use the DMN 1.3 xmlns at the top.

StrayAlien:
done.

StrayAlien:
happy 1st birthday to PR 355.  :-)

PR #356, opened 2020-01-10
model-local qualified element references

StrayAlien:
When an element reference href contains a namespace like this `"http://mynamespace#decision001"` is it tempting to immediately consider it as an import even though the given namespace may be the current model namespace. 

This PR has a single test that asserts that if a qualified href whose namespace refers to the current model then the reference is considered model-local and is effectively the same as having no namespace - like `#decision001"` in the above example.   

Comments welcome.

Greg.


StrayAlien:
The rebase to master seems to have created a lot of noise here.  Closing and will create a new PR.

PR #357, opened 2020-01-19
0092-feel-lambda

StrayAlien:
A number of tests to verify usage of user-defined functions and Function Definitions as lambdas.  Additionally, verification on usage and behaviour of closed over values ('closures') is included.  

A few things to note:

* no 'function' typeRefs or item definitions have been used here - so this is within what is describable in v1.2.  
* as a Decision may house any type of expression, in theory, it can also provide a lambda as a return value by using a user-defined function expression like `function (a,b) a+b` or a Function Definition expression.  I'm not saying this is how people _will_ use it, but tests are not about that - they're here to prove a mechanism works.  To that end, a number of the tests here have a Decision returning a lambda.
* I have also included some tests that show BKM, DS, and built-in functions can be passed about as arguments to other functions just like lambdas.   This is not 'lambda anonymous functions' per se, but, simply the extension that if we can pass lambdas about then it only makes sense that other functions are treated in exactly the same manner.  Like, for example, why would we _not_ be able to pass a BKM knowledge requirement to the `sort()` function when we are permitted to pass a user-defined function?

I have made each test as short and sweet as I can - passing functions about can get confusing.

As always, happy for comments.

Greg.


StrayAlien:
the rebase to master seems to have created a lot of noise.  Closing and will a new PR.

PR #365, opened 2020-06-01
Clear all results from all vendors, in preparation of website DMNv1.3

tarilabs:


tarilabs:
As discussed in TCK meeting 2020-05-22, clearing all results from all vendors, in preparation of website update/refresh of DMNv1.3

PR #366, opened 2020-06-08
0091-local-hrefs

StrayAlien:
re PR of here:  https://github.com/dmn-tck/tck/pull/356

When an element reference href contains a namespace like this "http://mynamespace#decision001" is it tempting to immediately consider it as an import even though the given namespace may be the current model namespace.

This PR has a single test that asserts that if a qualified href whose namespace refers to the current model then the reference is considered model-local and is effectively the same as having no namespace - like #decision001" in the above example.

Comments welcome.

Greg.

### DIFF
--- a/TestCases/compliance-level-3/0091-local-hrefs/0091-local-hrefs-test-01.xml
+++ b/TestCases/compliance-level-3/0091-local-hrefs/0091-local-hrefs-test-01.xml
@@ -13,7 +13,7 @@
     <testCase id="001">
         <description>qualified hrefs using model-local namespace</description>
         <inputNode name="input_001">
-            <value>input_001</value>
+            <value xsi:type="xsd:string">input_001</value>
         </inputNode>
         <resultNode name="decision_002" type="decision">
             <expected>
--- a/TestCases/compliance-level-3/0091-local-hrefs/0091-local-hrefs-test-01.xml
+++ b/TestCases/compliance-level-3/0091-local-hrefs/0091-local-hrefs-test-01.xml
@@ -1,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<!-- Contributed to DMN TCK by StrayAlien -->
+<testCases xmlns="http://www.omg.org/spec/DMN/20160719/testcase" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+    <modelName>0091-local-hrefs.dmn</modelName>
+
+    <labels>
+        <label>Compliance Level 3</label>
+        <label>FEEL Paths</label>
+        <label>FEEL Qualified Hrefs</label>
+    </labels>
+
+    <testCase id="001">
+        <description>qualified hrefs using model-local namespace</description>
+        <inputNode name="input_001">
+            <value>input_001</value>
+        </inputNode>
+        <resultNode name="decision_002" type="decision">
+            <expected>
+                <value xsi:type="xsd:string">decision_001 input_001 bkm_001</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+</testCases>
--- a/TestCases/compliance-level-3/0091-local-hrefs/0091-local-hrefs.dmn
+++ b/TestCases/compliance-level-3/0091-local-hrefs/0091-local-hrefs.dmn
@@ -1,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<definitions namespace="http://www.montera.com.au/spec/DMN/0091-local-hrefs"
+             name="0091-local-hrefs"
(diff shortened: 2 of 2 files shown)


opatrascoiu:
Looks good to me.

tarilabs:
I'm not an expert in XML spec, but based on the DMN description in "12.3.2 References within the DMN XSD" this seems to me a contradiction.

For this case, where the `path` component of the href is present, the spec elicit

> (e.g. the value of the locationURI attribute in the corresponding Import element)

and the case where the referenced element must be found, must be defined within the same xml document, is the case of no `path` and only `fragment` (the `#abc` part)

> a value of the href attribute that contains only a fragment and no path component references a DMN element that is defined in the same instance of XML file as the referencing element.

Analogously, because speaking now of reference by name, and not ID, is what concerns ItemDefinition, that is:

> If the referenced type is not imported, the prefix SHALL be omitted.

Conclusion. In my understanding, if we deem having and `namespaceofthisdmn#anidofthisdoc` should work and disalign from the intention of referencing names, it's more than okay to merge this case, but I believe needs a JIRA to the DMN RTF to explicit this behavior in the spec.

StrayAlien:
As always, very incisive @tarilabs -thank you. Very interesting about the definitions / typeDefs thing. 

I raised this PR because I operate a lot on the "developer principle of least surprise".   That is, "what would you _expect_ to happen"?  For example, in an HTML page an absolute URL to a resource in the same document is fine - if it weren't you'd probably be surprised.  Surprise = cognitive load, developer dislike and mistrust etc etc.  In this case, the 'local href' is absolute, but just a self reference.  Like what you would 'expect' of (say) an html doc.  If a DMN impl decided to store all references as absolute, then it would be surprising if that was illegal. 

But, I'm not pushing an agenda here in any way at all, so am completely okay if this does not get merged and clarification is sought with DMN heads of state.  No stress at all. But this was to highlight / verify that such a 'unsurprising'  thing might happen - a vendor may choose to store all references as absolute.  

I am very happy for the comments and insight. 

And as always, as long as we all do the same thing then we're sweet!  :-)

Greg.

tarilabs:
@StrayAlien if I take your analogy of HTML, you would usually write href relatives when you are referring within a document:

```
<a href="#detailsection">go to detail section</a>
```

and NOT:

```
<a href="https://mydomain.com/mypage#detailsection">go to detail section</a>
```

simply because of the reason that the latter would only work when in production and not while testing locally with say `127.0.0.1:8080`, while the former always works.

So I would conclude if I use your analogy of HTML that the "developer principle of least surprise" would argue _against_ this PR.

I repeat my perspective, the DMN specification does not have an explicit stance on importing ID, while it seems to have an opposing stance if importing NAME. If we want to merge this PR, based on the gap in DMNv1.3, we ought to raise a DMN RTF jira asking for the specification to explicit the _opposite_ behavior for ID reference, if compared to name reference ("_If the referenced type is not imported, the prefix SHALL be omitted_").

opatrascoiu:
I don't think DMN 1.2 / 1.3 spec prohibits such constructions for ID references. I think the restrictions in 12.3.2 are for relative URI references. 

I believe that "the reference type is not imported" means that the reference type is not defined in the current model.

agilepro:
Everyone seems to think this should go to the RTF for clarification before moving forward

opatrascoiu:
I raised the issue here https://issues.omg.org/browse/INBOX-1085.

It's in inbox, should become public after being processed by the OMG team.

tarilabs:
Ref https://github.com/dmn-tck/tck/pull/366#issuecomment-948539672
yes this accounted by the TCK validator module, but this PR was raised before we established GitHub Actions on this repo. I've raised a PR for amend accordingly.

opatrascoiu:
The test file looks good to me now after @tarilabs changes.

I think the model file has an issue: missing typeRef for input data. I believe we agreed to have typeRef unless we want to test type conformance, conversion etc.

<inputData name="input_001" id="_input_001">
        <variable name="input_001"/>
</inputData>

tarilabs:
about https://github.com/dmn-tck/tck/pull/366#issuecomment-963952949 I am planning to see if that is detected once the related PR is merged. I would expect at that moment for the GitHub action to trigger.

baldimir:
I see the issue in OMG is still Open. They are asking if there is a proposed solution. I see @opatrascoiu tagged there. @opatrascoiu do you want to propose the solution, so the issue moves in OMG or do you want me to add it there please?

baldimir:
This didn't get into the DMN 1.5, so leaving this open for future discussions in the RTF group.

baldimir:
We agreed that this makes sense. We need to create a proposal to clear this in DMN RTF.

opatrascoiu:
We discussed the issue in RTF, there is agreement to allow this use case (specify full namespace for local reference). There is a draft proposal.

opatrascoiu:
@baldimir As we already merged, could you help create the proposal officially based on this https://issues.omg.org/browse/DMN16-64? I don't have access rights.

PR #367, opened 2020-06-08
0092-feel-lambda

StrayAlien:
a re PR of here: https://github.com/dmn-tck/tck/pull/357

A number of tests to verify usage of user-defined functions and Function Definitions as lambdas. Additionally, verification on usage and behaviour of closed over values ('closures') is included.

A few things to note:

- no 'function' typeRefs or item definitions have been used here - so this is within what is describable in v1.2.
- as a Decision may house any type of expression, in theory, it can also provide a lambda as a return value by using a user-defined function expression like function (a,b) a+b or a Function Definition expression. I'm not saying this is how people will use it, but tests are not about that - they're here to prove a mechanism works. To that end, a number of the tests here have a Decision returning a lambda.
- I have also included some tests that show BKM, DS, and built-in functions can be passed about as arguments to other functions just like lambdas. This is not 'lambda anonymous functions' per se, but, simply the extension that if we can pass lambdas about then it only makes sense that other functions are treated in exactly the same manner. Like, for example, why would we not be able to pass a BKM knowledge requirement to the sort() function when we are permitted to pass a user-defined function?

I have made each test as short and sweet as I can - passing functions about can get confusing.

As always, happy for comments.

Greg.

### DIFF
--- a/TestCases/compliance-level-3/0092-feel-lambda/0092-feel-lambda-test-01.xml
+++ b/TestCases/compliance-level-3/0092-feel-lambda/0092-feel-lambda-test-01.xml
@@ -1,0 +1,204 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<!-- Contributed to DMN TCK by StrayAlien -->
+<testCases xmlns="http://www.omg.org/spec/DMN/20160719/testcase" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+    <modelName>0092-feel-lambda.dmn</modelName>
+    <labels>
+        <label>Compliance Level 3</label>
+        <label>FEEL Functions: lambda</label>
+    </labels>
+
+    <testCase id="001">
+        <description>A decision may return a user-defined-function (UDF) lambda</description>
+        <resultNode name="decision_001_1" type="decision">
+            <expected>
+                <value xsi:type="xsd:double">3</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="002">
+        <description>A decision may return a Function Definition lambda</description>
+        <resultNode name="decision_002_1" type="decision">
+            <expected>
+                <value xsi:type="xsd:double">4</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="003">
+        <description>A BKM may return a UDF lambda</description>
+        <resultNode name="decision_003_1" type="decision">
+            <expected>
+                <value xsi:type="xsd:double">5</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="004">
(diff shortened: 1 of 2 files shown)


tarilabs:
I should also note, there is imho a gap in the DMNv1.3 specification, since it's not possible to enforce a full type signature for test case `015`, but I also believe is an edge case.

opatrascoiu:
I think the test file contains same id twice : <testCase id="005">

tarilabs:
So that is to say thank you twice , this was an interesting test case also to analyze manually --as we should always do :)

tarilabs:
yes https://github.com/dmn-tck/tck/pull/367#issuecomment-646689128
well spotted !

TestCases/compliance-level-3/0092-feel-lambda/0092-feel-lambda-test-01.xml
```diff
@@ -0,0 +1,204 @@
...
+    <testCase id="005">
+        <description>A BKM may return a UDF lambda with a closed over formal param</description>
+        <resultNode name="decision_005_1" type="decision">
+            <expected>
+                <value xsi:type="xsd:double">20</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="005">
```
tarilabs, line 55:
```suggestion
    <testCase id="006">
```

opatrascoiu: 006 is missing - looks fine to me.

StrayAlien:
This guy is approved, can we get it merged please?

TestCases/compliance-level-3/0092-feel-lambda/0092-feel-lambda-test-01.xml
tarilabs: As discussed in the 2021-01-15 meeting, @StrayAlien please review this code change suggested above, originally spotted by Octavian, I think it still needs to be applied.

StrayAlien: Thanks @tarilabs @opatrascoiu - done.

StrayAlien: as a thought, it might be useful to change the xsd type of the testcase id from string to ID.  But ...... that would mean quite a lot of changes to testcase ids .... I just tried it and lot of errors .... :-( maybe when we get the current PRs cleared or reduced ....

StrayAlien:
@tarilabs @opatrascoiu - that change is in now.  Merge?

tarilabs:
Thank you for acting on the feedback @StrayAlien 
As you are aware, the TCK operates on the monthly basis, so this will be taken on-board at the next round which we planned for Feb (or the next available recurring, but I have hopes we will stick to the monthly schedule now :) )

For now I marked with a Label, but I am looking into other GH facilities to make sure not to forget.

PR #369, opened 2020-06-09
New tests case for DMN 1.3 interval functions

SimonRinguette:
New test case that validate the DMN 1.3 newly introduced interval functions

### DIFF
--- a/TestCases/compliance-level-3/1130-feel-interval/1130-feel-interval-test-01.xml
+++ b/TestCases/compliance-level-3/1130-feel-interval/1130-feel-interval-test-01.xml
@@ -1,0 +1,487 @@
+<tc:testCases xmlns:tc="http://www.omg.org/spec/DMN/20160719/testcase" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"  xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+    <tc:modelName>1130-feel-interval.dmn</tc:modelName>
+    <tc:labels>
+        <tc:label>Compliance Level 3</tc:label>
+        <tc:label>FEEL: Interval functions</tc:label>
+    </tc:labels>
+    <tc:testCase id="during">
+        <tc:resultNode type="decision" name="during">
+            <tc:expected>
+                <tc:component name="duringPRT">
+                    <tc:value>true</tc:value>
+                </tc:component>
+                <tc:component name="duringPRF">
+                    <tc:value>false</tc:value>
+                </tc:component>
+                <tc:component name="duringPRT2">
+                    <tc:value>true</tc:value>
+                </tc:component>
+                <tc:component name="duringPRT3">
+                    <tc:value>true</tc:value>
+                </tc:component>
+                <tc:component name="duringPRF2">
+                    <tc:value>false</tc:value>
+                </tc:component>
+                <tc:component name="duringPRF3">
+                    <tc:value>false</tc:value>
+                </tc:component>
+                <tc:component name="duringRRT">
+                    <tc:value>true</tc:value>
+                </tc:component>
+                <tc:component name="duringRRT2">
+                    <tc:value>true</tc:value>
+                </tc:component>
+                <tc:component name="duringRRT3">
+                    <tc:value>true</tc:value>
+                </tc:component>
+                <tc:component name="duringRRT4">
+                    <tc:value>true</tc:value>
+                </tc:component>
+                <tc:component name="duringRRT5">
+                    <tc:value>true</tc:value>
+                </tc:component>
+                <tc:component name="duringRRT6">
+                    <tc:value>true</tc:value>
+                </tc:component>
+                <tc:component name="duringRRT7">
+                    <tc:value>true</tc:value>
+                </tc:component>
+                <tc:component name="duringRRT8">
+                    <tc:value>true</tc:value>
+                </tc:component>
+            </tc:expected>
+        </tc:resultNode>
+    </tc:testCase>
+    <tc:testCase id="after">
+        <tc:resultNode type="decision" name="after">
+            <tc:expected>
+                <tc:component name="afterPPT">
+                    <tc:value>true</tc:value>
+                </tc:component>
+                <tc:component name="afterPPF">
+                    <tc:value>false</tc:value>
+                </tc:component>
+                <tc:component name="afterPRT">
+                    <tc:value>true</tc:value>
+                </tc:component>
+                <tc:component name="afterPRT2">
+                    <tc:value>true</tc:value>
+                </tc:component>
+                <tc:component name="afterPRF">
+                    <tc:value>false</tc:value>
+                </tc:component>
+                <tc:component name="afterRPF">
+                    <tc:value>false</tc:value>
+                </tc:component>
+                <tc:component name="afterRPT">
+                    <tc:value>true</tc:value>
+                </tc:component>
+                <tc:component name="afterRPT2">
+                    <tc:value>true</tc:value>
+                </tc:component>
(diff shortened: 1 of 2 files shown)


StrayAlien:
Nice one @SimonRinguette - I ewas hoping someone might chip in with these!  For completeness, many of the other function test suites also include assertions for each function for invalid number of arguments, invalid types, incorrectly named arguments and so on.  Also, for clarity, each function is it own test folder (well, the later ones are anyways) - I know this is a lot of new functions but worth splitting them up and asserting those other cases?  Happy to help.

tarilabs:
To me this looks fine, but there is something failing in comparing the structure types, which I'm not sure if it's a problem specifically of the Drools runner (could be) or it is a more general problem of the TCK test runner infra.

I believe this is meaningful to be fully investigated so I will work on it, with the priority to be sure it's not a problem of the TCK test runner infra

tarilabs:
TCK test case file is missing the `xsd:boolean` marker otherwise by default the TCK test marshaller would provide the element content as a `string`, when in this case it is a `boolean` value.

I have raised a modification suggestion here: https://github.com/SimonRinguette/tck/pull/1
to OP.
Once merged and squashed there, it will automatically reflect here.

Finally, it would be best to avoid the `1xxx` numbering since that was "reserved" https://github.com/dmn-tck/tck/wiki/How-to-Create-a-Simple-Test-Case#naming 
I don't personally concur with the reserve strategy since the batch work has completed years ago, but anyway best to stick to that for consistency reasons until we convene we can drop this reservation slot.

StrayAlien:
the word 'scalar' only appears once in the spec - and is used only with these range functions.  What is a DMN 'scalar'? - does it include date / time / datetime/ durations ?  Or just integers?

The `abs()` function now supports durations so those appear to be now considered scalar values.

Do these functions operate on those other types?

The spec says these functions return true or false, are there circumstances where they return null? 

The spec does say that two params are required - so I guess an invocation with not exactly true params is null?.  But other circumstances?
 
 Like what is the value of (say):

- `before("breakfast", false)`
- `before()` 
- `before( [1..10] )`  
- `before( 1,1,1)`  
- `before( range2: [1..10], point2: 5 )` (two params but no named params match)
- `before( null, null )` 
- `before( @”PT02H”, @”PT01H” )` 

Returning `null` (+ TCK error) would make these consistent with other functions - or do these guys never have an error as long as they get two params?

opatrascoiu:
According to 10.3.2.7 Ranges the endpoints can be of the following types: number, string, date, time, date and time, or
duration.

@SimonRinguette could you add some tests for each above type please? I believe some border edge cases suggested by @StrayAlien  above are also needed to have a proper coverage.

Also, I've noticed that the formulas from Table 78 for `overlaps` and `overlaps before` do not produce the expected results in the Example column. We need to raise an RTF issue.

tarilabs:
> I've noticed that the formulas from Table 78 for overlaps and overlaps before do not produce the expected results in the Example column. We need to raise an RTF issue.

I would be happy to know the specific cases which allegedly do not match, since the examples in table 78 were tested before being proposed. Or maybe it was a manual error when converting code to document. Thx

opatrascoiu:
@tarilabs The last 4 examples for `overlaps before` are failing for

```
(range1.start < range2.start or (range1.start = range2.start and range1.start included and range2.start included))
and (range1.end > range2.start or (range1.end = range2.start and range1.end included and range2.start included))
and (range1.end < range2.end  or (range1.end = range2.end and (not(range1.end included) or range2.end included )))
```

They pass after changing the formula to

```
(range1.start < range2.start or (range1.start = range2.start and (range1.start included and not(range2.start included))))
and (range1.end > range2.start or (range1.end = range2.start and range1.end included and range2.start included))
and (range1.end < range2.end  or (range1.end = range2.end and (not(range1.end included) or range2.end included )))
```

With `overlaps` is pretty much the same. Some tests are failing for

```
(range1.end > range2.start or (range1.end = range2.start and (range1.end included or range2.end included)))
and (range1.start < range2.end or (range1.start = range2.end and range1.start included and range2.end included))
```

but pass when defining `overlaps` as 

```
            overlaps before(range1, range2),
            || overlaps after(range1, range2),
            || includes(range1, range2),
            || includes(range2, range1)

```

StrayAlien:
I am about to implement these functions, and will be writing tests, so if @SimonRinguette is happy I can  embellish upon what is here.  Though, if so, I'll be doing each function in its own folder.  If not, no sweat. Easy.

tarilabs:
Hi @StrayAlien happy new 2021 and hope you too! It's national holiday here today, but I'd be happy if you can provide 1 example (1 example enoguh) of the mismatch you mention about, like with a specific link or a precise pointer/reference, so I will check it when I get the PC. Thx!

StrayAlien:
HI @tarilabs - good to hear from you and it sounds like you also have survived the zombie apocalypse - I hope your friends and family are also well.

An example is the `before` function: https://github.com/kiegroup/drools/blob/master/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/runtime/functions/interval/BeforeFunction.java#L35

It seems maybe RH uses 'valuex' where the spec uses 'pointx'.  I think many of the other functions in that family that use 'point' may also be represented as 'value' in RH.  Hope that helps.  Over to you.

opatrascoiu:
@SimonRinguette There is a duplicated key in the DMN and Test file

```
            <semantic:contextEntry id="_19b5e2ed-d1ff-44cc-b548-e4b70bedd320">
                <semantic:variable name="beforePRT" id="_e743e9f5-6369-40ed-bb1f-ab639225a7b4" typeRef="Any"/>
                <semantic:literalExpression expressionLanguage="https://www.omg.org/spec/DMN/20191111/FEEL/" triso:unparsed="true" id="_7a80e2c2-05c1-4f11-a2b6-387668ccc95b">
                    <semantic:text>before( 1, (1..10] )</semantic:text>
                </semantic:literalExpression>
            </semantic:contextEntry>
            <semantic:contextEntry id="_21cb9587-2d01-49fc-b44c-64d081962c7c">
                <semantic:variable name="beforePRT" id="_8090dbfe-74f6-4931-b017-1903cfd6095e" typeRef="Any"/>
                <semantic:literalExpression expressionLanguage="https://www.omg.org/spec/DMN/20191111/FEEL/" triso:unparsed="true" id="_4c6f479a-2604-4941-b870-299c099b1421">
                    <semantic:text>before( 1, [5..10] )</semantic:text>
                </semantic:literalExpression>
            </semantic:contextEntry>


```

agilepro:
I want to get back into touch with Simon, and see if he wants to update according to the comments.  The alternative is that Greg is developing a set of tests that cover this functionality.  Perhaps we should drop this, and switch to the newer effort.   Let's see what Simon wants to do.

tarilabs:
Many thanks @opatrascoiu , good catch on https://github.com/dmn-tck/tck/pull/369#issuecomment-756722834
this addresses it -> https://github.com/SimonRinguette/tck/pull/2

tarilabs:
@opatrascoiu I concur with your https://github.com/dmn-tck/tck/pull/369#issuecomment-754544858 that the formulas has the defects you described. The figures, description and examples look consistent.

I am happy to raise a DMN RTF jira for the correction, if you haven't already. Please let me know so I can know how to act accordingly 👍

opatrascoiu:
@tarilabs I already raised one https://issues.omg.org/browse/DMN14-125. Feel free to write a proposal to fix it.

tarilabs:
Done https://issues.omg.org/browse/DMN14-125?focusedCommentId=22087&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-22087

PR #373, opened 2020-06-21
Update marshaller for compatibility with JDK 9 upwards

tarilabs:


agilepro:
Would be best if we can get some of the other vendors to comment on whether they are OK with this.

tarilabs:
> Would be best if we can get some of the other vendors to comment on whether they are OK with this.

Agreed, as discussed even if a trivial change best to seek at least another vendor approval.
@StrayAlien @opatrascoiu @SimonRinguette any comments?

Issue #376, opened 2020-07-09
Types of BKMs


opatrascoiu:
A similar issue was raised by @DanielThanner  a while back https://github.com/dmn-tck/tck/issues/112: Several BKMs have incorrect typeRef in the variable attribute

According to DMN 1.2 & DMN 1.3, the semantics of the variable attribute of a BKM is

> This attribute defines a variable that is bound to the
> function defined by the FunctionDefinition, allowing
> decision logic to invoke the function by name.

Hence, the typeRef should point to an ItemDefinition that describes a function type (introduced in DMN 1.3), Any or not specified at all (in DMN 1.2). 

In some of the test cases the type of the variable is the type returned by the BKM:.
- Imported_Model.dmn - BKM "Say Hello"
- 0087-chapter-11-example -  BKMs "Eligibility rules" etc

PR #385, opened 2020-12-23
Fix variable.typeRef for BKMs

opatrascoiu:
This is to address https://github.com/dmn-tck/tck/issues/376

There are 3 ways to address above:
- remove typeRef and let the implementation infer the type if needed
- set typeRef to Any
- set typeRef to a proper TFunctionItem introduced in DMN 1.3

This request removes the typeRefs to be consistent with existing test cases.

We need to add extra test cases for remaning options (Any or proper TFunctionItem).

### DIFF
--- a/TestCases/compliance-level-3/0082-feel-coercion/0082-feel-coercion.dmn
+++ b/TestCases/compliance-level-3/0082-feel-coercion/0082-feel-coercion.dmn
@@ -147,8 +147,8 @@
     </decision>
 
     <businessKnowledgeModel name="bkm_004" id="_bkm_004">
-        <variable typeRef="tNumberList" name="bkm_004"/>
-        <encapsulatedLogic>
+        <variable name="bkm_004"/>
+        <encapsulatedLogic typeRef="tNumberList">
             <formalParameter name="arg"/>
             <literalExpression>
                 <text>arg</text>
@@ -157,8 +157,8 @@
     </businessKnowledgeModel>
 
     <businessKnowledgeModel name="bkm_005" id="_bkm_005">
-        <variable typeRef="number" name="bkm_005"/>
-        <encapsulatedLogic>
+        <variable name="bkm_005"/>
+        <encapsulatedLogic typeRef="number">
             <formalParameter name="arg"/>
             <literalExpression>
                 <text>[arg]</text>
--- a/TestCases/compliance-level-3/0087-chapter-11-example/0087-chapter-11-example.dmn
+++ b/TestCases/compliance-level-3/0087-chapter-11-example/0087-chapter-11-example.dmn
@@ -248,7 +248,7 @@
 	</semantic:decision>
 	<semantic:businessKnowledgeModel id="_dc2ca64d-2044-4806-9d0e-e705b3b14447" name="Eligibility rules">
 		<semantic:description>&lt;p&gt;&lt;span style="font-family: arial, helvetica, sans-serif; font-size: 10pt;"&gt;&lt;span lang="JA"&gt;The&amp;nbsp;&lt;/span&gt;&lt;strong&gt;&lt;span lang="JA"&gt;Eligibility rules&amp;nbsp;&lt;/span&gt;&lt;/strong&gt;&lt;span lang="JA"&gt;decision logic&amp;nbsp;defines a complete, priority-ordered single hit decision table&amp;nbsp;&lt;/span&gt;deriving Eligibility from Pre-Bureau Risk Category, Pre-Bureau Affordability and Age.&lt;/span&gt;&lt;/p&gt;</semantic:description>
-		<semantic:variable name="Eligibility rules" id="_fe2d2049-704b-4a88-88be-3d1a56c3d376" typeRef="tEligibility"/>
+		<semantic:variable name="Eligibility rules" id="_fe2d2049-704b-4a88-88be-3d1a56c3d376"/>
 		<semantic:encapsulatedLogic>
 			<semantic:formalParameter name="Pre-Bureau Risk Category" typeRef="tRiskCategory"/>
 			<semantic:formalParameter name="Pre-Bureau Affordability" typeRef="boolean"/>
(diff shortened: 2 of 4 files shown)


TestCases/compliance-level-3/0082-feel-coercion/0082-feel-coercion.dmn
```diff
@@ -157,8 +157,8 @@
     </businessKnowledgeModel>
 
     <businessKnowledgeModel name="bkm_005" id="_bkm_005">
-        <variable typeRef="number" name="bkm_005"/>
-        <encapsulatedLogic>
+        <variable name="bkm_005"/>
+        <encapsulatedLogic typeRef="number">
```
StrayAlien, line 161: Hi @opatrascoiu, in this case, the actual purpose of the test is to assert what an impl does in this exact circumstance - it is not an oversight.  The test description is "bkm type is non-list and bkm logic returns singleton list of conforming value - coercion list to val".  In essence, the logic returns an untyped value and that is coerced to the type of the bkm variable. It is not testing whether an impl does type inference, that is tested elsewhere.

To maintain the intention of the test, I believe the type should remain where it is.

TestCases/compliance-level-3/0082-feel-coercion/0082-feel-coercion.dmn
```diff
@@ -147,8 +147,8 @@
     </decision>
 
     <businessKnowledgeModel name="bkm_004" id="_bkm_004">
-        <variable typeRef="tNumberList" name="bkm_004"/>
-        <encapsulatedLogic>
+        <variable name="bkm_004"/>
+        <encapsulatedLogic typeRef="tNumberList">
```
StrayAlien, line 151: Hi @opatrascoiu - as per the comments on bkm_005.  The positioning of the type is part of the intention of the test. Though, I see the test for it is commented out referencing this: https://github.com/dmn-tck/tck/pull/238#issuecomment-497744315.  I think that was a back-office change and not by me.  It is maybe possible the intention of some tests that comment refers to may also have been misunderstood.   Unsure.

TestCases/compliance-level-3/0087-chapter-11-example/0087-chapter-11-example.dmn
```diff
@@ -248,7 +248,7 @@
 	</semantic:decision>
 	<semantic:businessKnowledgeModel id="_dc2ca64d-2044-4806-9d0e-e705b3b14447" name="Eligibility rules">
 		<semantic:description>&lt;p&gt;&lt;span style="font-family: arial, helvetica, sans-serif; font-size: 10pt;"&gt;&lt;span lang="JA"&gt;The&amp;nbsp;&lt;/span&gt;&lt;strong&gt;&lt;span lang="JA"&gt;Eligibility rules&amp;nbsp;&lt;/span&gt;&lt;/strong&gt;&lt;span lang="JA"&gt;decision logic&amp;nbsp;defines a complete, priority-ordered single hit decision table&amp;nbsp;&lt;/span&gt;deriving Eligibility from Pre-Bureau Risk Category, Pre-Bureau Affordability and Age.&lt;/span&gt;&lt;/p&gt;</semantic:description>
-		<semantic:variable name="Eligibility rules" id="_fe2d2049-704b-4a88-88be-3d1a56c3d376" typeRef="tEligibility"/>
```
StrayAlien, line 251: @opatrascoiu - this is the example from the spec.  My feeling is removing these types from the variables throughout changes the nature of the bkms from one where a variable value may be type-coerced due to the typing on the variable, into one where the variable is type-inferred.  So, whereas with the typings in place, there is a possibility that a null value will be returned because of null-coercion forced by the explicit type.  With the types removed this is no longer the case.

StrayAlien:
@opatrascoiu - regarding the three options, I don't think the first two are suitable as they change the meaning of tests (as noted above), but looking through the DMN spec, I can't quite see where it says that a BKM variable should be a function type.  I'm not saying it should not, just that it is often hard to interpret the tea leaves.  It does say it is a reference:

> An Invocable element contains an InformationItem that holds an invocable reference to the abstract business knowledge, which allows a Decision to invoke it by name

But does that mean it is a function type?

TestCases/compliance-level-3/0087-chapter-11-example/0087-chapter-11-example.dmn
opatrascoiu: I know the example is from the spec, but in my opinion is incorrect. I raised an issue to be discussed in RTF.

Currently is in INBOX
https://issues.omg.org/browse/INBOX-1081

opatrascoiu:
Table 15: Invocable attributes and model associations page 62.

@DanielThanner raised the initial issue (see https://github.com/dmn-tck/tck/issues/376), was discussed in TCK, was discussed in RTF, the agreement was that the type if present should be a function type. DMN 1.2 doesn't have function types, hence TFunctionItem was added in DMN 1.3.

The purpose of this PR is to fix the debatable / incorrect ones and be consistent with the rests of the tests cases. All the other BKMs in the tests don't have a variable.typeRef

TestCases/compliance-level-3/0082-feel-coercion/0082-feel-coercion.dmn
opatrascoiu: I think the description of the test should be ""bkm **return type/ output** type is non-list and bkm logic returns singleton list of conforming value - coercion list to val".

AFAIK variable.typeRef is the type of BKM (Any) -> List<number> encapsulated.typeRef is the return type List<number>.

I don't think moving the typeRef changes the intent of the test case.

StrayAlien: Hi @opatrascoiu. Merry Christmas to you.  If a variable may have a type, and that type may cause a coercion, then it should be the subject of a test case.  That is what this does.   I think maybe the gap is that we may not have a test that does a similar thing to the typeRef on the encapsulated logic.

TestCases/compliance-level-3/0087-chapter-11-example/0087-chapter-11-example.dmn
StrayAlien: Hi @opatrascoiu .  That link requires a login I am afraid.  Though, for the benefit of those without login creds there, I think this is it: https://issues.omg.org/issues/lists/unclassified#issue-47325.  I'm not sure the position of the typeRef is incorrect according to that ticket, having a typeRef on the variable is permissible, but if the typedef of an invokable is supposed to be function type then yes, I get your point.   Maybe leave the typedef where it is, but make them function types? (as per one of the options)

StrayAlien:
Hi @opatrascoiu - that would be (yet another) breaking change for 1.3 then - not sure if the spec team take into account making breaking changes in a public spec - maybe breaking changes should be part of the release notes (if they are, apols, I have missed them).  But okay.  We need more tests there then.  Though, I'm not sure the BKMs need to be 'consistent' in the TCK - having the typeRef on the variable is perfectly valid.  IMO.

opatrascoiu:
@StrayAlien I agree we having typeRefs in BKM.variable is valid. I agree we need tests for TFunctionItem as well, its'a new DMN 1.3 feature that is not covered. 

I think we should create a new test suite dedicated to this aspect to keep them localized. I thought about adding them in this PR but in the end I decided to minimize the changes and removed typeRef to match existing test cases we agreed on (we agreed that BKM.variable.typeRef is optional).

StrayAlien:
@opatrascoiu - maybe some of the tests could be be kept together, but I think they'll often belong with the test grouping - like equality, instance of, conformance - that sort of thing.   

Btw, writing the tests is not hard - getting a PR approved and merged is the hard bit ... I've got one there coming on 12 months old ... rather disheartening ... project feels dead.

But there is a lot of 1.3 not yet covered. way more than just functions.

tarilabs:
So the proposal for identifying the return type, is to move it to the `encapsulatedLogic`, am I getting it correct?

StrayAlien:
Hi @tarilabs - I forgot about this one.  I think the PR solves the issue of the types not being function types in the current test suite, but that will still leave a gap in the TCK if indeed the typeRef of an invokable is to be a function type.  If it so agreed I can get some tests in to assert it.

opatrascoiu:
@tarilabs Yes, that's correct. The type used in the bkm.variable.tyoeRef is the return / output type.

@StrayAlien Would be great to have test cases for the function types. My preference is to have a new test suite (separate xml file)

StrayAlien:
fyi, I have raised the plural nature of the FunctionItem `parameters` attribute.  inbox: https://issues.omg.org/issues/lists/unclassified#issue-47363

PR #388, opened 2021-01-06
DMN 1.3 abs

StrayAlien:
Assertions on the new support for durations.  Also, removed typeRefs from variables - they're serving no purpose here.

### DIFF
--- a/TestCases/compliance-level-3/0050-feel-abs-function/0050-feel-abs-function-test-01.xml
+++ b/TestCases/compliance-level-3/0050-feel-abs-function/0050-feel-abs-function-test-01.xml
@@ -90,18 +90,34 @@
         </resultNode>
     </testCase>
     <testCase id="011">
-        <description>Assert with days duration param</description>
-        <resultNode errorResult="true" name="decision011" type="decision">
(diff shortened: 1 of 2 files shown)


opatrascoiu:
My preference is to keep the typeRefs:
- type inference is faster
- helps to catch errors earlier in most of the cases (static semantics vs dynamic semantics)
- it could helps us catch other errors that are covered by the test cases (e.g. `conforms to` error conversion)

StrayAlien:
Thanks @opatrascoiu - For real models, yes, but in tests I prefer to leave off typeRefs if they are not actually part of what is being tested - keeps it focussed and does not introduce layers that can cause a false positive null-result by (say) introducing a null result in a place that is not the functionality under test. So, in effect, all values as 'passed through' and types do not get in the way, unless that is what is being tested.  Here the typeRefs are not under test.

tarilabs (approved):
This was interesting, thank you and nice catch!
I guess at least we confirmed coercion behaved as expected 😄 
Re-Double-checked with [ref](https://www.w3.org/TR/xmlschema11-2/#yearMonthDuration) and looks good to me 👍

PR #389, opened 2021-01-06
DMN 1.3 equality

StrayAlien:
Assertions for 1.3 changes plus some missing stuff

P0D - P0Y assertion: due to:

> Comparison operators are defined only when the two operands have the same type, except for years and months duration
and days and time duration, which can be compared for equality. Notice, however, that “with the exception of the zerolength
duration, no instance of xs:dayTimeDuration can ever be equal to an instance of xs:yearMonthDuration.

Function equality:

This sure needs more discussion I think as the spec refers to 'body' equality and distinguishes between internal and external functions.  So, for now, I have just added some tests based on identity to get the ball rolling  Comments welcome.

Ranges:

We did not have equality tests on ranges.  Now we have.  Also, the 'unary' syntax is officially a range now.  Before, some implementers (like us), may not have implemented them as ranges, but as a 'unary test' or, indeed, something else. There are assertions that ranges created with either syntax are equal if endpoint/inclusivity etc are equal - regardless of syntax used to create them.   

Comments welcome.

### DIFF
--- a/TestCases/compliance-level-3/0068-feel-equality/0068-feel-equality.dmn
+++ b/TestCases/compliance-level-3/0068-feel-equality/0068-feel-equality.dmn
@@ -803,6 +803,16 @@
         </literalExpression>
     </decision>
 
+    <decision name="dt_duration_008" id="_dt_duration_008">
+        <description>Tests FEEL expression: 'duration("P0D") = duration("P0Y")' and expects result: 'true (boolean)'</description>
+        <question>Result of FEEL expression 'duration("P0D") = duration("P0Y")'?</question>
+        <allowedAnswers>true (boolean)</allowedAnswers>
+        <variable typeRef="boolean" name="dt_duration_008"/>
+        <literalExpression>
+            <text>duration("P0D") = duration("P0Y")</text>
+        </literalExpression>
+    </decision>
+
     <decision name="ym_duration_001" id="_ym_duration_001">
         <description>Tests FEEL expression: 'duration("P1Y") = duration("P1Y")' and expects result: 'true (boolean)'</description>
         <question>Result of FEEL expression 'duration("P1Y") = duration("P1Y")'?</question>
@@ -953,6 +963,83 @@
         </literalExpression>
     </decision>
 
+    <decision name="fn_001" id="_fn_001">
+        <description>Tests FEEL expression: '(function() "foo") = (function() "foo")' and expects result: 'false (boolean)'</description>
+        <question>Tests FEEL expression: '(function() "foo") = (function() "foo")'?</question>
+        <allowedAnswers>false (boolean)</allowedAnswers>
+        <variable typeRef="boolean" name="fn_001"/>
+        <literalExpression>
+            <text>(function() "foo") = (function() "foo")</text>
+        </literalExpression>
+    </decision>
+
+    <decision name="fn_002" id="_fn_002">
+        <description>Tests FEEL expression: "max = mean" and expects result: 'false (boolean)'</description>
+        <question>Tests FEEL expression: 'max = mean"?</question>
+        <allowedAnswers>false (boolean)</allowedAnswers>
+        <variable typeRef="boolean" name="fn_002"/>
+        <literalExpression>
+            <text>max = mean</text>
+        </literalExpression>
+    </decision>
+
+    <decision name="fn_003" id="_fn_003">
+        <description>Tests FEEL expression: "count = count" and expects result: 'true (boolean)'</description>
+        <question>Tests FEEL expression: 'count = count"?</question>
+        <allowedAnswers>true (boolean)</allowedAnswers>
+        <variable typeRef="boolean" name="fn_003"/>
+        <literalExpression>
+            <text>count = count</text>
+        </literalExpression>
+    </decision>
+
+    <businessKnowledgeModel name="fn_bkm_001" id="_fn_bkm_001">
+        <variable name="fn_bkm_001"/>
+        <encapsulatedLogic>
+            <literalExpression>
+                <text>"foo"</text>
+            </literalExpression>
+        </encapsulatedLogic>
+    </businessKnowledgeModel>
+
+    <businessKnowledgeModel name="fn_bkm_002" id="_fn_bkm_002">
+        <variable name="fn_bkm_002"/>
+        <encapsulatedLogic>
+            <literalExpression>
+                <text>"foo"</text>
+            </literalExpression>
+        </encapsulatedLogic>
+    </businessKnowledgeModel>
+
+    <decision name="fn_004" id="_fn_004">
+        <description>Tests FEEL expression: "fn_bkm_001 = fn_bkm_001" and expects result: 'true (boolean)'</description>
+        <question>Tests FEEL expression: 'fn_bkm_001 = fn_bkm_001"?</question>
+        <allowedAnswers>true (boolean)</allowedAnswers>
+        <variable typeRef="boolean" name="fn_004"/>
+        <knowledgeRequirement>
+            <requiredKnowledge href="#_fn_bkm_001"/>
+        </knowledgeRequirement>
+        <literalExpression>
+            <text>fn_bkm_001 = fn_bkm_001</text>
+        </literalExpression>
+    </decision>
+
+    <decision name="fn_005" id="_fn_005">
+        <description>Tests FEEL expression: "fn_bkm_001 = fn_bkm_002" and expects result: 'false (boolean)'</description>
+        <question>Tests FEEL expression: 'fn_bkm_001 = fn_bkm_002"?</question>
+        <allowedAnswers>false (boolean)</allowedAnswers>
+        <variable typeRef="boolean" name="fn_005"/>
+        <knowledgeRequirement>
+            <requiredKnowledge href="#_fn_bkm_001"/>
+        </knowledgeRequirement>
+        <knowledgeRequirement>
+            <requiredKnowledge href="#_fn_bkm_002"/>
+        </knowledgeRequirement>
+        <literalExpression>
+            <text>fn_bkm_001 = fn_bkm_002</text>
+        </literalExpression>
+    </decision>
+
 <!--     <decision name="singleton_001" id="_singleton_001">
         <description>Tests FEEL expression: '[100] = 100' and expects result: 'null (boolean)'</description>
         <question>Result of FEEL expression '[100] = 100'?</question>
@@ -962,6 +1049,96 @@
             <text>[100] = 100</text>
         </literalExpression>
     </decision> -->
+
+    <decision name="range_001" id="_range_001">
+        <description>Tests FEEL expression: "[1..10] = [1..10]" and expects result: 'true (boolean)'</description>
+        <question>Tests FEEL expression: '[1..10] = [1..10]'?</question>
+        <allowedAnswers>true (boolean)</allowedAnswers>
+        <variable typeRef="boolean" name="range_001"/>
+        <literalExpression>
+            <text>[1..10] = [1..10]</text>
+        </literalExpression>
+    </decision>
+
+    <decision name="range_002" id="_range_002">
+        <description>Tests FEEL expression: "[1..10] = [1..11]" and expects result: 'false (boolean)'</description>
+        <question>Tests FEEL expression: '[1..10] = [1..11]'?</question>
+        <allowedAnswers>true (boolean)</allowedAnswers>
+        <variable typeRef="boolean" name="range_002"/>
+        <literalExpression>
+            <text>[1..2] = [1..3]</text>
+        </literalExpression>
+    </decision>
+
+    <decision name="range_003" id="_range_003">
+        <description>Tests FEEL expression: "[2..10] = [1..10]" and expects result: 'false (boolean)'</description>
+        <question>Tests FEEL expression: '[2..10] = [1..10]"?</question>
+        <allowedAnswers>true (boolean)</allowedAnswers>
+        <variable typeRef="boolean" name="range_003"/>
+        <literalExpression>
+            <text>[2..10] = [1..10]</text>
+        </literalExpression>
+    </decision>
+
+    <decision name="range_004" id="_range_004">
+        <description>Tests FEEL expression: "(1..10] = ]1..10]" and expects result: 'true (boolean)'</description>
+        <question>Tests FEEL expression: '(1..10] = ]1..10]"?</question>
+        <allowedAnswers>true (boolean)</allowedAnswers>
+        <variable typeRef="boolean" name="range_004"/>
+        <literalExpression>
+            <text>(1..10] = ]1..10]</text>
+        </literalExpression>
+    </decision>
+
+    <decision name="range_005" id="_range_005">
+        <description>Tests FEEL expression: "[1..10) = [1..10[" and expects result: 'true (boolean)'</description>
+        <question>Tests FEEL expression: '[1..10) = [1..10['?</question>
+        <allowedAnswers>true (boolean)</allowedAnswers>
+        <variable typeRef="boolean" name="range_005"/>
+        <literalExpression>
+            <text>[1..10) = [1..10[</text>
+        </literalExpression>
+    </decision>
+
+    <decision name="range_006" id="_range_006">
+        <description>Tests FEEL expression: "(&lt; 10) = (null..10)" and expects result: 'true (boolean)'</description>
+        <question>Tests FEEL expression: '(&lt; 10) = (null..10)'?</question>
+        <allowedAnswers>true (boolean)</allowedAnswers>
+        <variable typeRef="boolean" name="range_006"/>
+        <literalExpression>
+            <text>(&lt; 10) = (null..10)</text>
+        </literalExpression>
+    </decision>
+
+    <decision name="range_007" id="_range_007">
+        <description>Tests FEEL expression: "(&lt;= 10) = (null..10]" and expects result: 'true (boolean)'</description>
+        <question>Tests FEEL expression: '(&lt;= 10) = (null..10]'?</question>
+        <allowedAnswers>true (boolean)</allowedAnswers>
+        <variable typeRef="boolean" name="range_007"/>
+        <literalExpression>
+            <text>(&lt;= 10) = (null..10]</text>
+        </literalExpression>
+    </decision>
+
+    <decision name="range_008" id="_range_008">
+        <description>Tests FEEL expression: "(&gt; 10) = (10..null)" and expects result: 'true (boolean)'</description>
+        <question>Tests FEEL expression: '(&gt; 10) = (10..null)'?</question>
+        <allowedAnswers>true (boolean)</allowedAnswers>
+        <variable typeRef="boolean" name="range_008"/>
+        <literalExpression>
+            <text>(&gt; 10) = (10..null)</text>
+        </literalExpression>
+    </decision>
+
+    <decision name="range_009" id="_range_009">
+        <description>Tests FEEL expression: "(&gt;= 10) = [10..null)" and expects result: 'true (boolean)'</description>
+        <question>Tests FEEL expression: '(&gt;= 10) = [10..null)'?</question>
+        <allowedAnswers>true (boolean)</allowedAnswers>
+        <variable typeRef="boolean" name="range_009"/>
+        <literalExpression>
+            <text>(&gt;= 10) = [10..null)</text>
+        </literalExpression>
+    </decision>
 
 </definitions>
 
--- a/TestCases/compliance-level-3/0068-feel-equality/0068-feel-equality-test-01.xml
+++ b/TestCases/compliance-level-3/0068-feel-equality/0068-feel-equality-test-01.xml
@@ -727,6 +727,15 @@
         </resultNode>
     </testCase>
 
+    <testCase id="dt_duration_008">
+        <description>zero dt duration equals zero ym duration</description>
+        <resultNode name="dt_duration_008" type="decision">
+            <expected>
+                <value xsi:type="xsd:boolean">true</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
     <testCase id="ym_duration_001">
         <description>same durations are equal</description>
         <resultNode name="ym_duration_001" type="decision">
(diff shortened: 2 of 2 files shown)


TestCases/compliance-level-3/0068-feel-equality/0068-feel-equality.dmn
```diff
@@ -803,6 +803,16 @@
         </literalExpression>
     </decision>
 
+    <decision name="dt_duration_008" id="_dt_duration_008">
+        <description>Tests FEEL expression: 'duration("P0D") = duration("P0Y")' and expects result: 'true (boolean)'</description>
+        <question>Result of FEEL expression 'duration("P0D") = duration("P0Y")'?</question>
+        <allowedAnswers>true (boolean)</allowedAnswers>
+        <variable typeRef="boolean" name="dt_duration_008"/>
+        <literalExpression>
+            <text>duration("P0D") = duration("P0Y")</text>
```
tarilabs, line 812: Unless I am mistaken, the paragraph mentioned in the original Description while PR raised, pertains to
```
9.4 S-FEEL semantics
```

and does not correspond to FEEL semantics.

Friendly reminder, S-FEEL has been deprecated at RTF.

My understanding of FEEL semantic applicable here is:

Duration built-in function:
```
duration(from) | duration string | convert from to a days and time or years and months duration 
```

meaning:
```
duration("P0D") // an instance of days and time duration
```
then
```
duration("P0Y") // an instance of years and months duration
```

and accordingly to type lattice:
image

they are not the same type.

For as much as I believe this is one of the non-business-friendly area of FEEL, my understanding is that the result of this test is `null`, since not complying with the definition for:

```
Table 52: General semantics of equality and inequality
...
49.a | e1 = e2 |   e1 and e2 must both be of the same kind/datatype – both numbers, both strings, etc.
...

**Comment by tarilabs** [comment]
About the "function equality" part I find it very interesting, but I have some doubts.

On other instances it has been mentioned that built-in function are not a symbol per-se, i.e.: my understanding has been, cannot be referenced by their name, only invoked.

Personally, I would be leaning towards what you have currently as a test (even tho I am aware is failing for the Drools runner).

File: TestCases/compliance-level-3/0068-feel-equality/0068-feel-equality.dmn (hunk 1 above)
Comment (StrayAlien) [reply TestCases/compliance-level-3/0068-feel-equality/0068-feel-equality.dmn:812]: Thanks @tarilabs. Yes, i saw it was in the sfeel section, but im not sure that means it should be discarded. Some other things in that section are respected outside of sfeel. The definitions of arithmetic for example, like  op:add-dayTimeDuration etc. They are only discussed in sfeel section but apply in general. 

I was kind of expecting some discussion about this so comments welcome. Im not religious either way but my reading of it was it was not just sfeel, even though it is in that section. And the specific mention of an exception to the type lattice is worth a discussion. 👍

**Comment by StrayAlien** [comment]
I sure have my doubts too! Same. The equality of functions feels like a can of worms, especially when the definition refers to the body. I got that in there to get the ball rolling for clarification.

Re only being able to be invoked. Hmm. Not very lambda ish. If we can pass a bkm about by name, then a built in should be the same. IMO.

But, sticking to equality. Yes. Not sure what is correct here...

File: TestCases/compliance-level-3/0068-feel-equality/0068-feel-equality.dmn (hunk 1 above)
Comment (opatrascoiu) [reply TestCases/compliance-level-3/0068-feel-equality/0068-feel-equality.dmn:812]: Agree with @tarilabs on this

**Comment by opatrascoiu** [comment]
@StrayAlien I don't think `null` endpoints are supported for ranges / intervals. Here is my rationale: endpoints can be only `simple literals` and `qualified name` and 'null' is not a simple literal. See grammar rules below:

```
8. interval = ( open interval start | closed interval start ) , endpoint , ".." , endpoint , ( open interval end | closed interval
end ) ;
16. endpoint = simple value ;
17. simple value = qualified name | simple literal ;
32. simple literal = numeric literal | string literal | boolean literal | date time literal ;
```
A long time ago I raised a RTF ticket to solve some ambiguities & extend the endpoints to be expressions.
https://issues.omg.org/browse/DMN14-36

I think we should comment out these tests until this issue is addressed.

**Comment by opatrascoiu** [comment]
@StrayAlien I don't think `null` endpoints are supported for ranges / intervals. Here is my rationale: endpoints can be only `simple literals` and `qualified name` and 'null' is not a simple literal. See grammar rules below:

```
8. interval = ( open interval start | closed interval start ) , endpoint , ".." , endpoint , ( open interval end | closed interval
end ) ;
16. endpoint = simple value ;
17. simple value = qualified name | simple literal ;
32. simple literal = numeric literal | string literal | boolean literal | date time literal ;
```
A long time ago I raised a RTF ticket to solve some ambiguities & extend the endpoints to be expressions.
https://issues.omg.org/browse/DMN14-36

I think we should comment out these tests until this issue is addressed.

**Comment by StrayAlien** [comment]
Hi @opatrascoiu, thanks. The intention is to test a null value there, not to test the grammar.  I can change those tests to reference a qualified name that returns a null value if that is preferable.

File: TestCases/compliance-level-3/0068-feel-equality/0068-feel-equality.dmn (hunk 1 above)
Comment (StrayAlien) [reply TestCases/compliance-level-3/0068-feel-equality/0068-feel-equality.dmn:812]: just playing "devil's advocate" here. Should the "op:add-dayTimeDuration" (etc) explanations in that section also not apply just because they in the sfeel section?  My personal feeling is that the equality thing does apply, but maybe it is best we seek clarification as to whether it is intended to be sfeel only.  The spec is not exactly a shining example of unambiguous clarity.  To my mind, it would seem odd to make sfeel have a superset of some feel functionality.

Comment (StrayAlien) [reply TestCases/compliance-level-3/0068-feel-equality/0068-feel-equality.dmn:812]: > The semantics of S-FEEL expressions are defined in this section, in terms of the semantics of the XML Schema datatypes and the XQuery 1.0 and XPath 2.0 Data Model datatypes, and in terms of the corresponding functions and operators defined by XQuery 1.0 and XPath 2.0 Functions and Operators (prefixed by “op:” below). **_A complete stand-alone specification of the semantics is to be found in clause 10.3.2, as part of the definition of FEEL. Within the scope of SFEEL,the two definitions are equivalent and equally normative._**

I have to say, I'm not sure exactly what this is saying, but I _think_ it says the operations described for SFEEL and FEEL are the same.  It describes duration equality as conforming to `op:duration-equal` and that has a specific example here https://www.w3.org/TR/xpath-functions-31/#func-duration-equal that shows it:

> The expression op:duration-equal(xs:yearMonthDuration("P0Y"), xs:dayTimeDuration("P0D")) returns true().

Just sayin' ....

**Comment by opatrascoiu** [comment]
@StrayAlien I don't quite understand why the expected value for 

`(function() "foo") = (function() "foo")`

is false.

The semantics of equality in Table 53  states:

```
internal functions must have the same parameters, body, and scope. Externally defined functions must have the same
parameters and external mapping information. 
```

The above function definitions are internal, no parameters and same body. I am missing something ?

**Comment by tarilabs** [comment]
I suppose this was a first cut at it as said:

> for now, I have just added some tests based on identity to get the ball rolling 

frankly, I would lean to agree on the identity definition, since in the spec you have no equivalence defined for the expression per se (unless I missed something)

```
e.g.:
a + b
a  + b
```

I am not sure there is any provision in the spec which would allow us to assert they are equivalent function body (if placed in a function body).

Frankly speaking, I doubt there can be a consensus to define an equivalence for expression at the RTF (it was not possible to reach a consensus for `string()`) and so my pragmatic conclusion would be to lean to what Greg drafted here, based on identity.
That said:
* if someone volunteer to define that at the RTF, it is always welcome.
* if based on the above, there is no consensus on identity equivalence, and therefore the function equivalence tests are to be commented, good with me also.

**Comment by opatrascoiu** [comment]
I have a comment on `max = mean`

I think the expected result for this should be **null** / error: both functions are overloaded, hence resolution based on name only is ambiguous. 

I suggest using `floor = ceiling`

**Comment by opatrascoiu** [comment]
@StrayAlien @tarilabs 

I think we agree the spec is ambiguous regarding the function equality :). I don't mind defining `=` as `is` for functions, probably is our best chance to have a sane behaviour in this area. However, it has to be agreed in RTF. Hence, my proposal is to comment them out until we change the spec.

We should propose extending the existing identity is() function in Table 77 (date and time functions) to support all the semantic domains or have another operator `==` for identity.

**Comment by opatrascoiu** [comment]
@StrayAlien Thanks for investigating the grey areas in equality !!

**Comment by StrayAlien** [comment]
yes, I leaned towards identity as equality for functions - and it is a minefield.  Nice example @tarilabs with the spaces in an expression.  How on earth would we all do equality on a function body!?  Not sure I can think of a language that does - it'll be identity.  I can't comment on `is()`because I don't really understand what it is needed for, and if it is truly identity, why is it only for date/times and not lists, contexts, etc.  Sorry, I am missing something there.

**Comment by opatrascoiu** [comment]
The story behind `is()` is the following: In DMN 1.1 the equality of date & times was defined using identity. A defect was raised as `=`, `<` and `<=` were not consistent and `=` was changed as it is now. To offer support for previous use cases and expose the identity the function `is()` was introduced.

To check equality of expressions all you need the AST and a traversal. The spaces are usually removed during parsing.

**Comment by StrayAlien** [comment]
Thanks @opatrascoiu - so, paraphrasing, now as of 1.3, '=' (and other comparison operators) for dates/etc is defined by the xpath op:*-equals specifications and `is()` compares based on the components of the value, like year, month, and day, for example.

If that is correct, then the changes to equality may be a little sobering - especially if you're relying on `equals()` in Java date/time types.  For example, Java equals is doing what `is()` does now (if I understand you correctly), not what xpath spec says `=` should.  Online xpath tester give `true` for 

```
xsd:dateTime("2002-04-02T12:00:00-01:00") = xsd:dateTime("2002-04-02T17:00:00+04:00")
```

... Java will give `false`.  Looks like we need those tests!

```
    @Test
    public void dateTimeEquals() {

        LocalDate foo = LocalDate.of(2000, 10, 10);

        // "2002-04-02T12:00:00-01:00"
        ZonedDateTime d1 = ZonedDateTime.of(2002, 04, 02, 12, 0, 0, 0, ZoneOffset.ofHours(-1));

        // "2002-04-02T17:00:00+04:00"
        ZonedDateTime d2 = ZonedDateTime.of(2002, 04, 02, 17, 0, 0, 0, ZoneOffset.ofHours(4));

        // java says false, xpath says true
        Assert.assertFalse(d1.equals(d2));
    }

    @Test
    public void xpathDateTimeEquals() throws Exception {

        String xml = "<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"no\"?>" +
                "<foo xmlns:xsd=\"http://www.w3.org/2001/XMLSchema\"></foo>";

        String expr = "xsd:dateTime(\"2002-04-02T12:00:00-01:00\") = xsd:dateTime(\"2002-04-02T17:00:00+04:00\")";

        Processor processor = new Processor(false);
        XdmNode xdm = processor.newDocumentBuilder().build(new StreamSource(new StringReader(xml)));

        XPathCompiler c = processor.newXPathCompiler();
        c.declareNamespace("xsd", "http://www.w3.org/2001/XMLSchema");

        XdmValue result = c.evaluate(expr, xdm);
        Assert.assertTrue(result.getUnderlyingValue().equals(BooleanValue.TRUE));
    }
    ```

StrayAlien:
As a small addendum to the above which also applies to comparisons and the current arithmetic PR. I have changed our Java runtime to perform both stock Java stuff and also (at the same time) an equivalent xpath based operation for same operands and operator based on the latest saxon xpath impl - and to fail if they differ. 

Im happy to report that (with some caveats) all current tests also comply with xpath.  I'm almost surprised. 😁

The caveats are essentially around DMN usage of '@' time zones on datetime and time and the requirement to assume datetimes with no time zone are utc. The '@' time zones need conversion to an offset to feed into xpath, and conversion back again afterwards. 

Ill keep the 'compare with xpath' mod in to  get some more tests in re equality and comparisons and make sure we're sweet there. 

Happy to share code if anybody wants it.

opatrascoiu:
Can someone remind me why the `time` and `date and time` test cases are commented out ? We need coverage in this space.

Thanks

StrayAlien:
Hi @opatrascoiu - that was because of the uncertainties around equality with timezones etc.  I have to admit, I did not understand at the time why they were commented out - but I reckon I get it now.,  Hopefully.  I think some of them are actually wrong anyways given the 1.3 update.  Anyways, for sure, I'll get those reinstated or at least get that covered.  At the moment, I'm just looking at moving (erm duplicating for now) the equality tests into a new suite with auto-generated tests.  So that, for example, each equality test gets an automatic inequality test as well.  Stuff like that.

opatrascoiu:
@StrayAlien Sounds good to me. Please bear in mind that there were some changes in this area in DMN 1.3: the relational operators are based on value() functions and the precision for `time` and `date and time` is `seconds`. There is a commented test in the above file that assumes the precision is milliseconds.

tarilabs:
> Can someone remind me why the time and date and time test cases are commented out ? We need coverage in this space.
> Thanks

If you can give a reference to 1 example (copying permalink from github line for instance) I am always happy to dig into the meetings notes, to find back why we agreed on the comment

tarilabs:
I wil raise JIRA for OMG RTF to discuss
 - eq of Fn by Identity or Equality or null ?
 - what happens if you `=` two built-in Fns which have overloading: do they equal ?

StrayAlien:
function equality commented out as per TCK meeting discussion 2021-02-13.  Function body equality is not specified, nor is how a built-in function may or may not be equal to itself if it is an overridden function.

StrayAlien:
Also, P0D == P0Y test commented out.

StrayAlien:
Guys, I have added the date / time / datetime tests back in and bolstered them with some more to cater for the changes in DMN 1.3 equality.  I have also removed the typeref='boolean' for all tests.  It'd be great to get your feedback on this stuff kind soon as it is pretty important to comparison and therefore more tests around range functions.

As a note, as DMN supports a time with at zone like `10:15:00@Europe/Paris` - which is not a hard offset like `+01:00` it is possible that some implementations may use the equivalent of `now()` to get a zoned date time to get an offset from.  This is a little dangerous because the meaning of `@Europe/Paris` and its offset changes during the year.  So a test like `@"0:15:00+01:00" = @"10:15:00@Europe/Paris"` will be correct only for some part of the year.  At other times the offset will be `+02:00`.  So, my suggestion is, do not use `now()` with DMN zoned times, use a standard epoch date like 1970-01-01T00:00:00 to get an offset, then it'll be constant.  The spec does not say that, but, maybe it should.

Also, I have a test in here for a 'local date' = 'the same date at UTC'.  This is the spec section `10.3.2.3.5 date`:

> Where necessary, including the __valuedt__ function (see 10.3.2.3.6), a date value is considered to be equivalent to a date time value in which the time of day is UTC midnight (00:00:00).

So for the purposes of equality (and inequality) a local date is considered to be a UTC zero time date.

opatrascoiu:
@StrayAlien
There are still a few remaining issues with this PR:
1.  There are 2 test cases with same id     `<testCase id="datetime_014">`
2.  Test decisions range_006, range_007, range_008, range_009 have syntactic errors (see grammar rules for interval above). The extension has been discussed in the last meeting and it looks like is going in DMN 1.4. All I have to do is prepare the formal proposal and start the discussion again.  I suggest we change the expected result to **null** as this is the correct value in DMN 1.3.

opatrascoiu:
I have some reservations about datetime_014.  

Yes, there an equivalence relationship defined between date and date and date. However, according to `Table 52: General semantics of equality and inequality`, the compared values have to have the same kind / type. `date` and `date and time` are different kinds :)  Also, in `Table 54: Specific semantics of inequality` is not possible to compare date and date and time via <. Operators = and < should be defined in similar ways, otherwise there will be confusion. It's weird to be able to compare via = two date / time types (after applying an implicit conversion) but not via <.

We need to specify very precisely where the implicit conversion `date to date and time` is applied. Right now is too vague. I was planning to address this kind of ambiguities via a proposal  (define a new implicit conversion and add it in all the places) but I had to postpone the discussion as there is going to be another proposal in this area (partial date / time) and it makes sense to discuss them in the same time.

Until we have the above discussion, I am inclining towards adding tests that match the above tables - the phrase "Where necessary" from the definition of the equivalence relationship is too vague. We'll probably change the the tables in DMN 1.4 :) If we agree on following the tables we should keep the datetime_014 test and change the expected result to **null**.

StrayAlien:
Thanks @opatrascoiu.  I have fixed the duplicate test name, and as well as (rather than assert null), I have provided a null value as info req to those range tests so the meaning of the test is maintained.  So the null value is provided as a variable, not as a literal null. Thanks again.

Re  `datetime_014`, I was figuring that might get some comment! :-). Thank you.  And yes, a spec having "Where necessary"  in it made me laugh when I saw it.  "Where necessary"?!!.  Nice.

Actually, my reading of it is that is should also apply to inequality, not just equality.  I think it makes "business sense" to be able to check if a date is (say) between two datetimes.  

My interpretation of `Table 54: Specific semantics of inequality` is that it is possible to compare a date and a datetime.  The datetime inequality spec is based on value(dt), and the semantics of date say it acts like a datetime for value(dt).  So, the `<`, and other operators would work like `@"2021-02-10" < @"2021-02-10T12:00:00"` is `true` for example.  It seems a natural thing to do and seems fine to me.

Maybe though, for DMN 1.4 the entry for `date` in table 54 should reference `10.3.2.3.5 date` and a value(d).  And yes, actually specify where the implicit conversion can be used.  Though, I would say broadly any use of a date where a value(dt) is used may do an implicit conversion - and I think the spec already says that mostly, or I think maybe that was the intention.

opatrascoiu:
@StrayAlien @tarilabs 
I think this PR is very close to be merged. IMO there are 2 outstanding issues:
- the typeRef of InputData null_input is missing. Adding it as "number" should solve it
- the expected result for tests datetime_014 datetime_015 is incorrect as it breaches the semantics described in Table 52: General semantics of equality and inequality.

tarilabs:
I see some test cases for time,date,date and time comparison with TZ and non-TZ are reinstated in the test cases.

Let's focus on one example.
Accordingly to DMNv1.3 if you checked for 
`date and time("2018-12-08T00:00:00") < date and time("2018-12-08T00:00:00@Asia/Dhaka")`
it would say `false`,
because for the DMNv1.3 Table 54: Specific semantics of inequality they are actually expected to be EQUAL when performing this inequality comparison.
Test `datetime_007`.
This test checks for `date and time("2018-12-08T00:00:00") = date and time("2018-12-08T00:00:00@Asia/Dhaka")` and expect `false`.
Which if taken in isolation I could understand the reason why, but as demonstrated is quite ambiguous when considering the full DMNv1.3.
In simpler terms considering DMNv1.3: ~Are they different? No as they are equal. Ah so they are equal? No. 🤷‍♂️ 

This is the part where we already know the spec is **not** consistent.
I don't think it make sense we "insists" on these tests when we are aware the RTF has ongoing work for these kind of time,date, date and time topics.

btw, line 544 in the `.dmn` file has a spurious trailing space which needs to be amended anyway on the basis that variable name must correspond to the DRG parent name.

StrayAlien:
Thanks @opatrascoiu / @tarilabs .

@opatrascoiu - the datetime_014 and datetime_015 tests are to cover section `10.3.2.3.5`.  To me, it makes 'business' sense.

> Where necessary, including the valuedt function (see 10.3.2.3.6), a date value is considered to be equivalent to a date time
value in which the time of day is UTC midnight (00:00:00).

And 'equality' implies 'inequality' and thus I think we should be able to include that in other comparison operators.  but, it may be contentious.  If it is the subject of RTF discussion maybe we can comment them out.  

@tarilabs - I think I get your point.  The equality op for date time does not make the same statement `If one input has a null timezone offset, that input uses the timezone offset of the other input.` that inequality does.  Is there a series of tests we should comment out (again!) subject to further clarification?  

Also, should we comment out the range tests range_006 -> range_009 as they deal with the 'null equality' in ranges?  or is that stuff okay by you guys.

btw, I may be late to tonight's meeting.  Looks like we're likely to go back into lockdown and today is Mother's birthday - so getting out for a meal while we can!  I'll try to be on time.

tarilabs:
@StrayAlien family first.

Yes, imho, the tests which are subject to this topic and shall be commented out, or that _I believe_ are actually wrong and shall be commented out (or receive another evaluation results), are:
```
datetime_007
datetime_014
datetime_015
time_007
time_008
time_009_a
time_009_b
```

> Also, should we comment out the range tests range_006 -> range_009 as they deal with the 'null equality' in ranges?

TBH I didn't have time to dig into those ones yet.
On a brief look from formality PoV, I agree with Octavian the InputData node in the DMN .xml shall receive `typeRef="Any"` in the InputData's variable, and the TCK test case xml is not complying with supplying the InformationRequirement accordingly to the DRG, albeit `null` value, e.g. the test cases are missing
```
        <inputNode name="null_input">
            <value xsi:nil="true"/>
        </inputNode>
```
but as mentioned I will need to further dig into the test semantics itself.

opatrascoiu:
@StrayAlien @tarilabs I suggest we discuss all tests one by one and figure out what needs to be done (e.g. keep them, comment them or raise an RTF to clarify the meaning)

tarilabs:
Here is what I detected so far.
(beyond some formalisms correction raised in https://github.com/StrayAlien/tck/pull/7 so a Squash+Merge there, should solve them).

# datetime_007

Described in https://github.com/dmn-tck/tck/pull/389#issuecomment-879790279
Me, I would comment it out.

# datetime_014

described in https://github.com/dmn-tck/tck/pull/389#issuecomment-879738664
Me, I would comment it out.

# datetime_015

described in https://github.com/dmn-tck/tck/pull/389#issuecomment-879738664
Me, I would comment it out.

# time_007

https://github.com/dmn-tck/tck/blob/f27c2d7f4ae3228d300d18fcc096c5c9833cbc95/TestCases/compliance-level-3/0068-feel-equality/0068-feel-equality.dmn#L600

the spec:

> Table 53: Specific semantics of equality
>> time | value(e1) = value(e2). Value is defined in 10.3.2.3.4
>>> as the number of seconds since midnight

so semantically on paper:

```
valuet(time("10:30:00@Europe/Paris")) = valuet(time("10:30:00"))
37800 = 37800
true
```

the test case atm:

https://github.com/dmn-tck/tck/blob/f27c2d7f4ae3228d300d18fcc096c5c9833cbc95/TestCases/compliance-level-3/0068-feel-equality/0068-feel-equality-test-01.xml#L532-L539

Me, I would either correct the test case (expect `true`), or comment it out on the same premises of the discussed `datetime_...` cases.

# time_008

Analogous consideration of `time_007`.
Me, I would either correct the test case (expect `true`), or comment it out on the same premises of the discussed `datetime_...` cases.

# time_009_a

Analogous consideration of `time_007`.
Me, I would either correct the test case (expect `true`), or comment it out on the same premises of the discussed `datetime_...` cases.

# time_009_b

Analogous consideration of `time_007`.
Me, I would either correct the test case (expect `true`), or comment it out on the same premises of the discussed `datetime_...` cases.

---

I hope this helps :)

tarilabs:
Adding to Milestone since https://github.com/dmn-tck/tck/pull/389#issuecomment-881322326 was act upon, with 
- https://github.com/StrayAlien/tck/pull/7
- a06ef7b

tarilabs:
discussed in meeting TCK 2021-10-14, merging.

PR #390, opened 2021-01-06
DMN 1.3 instance of

StrayAlien:
DMN 1.3 instance of changes - and some stuff

Some previously-commented tests have been uncommented now that types are a supported syntax. 

A series of tests for list, context and function type conformity.

Note (very oddly) there is no syntax in the DMN 1.3 spec for range types, so we officially can't check (say) if `[1..2] instance of range<number>`.

Also, it is not possible to check if something is simply a context or a function.  The described grammar for context requires _at least_ one property.  So, it is not possible to assert (say) `{foo: "bar"} instance of context`, or even worse, `{} instance of context`.  There seems to be no way to describe an empty context type.  So, that is a gap (correct me if I am wrong) and thus there are no assertions included here for same.

Re functions, the same, there is no grammar to describe just a function.  So (say) `function() "foo" instance of function` is not possible, you have to supply params and a return type.  And `function<> -> any` does not mean 'any function', it means a function with no params that returns any - quite different.  So, again, no assertions here for the 'is a function' meaning.

The function covariance/contravariance stuff might be pretty hard to read so hopefully they make sense.  And hopefully they are correct! :-)    

Also, I have added some assertions proving that allowed values on items definition types to not affect conformity.  The spec does not include any notion of allowed values in the equivalence section so I thought it best to assert that as that could be a point of 'implicit difference' between runtimes.

Comments welcome.

### DIFF
--- a/TestCases/compliance-level-3/0070-feel-instance-of/0070-feel-instance-of.dmn
+++ b/TestCases/compliance-level-3/0070-feel-instance-of/0070-feel-instance-of.dmn
@@ -1340,6 +1340,13 @@
         </literalExpression>
     </decision>
 
+    <decision name="function_031" id="_function_031">
+        <variable name="function_031"/>
+        <literalExpression>
+            <text>[(function(a: string, b: number) "123"), "foo"] instance of tListOfFunctions</text>
+        </literalExpression>
+    </decision>
+
 
 </definitions>
 
--- a/TestCases/compliance-level-3/0070-feel-instance-of/0070-feel-instance-of.dmn
+++ b/TestCases/compliance-level-3/0070-feel-instance-of/0070-feel-instance-of.dmn
@@ -85,7 +85,7 @@
     <decision name="null_008" id="_null_008">
         <variable name="null_008"/>
         <literalExpression>
-            <text>null instance of list&lt;any&gt;</text>
+            <text>null instance of list&lt;Any&gt;</text>
         </literalExpression>
     </decision>
 
@@ -114,8 +114,8 @@
     <decision name="null_012" id="_null_012">
         <variable name="null_012"/>
         <literalExpression>
-            <!-- function<> -> any -->
-            <text>null instance of function&lt;&gt; -&gt; any</text>
+            <!-- function<> -> Any -->
+            <text>null instance of function&lt;&gt; -&gt; Any</text>
         </literalExpression>
     </decision>
 
@@ -171,7 +171,7 @@
     <decision name="number_008" id="_number_008">
         <variable name="number_008"/>
         <literalExpression>
-            <text>123.01 instance of list&lt;any&gt;</text>
+            <text>123.01 instance of list&lt;Any&gt;</text>
         </literalExpression>
     </decision>
 
@@ -199,7 +199,7 @@
     <decision name="number_012" id="_number_012">
         <variable name="number_012"/>
         <literalExpression>
-            <text>123.01 instance of function&lt;&gt;-&gt;any</text>
+            <text>123.01 instance of function&lt;&gt;-&gt;Any</text>
         </literalExpression>
     </decision>
 
@@ -263,7 +263,7 @@
     <decision name="string_008" id="_string_008">
         <variable name="string_008"/>
         <literalExpression>
-            <text>"foo" instance of list&lt;any&gt;</text>
+            <text>"foo" instance of list&lt;Any&gt;</text>
         </literalExpression>
     </decision>
 
@@ -291,7 +291,7 @@
     <decision name="string_012" id="_string_012">
         <variable name="string_012"/>
         <literalExpression>
-            <text>"foo" instance of function&lt;&gt;-&gt;any</text>
+            <text>"foo" instance of function&lt;&gt;-&gt;Any</text>
         </literalExpression>
     </decision>
 
@@ -356,7 +356,7 @@
     <decision name="boolean_008" id="_boolean_008">
         <variable name="boolean_008"/>
         <literalExpression>
-            <text>true instance of list&lt;any&gt;</text>
+            <text>true instance of list&lt;Any&gt;</text>
         </literalExpression>
     </decision>
 
@@ -386,7 +386,7 @@
     <decision name="boolean_012" id="_boolean_012">
         <variable name="boolean_012"/>
         <literalExpression>
-            <text>true instance of function&lt;&gt;-&gt;any</text>
+            <text>true instance of function&lt;&gt;-&gt;Any</text>
         </literalExpression>
     </decision>
 
@@ -443,7 +443,7 @@
     <decision name="date_008" id="_date_008">
         <variable name="date_008"/>
         <literalExpression>
-            <text>date("2018-12-08") instance of list&lt;any&gt;</text>
+            <text>date("2018-12-08") instance of list&lt;Any&gt;</text>
         </literalExpression>
     </decision>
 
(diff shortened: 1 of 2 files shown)


opatrascoiu:
I've noticed that some of the types use `any` in lower cases. Is that intentional? The correct name is `Any`.

StrayAlien:
Thanks @opatrascoiu - corrected.  Always seems strange to me that other types like 'string' are lowercase and any is 'Any'.  Our runtime supports both.

opatrascoiu:
My initial proposal for the type lattice was to user lowercases, however the final decision was to have `Any` and `Null` types with upper case as they are the upper and lower bound in the lattice and `null` was already used as literal.

Also, I believe the types and variable names are case sensitive. I don't think we have test cases for that.

StrayAlien:
@opatrascoiu - re case sensitivity, the spec (as far as I can see) makes no mention of it at all.  For a business-friendly language, it would seem odd to me to make it case sensitive. But, agree, if that is the case we should definitely have tests for it.

StrayAlien:
Some test additions to cater for creating types using DMN 1.3 FunctionItems.  Comments welcome.

A slightly hairy thing about this stuff is determining the type of a list of functions (or indeed the type of a list of stuff that isn't (say) just numbers) .... I've put in a couple of assertions but this is a fairly complex thing to think about.  The spec does not consider type intersection.

I think some additional tests are needed in the 'coercion' suite to deal with invokables having FunctionItem typeRefs and what to do when they do not match.

As always, comments welcome.

StrayAlien:
@tarilabs / @opatrascoiu  - some of the tests I've put in here use a syntax that I do not believe is DMN compliant, so, I'll have to update them to use function itemdefinitions.  

The types on named params for a function definition do not support the same syntax as the types in instanceof so it not yet possible to do 

```
function(a: list<string> ..etc)
```

It seems the types here can only be names.  I'll get that amended and raise an issue to rectify.

StrayAlien:
Just looking at this again, I think the above `function(a: list<string> ..etc)` is indeed legal - not sure what I was thinking.  According to rules 55, 56, 52.  A function definition has formal params and formal params are `parameter name [":" type ] ;`.  'type' here being the new syntax.  We're sweet.  I might update/revert these tests to reflect that.

StrayAlien:
Well as it happens, it seems I neither took them out, not put them back in.  So sweet here.

TestCases/compliance-level-3/0070-feel-instance-of/0070-feel-instance-of.dmn
```diff
@@ -173,19 +227,27 @@
...
-            <text>123.01 instance of function</text>
+            <text>123.01 instance of function&lt;&gt;-&gt;Any</text>
+        </literalExpression>
+    </decision>
+
+    <decision name="number_013" id="_number_013">
+        <variable name="number_013"/>
+        <literalExpression>
+            <!-- true: does not take allowedValues into account -->
+            <text>256 instance of t255</text>
```
tarilabs, line 248: We are revising this PR/tests, and while I fully agree with this test, I am wondering how to best report this back to the RTF.

On one side, `allowedValues` is a concept of the DMN metamodel.
On the other side, DMN specification at a point introduced for `instance of` to be working with the types defined in the DMN model, albeit there is no way to do equivalently in pure FEEL.

As an end-result, this case is uncovered which might be a bit counter-intuitive...

tarilabs:
Hi @StrayAlien we definitely need https://github.com/StrayAlien/tck/pull/9 for two purposes:

1. as spotted by @tkobayas likely in `context_014` it was meant something else based on the comment.
2. since this PR is now introducing reference to USER type ItemDefinition in this model, need to reference them in the correct namespace, meaning the same correction as of https://github.com/dmn-tck/tck/pull/422#issuecomment-877004141 "point 1." is needed also here.

The PR https://github.com/StrayAlien/tck/pull/9 fixes these issues.

StrayAlien:
@tarilabs, @tkobayas - that is merged in - thanks.  Something also (noted in the original post here), further to the TCK meeting discussion on types and some vagaries in the spec is that AFAICT there is no means to specify 'any context', or 'any function'.  Like for a list we can go `list<Any>`, but the grammar for a context type requires at least one name `'context' '<' name ':' type { ',' name ':' type } '>' |`. It is not possible to simply say `{foo: "bar"} instance of context`.   So, that means stuff like it is not possible to have a list type that contains contexts - any context - they must have at least one field in common.

Same for functions.  It is not possible to simply test if something is a function like `x instance of function`.  

Unless, of course, this has been fixed in 1.4 ...

tarilabs:
Per TCK meeting 2022-01-11, some tests here related to functions allegedly have no ground in any capability specified in the DMN spec.

Specifically,

- using FEEL "anonymous functions" definition, by DMNv1.3, there is no way to specify the signature return type (but there way to specify parameter type)
- using boxed expression in DMN, that become possible, either using a ItemDefinition for a FunctionItem, or with a boxed Function Definition in a BKM for instance.

So the problem of these tests lies in the first case.

For example in FEEL

```
function(a: number) 47 + a
```

we, as human, are led to say that the return of this lambda is a number, but there is no provision in the spec to determine the type in this way, "statically".

Similarly for `function() 123`, even when the body of the lambda is a number literal of FEEL, there is no provision in the spec to determine statically based on the "content" of the function, as the lambda is considered a black box.

Similarly more complex case, `function() 123 + xyz` where `xyz` is something in the closure's context, same principles apply.

Granted, when the value out of the value invocation, is checked for type conformance in the hosting boxed logic, then as usual the provision of the spec will kick in, but to be noticed that is not inherently to the lambda definition in FEEL (see above).

Granted, a Vendor is always free to implement _in addition_ to the specification (like walking the AST of the body of the function to infer the return type at compile-time), but that is beyond the role of the TCK tests.

Conclusion: those tests relying on the anonymous function return type "inferred" when it's actually not declared by the signature, are not conformant to the DMN specification, and shall better be removed.

Specifically re https://github.com/dmn-tck/tck/pull/390#issuecomment-1010394634, in Drools we maintained that backward compatibility and so far no tests in the TCK violates the b/c principle, if that helps the analysis. However truly, to be formally backward compatible, it might be better to address the case in the spec.

StrayAlien:
Hi @tarilabs - Thanks.  IMO, I don't see an untyped lambda as being any different to a BKM that has no explicit typeRef.  Indeed, other literal expression elements have no explicit type either but we infer their type for tests - like a list of 'similar' contexts has no defined type but we infer that - it is not defined anywhere in the spec whether we take the context type of the most specific, or least specific (I think it _should_ be defined in the spec . We're not typing it like `<list-type>[list-type instance, subtype of list-type instance ]`.  Not sure why we need to with lambdas.

But, to help get the PR through, I have commented out most the new function-related tests here.  Which is a bit of a shame as they exercised the notions of type conformance and the covariant / contravariant parts of the spec.  

I guess they could be rewritten to use BKMs, but this PR is already more than a year old, so I'll leave investing more time on it to when/if it gets through.

Thanks again

tarilabs:
I see your perspective Greg, but I hope the concrete examples in https://github.com/dmn-tck/tck/pull/390#issuecomment-1053664450 should highlight the difference between a BKM as a boxed expression (with an associated typeRef for ItemDefinition of FunctionItem) and an anonymous function definition as part of a FEEL expression.

baldimir:
Hi @StrayAlien @tarilabs, what is the status of this please?

tarilabs:
If the memory serves me well, some tests where debatable on the rationales explained in https://github.com/dmn-tck/tck/pull/390#issuecomment-1053664450

StrayAlien:
@baldimir - in the interest of getting the PR moving those tests mentioned by @tarilabs were commented out Jan 6, 2021.  I can't say I exactly agree :-) but better to get some tests in rather than none.  Perhaps covariance/contravariance can be tested in other ways.

baldimir:
As those tests are commented out, @tarilabs could you please approve?

baldimir:
@StrayAlien, I plan to take a look at this next week. Does someone from @dmn-tck/contributors has apart from the existing comments any current feedback on this please?

PR #393, opened 2021-01-06
DMN1-3 feel in

StrayAlien:
Some additional assertions on what happens when nulls are part of ranges and when endpoints are null-coerced and incompatible.

The unary syntax for range and equivalence thereof with the non unary syntax kind of means non unary ranges can act like unary ranges.

Note the specs do not seem to support the usage on a unary syntax range with the 'in' keyword, so no assertions for that are made here - just what happens with nulls in non-unary ranges.

Comments welcome.

### DIFF
--- a/TestCases/compliance-level-3/0072-feel-in/0072-feel-in-test-01.xml
+++ b/TestCases/compliance-level-3/0072-feel-in/0072-feel-in-test-01.xml
@@ -2542,6 +2542,7 @@
         </resultNode>
     </testCase>
 
+<!--
     <testCase id="null_001">
         <description>range null test value</description>
         <resultNode errorResult="true" name="null_001" type="decision">
@@ -2586,13 +2587,15 @@
             </expected>
         </resultNode>
     </testCase>
+-->
 
     <testCase id="null_002">
         <!-- as null is a valid value for a range start, this test asserts that a null start caused by a
         invalid type coercion is not the same as simply supplying a null value -->
-        <description>range - invalid start type coerces to null and makes range check invalid</description>
+        <description>range - invalid inclusive start type coerces to null and makes range check invalid</description>
 
         <!-- invalid value provided as input so compiler type inference (if any) is not relevant -->
+        <!-- the range has a number end and we're supplying a list as the start -->
         <inputNode name="input_for_null_tests">
             <list></list>
         </inputNode>
@@ -2605,11 +2608,12 @@
     </testCase>
 
     <testCase id="null_003">
-        <!-- as null is a valid value for a range end, this test asserts that a null end caused by a
+        <!-- as null is a valid value for a range start, this test asserts that a null start caused by a
         invalid type coercion is not the same as simply supplying a null value -->
-        <description>range - invalid end type coerces to null and makes range check invalid</description>
+        <description>range - invalid non-inclusive start type coerces to null and makes range check invalid</description>
 
         <!-- invalid value provided as input so compiler type inference (if any) is not relevant -->
+        <!-- the range has a number end and we're supplying a list as the start -->
         <inputNode name="input_for_null_tests">
             <list></list>
         </inputNode>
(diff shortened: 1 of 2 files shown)


tarilabs (commented):
I am looking into this, and I believe we just need a convergence on one "use-case"

the PR says:
```
  null_001_c: 5 in [1..null) = null // and an error
  null_001_d: 5 in [null..10] = null // and an error
```

but for me and looking at the spec,
that would converge the range check to be equivalently:
```
>=1
<=10
```

respectively, hence true in both cases, like you have for the other `(null...` literal ranges.

Or at least, I don't find in the spec any ground to support the case they are not that equivalent.

What am I missing?

opatrascoiu:
@tarilabs AFAIK [1..null) is syntactically incorrect. DMN 1.3 FEEL grammar does not support `null` as endpoints.

tarilabs:
> https://github.com/dmn-tck/tck/pull/393#issuecomment-796862271
the `null`-literal I assume specifically.

Fair comment

In that case decision nodes and test cases `null_001_(a|b|c|d)` ought to be removed.

StrayAlien:
Hi @tarilabs / @opatrascoiu . Thanks.  This tests are assertions on whether an endpoint can be both inclusive _and_ null at the same time.

If the inclusivity flag is true, and the respective endpoint is null, then we cannot check it for inclusion as any comparison against null is null. 

So `null_001_c: 5 in [1..null] = null` (btw, your post above had it as `[1..null)` - note the different inclusivity) would be trying to check if `5 <= null`.  Likewise with `5 in [null..10]`, that needs to check if `5 >= null`.

It also kind of ties in with this:

> If the syntax with a single endpoint and an operator is used, then the other endpoint is undefined (represented by a null value) and the inclusivity flag is set to _false_.

The `<= 10` style syntax equates to one endpoint inclusivity flag being set to false to cater for the above.  Or that is my take on it at least.

That was my thinking - hopefully that makes sense.  Happy for thoughts.

Greg.

StrayAlien:
Rather than remove those tests, I have updated them to accept an input that will resolve to null.  I did that for a few other range tests but missed these.  This way, we keep the tests yet have a supported syntax.

StrayAlien:
btw, thinking over my response above re `5 <= null` and so on - not quite sure I have it right.  For sure, the spec is not helpful here.  Let's discuss.

StrayAlien:
as per our conv at TCK meeting, most assertions for null end points have been removed.  I have left in the ones that assert the range is null if an invalid endpoint type is used (like a list).  Happy for comments re that.

tarilabs:
🔗  https://github.com/dmn-tck/tck/discussions/429#discussioncomment-844147

tarilabs:
TCK meeting 2021-06-10 agreed to raise to RTF about https://github.com/dmn-tck/tck/pull/393/files#diff-c74d64cde5bb0def8be48cb9a120bb79c905d2d9bf245498a3daaa17a7346056R2069-R2107 and understand of the group.
Greg will send also some link about Eq comparison of ranges
then this PR will be marked as waiting-for-rTF

StrayAlien:
@tarilabs - the range equality tests I spoke about are in https://github.com/dmn-tck/tck/pull/389.  tests `range_006` to `range_009`.  They explore the equality between the two range syntaxes like, for example, whether `(>= 10) = [10..null_input)` is true.

StrayAlien:
@tarilabs - I think this one can be tagged as "waiting on RTF".

tarilabs:
Per https://github.com/dmn-tck/tck/pull/393#issuecomment-858580194 ,

raised as: https://issues.omg.org/browse/INBOX-1337 , https://issues.omg.org/browse/INBOX-1338 .

As with https://github.com/dmn-tck/tck/pull/393#issuecomment-885956418 , agreed, but I do also believe you can directly erase lines:
https://github.com/dmn-tck/tck/pull/393/files#diff-c74d64cde5bb0def8be48cb9a120bb79c905d2d9bf245498a3daaa17a7346056R2039-R2065
since Octavian rightfully pointed out in https://github.com/dmn-tck/tck/pull/393#issuecomment-796862271 too.

StrayAlien:
@tarilabs / @opatrascoiu - if you now have some clarification from the RTF, rather than remove those tests, would it not be more useful to uncomment them, but ... change the usage of `null` to use an input that resolves to null ? - thus keeping the meaning of the tests.  Seems a `null` literal is not permitted there, but null values are (?).

baldimir:
@StrayAlien could you please rebase this? As this was clarified in the RTF group, we may review this. @dmn-tck/contributors

opatrascoiu:
I checked the test cases that are not commented out - they look fine to me. We should review the commented ones on the DMN 1.4 branch.

PR #394, opened 2021-01-06
DMN1-3 range properties

StrayAlien:
Assertions on the new 1.3 range properties for both range syntaxes.

Note the unary style syntax are disambiguated with ().  Otherwise, the dot property will be taken to be part of the end range expression.

### DIFF
--- a/TestCases/compliance-level-3/0074-feel-properties/0074-feel-properties-test-01.xml
+++ b/TestCases/compliance-level-3/0074-feel-properties/0074-feel-properties-test-01.xml
@@ -407,4 +407,220 @@
         </resultNode>
     </testCase>
 
+    <testCase id="range_001">
+        <description>range - start</description>
+        <resultNode name="range_001" type="decision">
+            <expected>
+                <value xsi:type="xsd:decimal">1</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="range_002">
+        <description>range - end</description>
+        <resultNode name="range_002" type="decision">
+            <expected>
+                <value xsi:type="xsd:decimal">10</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="range_003">
+        <description>range - start included</description>
+        <resultNode name="range_003" type="decision">
+            <expected>
+                <value xsi:type="xsd:boolean">true</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="range_004">
+        <description>range - start included '(' syntax</description>
+        <resultNode name="range_004" type="decision">
+            <expected>
+                <value xsi:type="xsd:boolean">false</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="range_005">
+        <description>range - start included ']' syntax</description>
+        <resultNode name="range_005" type="decision">
+            <expected>
+                <value xsi:type="xsd:boolean">false</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="range_006">
+        <description>range - end included</description>
+        <resultNode name="range_006" type="decision">
+            <expected>
+                <value xsi:type="xsd:boolean">true</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="range_007">
+        <description>range - end included ')' syntax </description>
+        <resultNode name="range_007" type="decision">
+            <expected>
+                <value xsi:type="xsd:boolean">false</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="range_008">
+        <description>range - end included '[' syntax</description>
+        <resultNode name="range_008" type="decision">
+            <expected>
+                <value xsi:type="xsd:boolean">false</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="range_009">
+        <description>range - start: LT unary syntax</description>
+        <resultNode name="range_009" type="decision">
+            <expected>
+                <value xsi:nil="true"/>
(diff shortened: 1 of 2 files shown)


opatrascoiu:
Thanks @StrayAlien for the PR

We should probably add a test for `(10).start` as well.  It's a very edge case but grammar rule 7.a derives in the above as well :).

tarilabs:
or maybe we could take the pragmatical approach, and raise to the RTF if the spec where is says

> Ranges can be syntactically represented either:
> a. as a comparison operator **_and_** a single endpoint (grammar rule 7.a.)

if it weight on the case for `7.a` with the "and", or any possible case in that rule ;)

StrayAlien:
@opatrascoiu , @tarilabs - phew.  I never would have seen that case!  The 'and' in "as a comparison operator and a single endpoint" would have led me to think that syntax was not for a range but for a unary test ... but they sure have blurry edges now.   Matteo, are you saying we should maybe seek clarification before adding a test?

tarilabs:
@StrayAlien
> Matteo, are you saying we should maybe seek clarification before adding a test?

I am saying I do not concur to add a test for "`(10).start`", and that if you believe is meaningful, I respect that position but as in other case of non-agreement, someone needs to raise an issue to the RTF first and would need to move the test as commented-out.

The paragraph I quoted --like others in the spec-- it is my understanding it is meant to aid in attributing the semantic to the grammar rules, which would be of limited utility if just taken in isolation.

Also, the grammar per-se cannot almost never taken in isolation.
Especially when dealing non context-free grammar sounds a bit strange to me 🤷‍♂️ 

But let's do the exercise on paper.
Exercise for `(10)`

- Token for `(`
- grammar rule 1.b
- grammar rule 2.h, 4th case (` ...  | "(" , expression , ")" ...`)
- Token for `10`
- grammar rule 1.b
- grammar rule 2.h, 1st case (`literal | ... `)
- grammar rule 31, 1st case (`simple literal | ...`)
- grammar rule 32, 1st case (`numeric literal | ...`)
- grammar rule 35, matching `digits`
- grammar rule 37, matching for the expansion `digit, digit`

etc.

Results: `(10)` is a simple numeric literal within parens, as one would pragmatically expect.

So my pragmatical conclusion is that I would NOT expect `(10).start` to be a valid FEEL statement, for the reasons above.
I have no objection if the test is commented out with a reference to an RTF jira tho!

opatrascoiu:
@StrayAlien @tarilabs

The FEEL grammar is ambiguous according to the classic definition of ambiguity: for one word there are several parse trees. 
It has been like this since DMN 1.1. I raised a ticket a while back but never had time to write a proposal. https://issues.omg.org/browse/DMN14-36. I am planning to look into it again in the next 2-3 weeks and write a proposal. 

The FEEL grammar has been improved over time. However, there are still several remaining issues. For example:
- rule 1, is ambiguous because `boxed expressions` and `textual expressions` overlap
- rule 2.h is ambiguous `literal` and `simple positive unary test`  overlap because the operator is optional in rule 7.a

Unfortunately these design flaws are not reported by some parser generators (e.g. ANTLR) as errors / warnings and instead try to recover from errors by using other techniques (e.g. order of productions). 

I don't think that 123 or (123) is a unary test / range either :).  We should have a test for `(123).start` with the expected value **null**.

If we make the operator mandatory in rule 7.a the ambiguity goes away.

tarilabs:
I see @opatrascoiu , thank you for the feedback.

The reason I always do the exercise "on paper" is indeed not to be bound by tool specific behaviour.
That's the reason I did the exercise **_manually_** in https://github.com/dmn-tck/tck/pull/394#issuecomment-768135876
just by based on the content of the spec (ch. 10.3.1.2 ).

If I understand correctly, you are suggesting to augment the spec so not having to rely on ordering of the grammar rules to disambiguate when at some point ones need to eventually derive the parser (manually in this case, or automatically for impls.)
👍  thank you in advance

StrayAlien:
I have added some assertions that `(10).start` and so on will be a null/error result.

tarilabs:
To recap my understanding is unchanged since https://github.com/dmn-tck/tck/pull/394#issuecomment-768198511 , even with the new commit 7270aeb.

That is, we concur on the results of these tests, but an ambiguity of the grammar is worth raising to the RTF.
Correct?

opatrascoiu:
I raised an issue a while back https://issues.omg.org/browse/DMN14-36

I added a proposal to be discussed in the next meeting.

tarilabs:
We have 2 options for this PR, I believe:

- if we all agree even before the disambiguation of the Spec with  https://issues.omg.org/browse/DMN14-36 , I believe we can merge this PR
- otherwise, we will formally await DMNv1.4 and then merge this.

tarilabs:
RTF 2021-06-10 Greg will comment out the ones taking ther form `(x).start` etc, while the other will be merged.
The `(x).start` will be merged with DMNv1.4

StrayAlien:
@tarilabs - those tests are commented out now.

tarilabs:
Removing the https://github.com/dmn-tck/tck/labels/Waiting%20on%20RTF label since https://github.com/dmn-tck/tck/pull/394#event-4973823123 comment out those tests in reference.

tarilabs:
@opatrascoiu this could be a low-hanging fruit to be merged, thank you for having raised https://github.com/dmn-tck/tck/pull/394#issuecomment-797636474 and in the meantime Greg commented out impacted tests with 
0ac4ab7

this could look like easy to converge on and to merge, wdyt ?

opatrascoiu:
@tarilabs looks good to merge to me

tarilabs:
DMN TCK 2021-12-14 going to merge as discussed per scroll.

PR #395, opened 2021-01-06
DMN1-3 unicode

StrayAlien:
Some additional assertions on the new 6 digit unicode.  Maybe waaay less than required for such a topic, but something is better than nothing.  Hopefully I got it right .... 

comments welcome.

### DIFF
--- a/TestCases/compliance-level-3/0083-feel-unicode/0083-feel-unicode.dmn
+++ b/TestCases/compliance-level-3/0083-feel-unicode/0083-feel-unicode.dmn
@@ -9,6 +9,21 @@
         </literalExpression>
     </decision>
 
+    <decision name="decision_001_a" id="_decision_001_a">
+        <variable name="decision_001_a"/>
+        <literalExpression>
+            <text>string length("\U01F40E")</text>
+        </literalExpression>
+    </decision>
+
+    <decision name="decision_001_b" id="_decision_001_b">
+        <variable name="decision_001_b"/>
+        <literalExpression>
+            <!-- NOTE: DMN 1.3 spec says this should be a "ὀ" -->
+            <text>"\U01F40E" = "🐎"</text>
+        </literalExpression>
+    </decision>
+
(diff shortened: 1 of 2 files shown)


TestCases/compliance-level-3/0083-feel-unicode/0083-feel-unicode.dmn
```diff
@@ -9,6 +9,21 @@
...
+        <variable name="decision_001_a"/>
+        <literalExpression>
+            <text>string length("\U01F40E")</text>
+        </literalExpression>
+    </decision>
+
+    <decision name="decision_001_b" id="_decision_001_b">
+        <variable name="decision_001_b"/>
+        <literalExpression>
+            <!-- NOTE: DMN 1.3 spec says this should be a "ὀ" -->
```
tarilabs, line 22:
```suggestion
            <!-- NOTE: the published DMN 1.3 spec has a typo reporting "ὀ" instead of the emoji, which has been redacted with https://issues.omg.org/browse/DMN14-72 -->
```

tarilabs (commented):
Thank you for this PR !

Good with me, but please accept comment link to the rtf jira.

TestCases/compliance-level-3/0083-feel-unicode/0083-feel-unicode.dmn
StrayAlien: Thanks @tarilabs - appreciated.  I have pushed your requested change but updated the given issue URL.  That URL you've supplied there requires a login.  The https://issues.omg.org/issues/DMN14-72 is public and does not.

btw, nice to know I wasn't crazy with that wrong character in the spec.  I really did look at that one for some time thinking I must be wrong, but ended up figuring I'd just go with what I thought was correct.  I later saw the annotation in the DMN 1.3 PDF changelog saying "note: special character may be wrong".

opatrascoiu:
Looks good to me. Thanks @StrayAlien @tarilabs

PR #398, opened 2021-01-09
DMN1-3 temporal functions

StrayAlien:
The four new temporal functions:

* day of year
* day of week
* month of year
* week of year

Hopefully nothing too controversial in here.   Tests for each func for date, datetime, null, named-params, etc.  Learning here: who would have thought that sometimes the last days of last year are part of the first week of this year .. and vice-versa.

As always, comments welcome.  Lemme know.

### DIFF
--- a/TestCases/compliance-level-3/0095-day-of-year-function/0095-day-of-year-function-test-01.xml
+++ b/TestCases/compliance-level-3/0095-day-of-year-function/0095-day-of-year-function-test-01.xml
@@ -1,0 +1,190 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<!-- Contributed to DMN TCK by StrayAlien -->
+<testCases xmlns="http://www.omg.org/spec/DMN/20160719/testcase"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation=""
+           xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+    <modelName>0095-day-of-year-function.dmn</modelName>
+
+    <testCase id="date_001">
+        <description>date: first day of year</description>
+        <resultNode name="date_001" type="decision">
+            <expected>
+                <value xsi:type="xsd:decimal">1</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="date_002">
+        <description>date: last day of year</description>
+        <resultNode name="date_002" type="decision">
+            <expected>
+                <value xsi:type="xsd:decimal">365</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="date_003">
+        <description>date: leap year has 366 days</description>
+        <resultNode name="date_003" type="decision">
+            <expected>
+                <value xsi:type="xsd:decimal">366</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="date_004">
+        <description>date: spec example</description>
+        <resultNode name="date_004" type="decision">
+            <expected>
+                <value xsi:type="xsd:decimal">260</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="date_005">
+        <description>date: unknown type at compile time with valid date as runtime input</description>
+        <inputNode name="date_input_001">
+            <value xsi:type="xsd:date">1970-01-01</value>
+        </inputNode>
+        <resultNode name="date_005" type="decision">
+            <expected>
+                <value xsi:type="xsd:decimal">1</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="date_006">
+        <description>date: will accept date as named param</description>
+        <resultNode name="date_006" type="decision">
+            <expected>
+                <value xsi:type="xsd:decimal">260</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="datetime_001">
+        <description>datetime: first day of year</description>
+        <resultNode name="datetime_001" type="decision">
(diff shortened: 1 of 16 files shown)


tarilabs:
Thank you for these tests!

I've raised a PR on your fork, so once you merge-with-squash there, it should reflect here: https://github.com/StrayAlien/tck/pull/2
That is because according to FEEL grammar rules 25-28 names cannot start with a digit, i.e.: cannot start with `2021...`.
The linked PR fixes that.

StrayAlien:
Thanks @tarilabs - much appreciated.  That is merged over on the fork.  Seem I have missed the 'names cannot start with a digit' thing.  Doh.  Thanks again.

StrayAlien:
Actually @tarilabs - I may have been hasty there - though, still (sort of ....) to accept the PR.  But kind of feel it should be left as it was.

Why?  As far as I can tell from the spec there is no requirement for "named-element" names to conform to FEEL.  It is a bit grey, but FEEL is not mandated by the spec here.  In theory, I could use other languages instead of FEEL and the component names then may match whatever rules or naming conventions are needed there.  At least that is my take.

Unfortunately, FEEL does 'bleed through' though in place - so the multi-language support thing is not that well though out.  For example, a decision service returns a context - which, as far as I can tell is a FEEL type.  If the model definition language supports other data types, and maybe not 'context' then this become very grey indeed.  Depending on where you read in the spec, the list of data types supported by FEEL also changes.  Section "10.3.1.3" lists list, range, context, function but section 7.3.3 does not list those.

But re named elements - Table 4 (for NamedElement metamodel) just says 'name' is a string.

The reason why our compiler did not pick this up is exactly because of the above - it is unclear what conventions named element names should conform to.   We've left named element names as being a 'string' and not having to conform to FEEL.  It is is used inside a FEEL expression it has to conform.

I'd be inclined to leave the PR as it was with the names with leading digits - I think that is legal.  The variable names are not even used inside of FEEL literal expressions - but happy for thoughts. And to be corrected!

tarilabs:
I think we've already been through a very similar discussion couple years ago; Named elements must conform to FEEL name grammar rules.

Otherwise, you wouldn't be able to access it via FEEL qualified names, e.g.: cannot access `myThing.123star` as `123star` is not a valid name (for, say, a field). 
Another example, you cannot name an InputData as `123star` for analogous reasons.

In short, it's an identifier for FEEL and must conform to the rules.

StrayAlien:
Thanks @tarilabs - I think we agree that FEEL requires FEEL names.  But not that all DMN named elements must conform to FEEL.

It needs to be a valid FEEL name _IF_ (and only if) it is used in FEEL.  Otherwise, it can be just a string.  As you've said:

> "Otherwise, you wouldn't be able to access it via FEEL qualified names,"  It is a string, but to use it in FEEL it should be a valid FEEL name.  

DMN supports any expression language you care to implement - it does not have to be FEEL - it could be python or JavaScript or anything.  A literal expression may be a different language to FEEL and therefore may have different naming requirements.  I would say that to use a name in a FEEL expression then yes, it does have to conform to FEEL.  But if not used by language 'x' then it does not have to conform to language 'x's requirements. including FEEL.

Otherwise, how could you satisfy DMN's support for other languages when saying all named elements must conform to FEEL?  

I think it is important to realise that DMN is not FEEL.

tarilabs:
@StrayAlien Greg,

in your https://github.com/dmn-tck/tck/pull/398#issuecomment-812444225 I believe editorially you are quoting me on something I didn't actually say:

image

in your https://github.com/dmn-tck/tck/pull/398#issuecomment-812444225 you state:

> I think it is important to realise that DMN is not FEEL

I do 101% percent agree with you.
I have also perspectives on what/should/shouldn't happen if an implementation wants to support other languages, but this PR github entry is probably not the most appropriate place to discuss such a broad topic.

To keep this PR discussion on topic, here we are on `Conformance Level 3` tests of the TCK.
dtc/19-12-06, chapter `10.2.1.4 Boxed Context`, page 113 states:

> The names SHALL be legal FEEL names.

Only including commit 4ff1c32 the DMN model, and specifically the boxed contexts, do conforms.

### In conclusion
You raised a PR to test some _temporal functions_.
I spotted a non-conformance in the DMN model you submitted, which is resolved with commit 4ff1c32 included in this PR page.
I believe as it stands _now_, we can agree on the semantic of these tests for _temporal functions_.

If you still have doubts about _names_ having to conform to FEEL rules for being valid identifiers ("name..." rules in the grammar) to comply with conformance level 3, even after the evidence I provided from the spec itself, I believe you can concur that can be discussed separately from these tests on temporal functions.

Hope this helps. What do you think?

StrayAlien:
Thanks @tarilabs - apols for tardy reply.  It was easter long weekend here so I put down the tools for a few days.  I mean this nicely, but sometimes I think the spec it tattooed under your eyelids!  :-) For sure, well done for spotting that specific text for boxed context.  I have combed the spec for 'SHALL' and 'MUST' to make sure we get as much verification as we can into the compiler but kind of missed that.  Doh!

FEEL names vs DMN entities is, yes, another topic.  Let's just get the tests in with those changes of your and we're all sweet.

tarilabs:
I suggest we add this to https://github.com/dmn-tck/tck/milestone/3 on the premise this is mainly covering for existing examples in DMNv1.3 (plus coverage thanks to Greg)

opatrascoiu:
Several InputData do not have a  typeRef. It's not the only test file. I have raised an issue #430. Lets discuss in the call.

tarilabs:
thank you @StrayAlien for updating the test by merging the latest PR.
As per 2021-05-13 (https://github.com/dmn-tck/tck/discussions/437#discussion-3366038) proceeding to merge.

PR #399, opened 2021-01-11
DMN1-3 context functions

StrayAlien:
the `entries()` and `get value()` functions were kind of only half in the 1.2 spec.  We had tests for them but not for named params - they were commented out because (if I recall) those names were not yet official, or something. Those tests are now uncommented here. Plus, I've added two tests for the `get value` function that assert it will accept a non-literal string key name - all existing tests were providing a literal string 'key' param and for those that implement type inference here based on the literal key name and its corresponding type in the context - it could be a gotcha. 

Apols for the inexactly named branch.

### DIFF
--- a/TestCases/compliance-level-3/0080-feel-getvalue-function/0080-feel-getvalue-function-test-01.xml
+++ b/TestCases/compliance-level-3/0080-feel-getvalue-function/0080-feel-getvalue-function-test-01.xml
@@ -61,7 +61,6 @@
         </resultNode>
     </testCase>
 
-<!--
     <testCase id="decision_007">
         <description>named params</description>
         <resultNode name="decision_007" type="decision">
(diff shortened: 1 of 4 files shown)


tarilabs (approved):
Thank you!

This is uncommenting and reinstating named parameters tests for "10.3.4.10 Context function", Table 81, for which RTF formalized the parameter names during DMNv1.3.
Plus demonstrating wire of inputdata for one of the functions, as in `get value(..., inputdata)`.

To me is okay

opatrascoiu:
I have raised a related issue here https://github.com/dmn-tck/tck/issues/430. It's more general.

tarilabs:
Thank you for the link @opatrascoiu I might have missed that.
Do you want me to remove this from the https://github.com/dmn-tck/tck/milestone/3 and we shall discuss this at the meeting?

opatrascoiu:
@tarilabs Lets discuss in the call.

tarilabs:
Thank you Greg for updating the test by merging the PR

As agreed on the TCK meeting of 2021-05-14, proceeding to merge.

PR #401, opened 2021-01-19
0099-arithmetic-negation

StrayAlien:
Some specific tests re arithmetic negation.  Just asserting what types may be negated.  And negation of already negated stuff.

There is one thing I'd like to highlight about arithmetic negation and my feeling is it is an oversight and may need to go for clarification and that is this in the S-FEEL section 9.5:

> Arithmetic negation (grammar rule 2.d) is defined only when its operand is a number: in that case, its semantics is
according to the specification of op:numeric-unary-minus

The same 'number' restriction is not mentioned in the FEEL (proper) section.

But as of 1.3 .. both duration types are now supported by abs() .  They already are supported by multiplications, so `@"P1D" * -1` is `-P1D` - arithmetic negation is effectively the same thing.  Indeed, a runtime is likely to implement arithmetic negation using the same logic it uses for `<duration> * <number>` (and vice versa).

So (IMO) it doesn't make too much sense to permit `@"P1D" * -1`, but not permit `-@"P1D"`, or `-foo`, where `foo` is a duration variable.

To that end the tests here include assertions that numbers and durations can be the subject of arithmetic negation.

If this requires clarification for FEEL (as opposed S-FEEL) then I'm happy to comment out those assertions and leave the rest.

Comments welcome.

### DIFF
--- a/TestCases/compliance-level-3/0099-arithmetic-negation/0099-arithmetic-negation-test-01.xml
+++ b/TestCases/compliance-level-3/0099-arithmetic-negation/0099-arithmetic-negation-test-01.xml
@@ -1,0 +1,135 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<!-- Contributed to DMN TCK by StrayAlien -->
+<testCases xmlns="http://www.omg.org/spec/DMN/20160719/testcase"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation=""
+           xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+    <modelName>0099-arithmetic-negation.dmn</modelName>
+
+    <testCase id="decision_001">
+        <description>negate number</description>
+        <resultNode name="decision_001" type="decision">
+            <expected>
+                <value xsi:type="xsd:decimal">-10</value>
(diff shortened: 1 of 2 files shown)


tarilabs:
Thank you as always for submitting relevant tests.

I do not concur with tests decision_003, decision_003a, decision_004, decision_004a.

image

so FEEL syntax `-e` in the DMN spec has semantic only for negating numbers for the time being.
I agree it is not great, since you can take a ym duration and dt duration and multiply by `-1` achieving the same result, so this is worthy to be raised at the RTF.
Since the RTF at OMG is wrapping up, for the time being I noted https://github.com/dmn-tck/tck/issues/482 and I will raise the JIRA at OMG once a new RTF is created.

All the other tests looks correct to me.
Can you kindly comment out the mentioned tests, please? This way we can proceed to merge of the others. Thanks!

StrayAlien:
Those tests are now commented out.

tarilabs:
@opatrascoiu would you concur we can merge this one, please?

opatrascoiu:
@StrayAlien Thank you for writing the tests. @tarilabs  looks good to me.

tarilabs:
Thank you for the confirmation Octavian; per TCK meeting 2021-11-12, merging.

PR #402, opened 2021-01-27
Drools version refresh and Jan 2021 results

tarilabs:


tarilabs (commented):
A Vendor's results refresh, merging.

PR #406, opened 2021-01-29
0100-arithmetic

StrayAlien:
All, a PR for assertions for “Table 59: Specific semantics of multiplication and division” and "Table 57: Specific semantics of addition and subtraction" and "Table 60: Semantics of exponentiation".

Additionally, invalid combinations of types for the above operations get a null 'error' test as well.

To my mind this has been a pretty big gap in the tests.  The 'red' in my code coverage here has always bugged me.

There is a _lot_ of boilerplate to these tests so they are generated by `generateTests.js`.

There is a readme file explaining some more.  But, in short, `generateTests.js` contains some data objects that it uses to generate the tests and for every lhs/rhs type combination that does not have an explicit test, a null/error test is also automatically generated.  The data objects cover tests lhs/rhs scenarios for the type combinations in the spec.

The tests are not exhaustive and do not cover all imaginable scenarios - more like some sanity checks for each lhs/rhs combination to verify they work and do cover some specific things mentioned in the spec - like dates being an implicate date time at UTC 00:00:00 for date time subtraction.  

There are some things to note, and @tarilabs, I think it possible the RH runtime may have issues with some of these.  Some combinations that the spec says are invalid appear to be supported in RH.  I'm not a RH runtime expert but, for example `<number> / <days and time duration>` and `<number> / <years and months duration>` seem to be non-error conditions in the RH code.  Other examples of same are `<days and time duration> * <days and time duration>` and `<years and months duration> * <years and months duration>`.

Additionally, when durations are multiplied or divided by numbers, only the integer part of the number is used - no decimals.  I'm not an RH expert, but it kind of looks like that.  The code is in https://github.com/kiegroup/drools/blob/master/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/lang/ast/InfixOpNode.java.  There are some other scenarios as well.  My apologies if I am incorrect - would not be the first time for sure.

And apologies for the enormous PR!  Hopefully the auto-generated test names are helpful to track down issues if you have any.

As always, comments and feedback welcome.

Greg.


opatrascoiu:
Wow - lots for tests. Thank you!

I had a look and so far they seem to be correct apart from:

```
    <decision name="subtract_lhs_date_minus_rhs_dtDuration_002" id="_subtract_lhs_date_minus_rhs_dtDuration_002">
        <description>subtract_lhs_date_minus_rhs_dtDuration_002</description>
        <allowedAnswers>2021-01-02 (date)</allowedAnswers>
        <variable name="subtract_lhs_date_minus_rhs_dtDuration_002"/>
        <literalExpression>
            <text>@"2021-01-02" - @"PT1H"</text>
        </literalExpression>
    </decision>
```

Based on Table 57, the expected result is  a `date` and should be calculated as 

1. value(@"2021-01-02") = 1609545600 - seconds from EPOCH
2. value(@"PT1H") = 3600 - seconds
3. value(@"2021-01-02") - value(@"PT1H") = 1609542000

The inverse function should return 2021-01-01 - previous day. It is also consistent with the fact @"2021-01-02" is equivalent with UTC midnight, hence when removing one hour it gives previous day.

StrayAlien:
Hi @opatrascoiu.  Thanks for comments, and thanks for timely response.   I believe you are correct.  But ... this one threw me a bit.  What you say above seems logical.  Except, that particular operation is not supported by Java LocalDate, and the example for the `https://www.w3.org/2003/01/DIFF-xquery-operators.html#func-subtract-dayTimeDuration-from-date` operation given here: https://docs.oracle.com/cd/E13214_01/wli/docs92/xref/xqdtopref.html#wp1259416 shows a period of 1 minute being _ignored_ as per this test.  The actual docs for the operation here: https://www.w3.org/TR/xmlschema-2/#adding-durations-to-dateTimes maybe _look_ like it is supposed to be as you say .... (?).

But, testing the XML operation (as the DMN specs reference) shows you are correct.  

Online xpath testers with this query: `xsd:date("2021-01-02") - xsd:dayTimeDuration("PT1H")` give `2021-01-01`.

(using this simple doc):

```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<foo
        xmlns:xsd="http://www.w3.org/2001/XMLSchema">
</foo>
```

Thanks for being on the ball.   I'll amend that and maybe look a little deeper into the xpath function specs.

StrayAlien:
I've updated that test and also added a few more around that same topic.

StrayAlien:
@opatrascoiu @tarilabs - I think we need to do a similar 'generate the tests' treatment as here but for 'comparison' tests.  We have very little coverage on that with the exception of equality.  For example,  `<=`  appears once in the TCK and only then in a comment. Some other operators get a little coverage in list and 'in' tests, but only incidentally. As the comparison operators entirely underpin the new DMN 1.3 range functions, I am inclined to give them the royal treatment first.  And maybe replicate all the 'equality' tests as part of a new 'comparison' suite.   yes, it is another big gap in coverage that has been staring at me ....

If you are happy with the way these tests are generated, I'll do similar for comparison operators.

opatrascoiu:
I totally agree with having a suite for relational tests :)

StrayAlien:
Thanks @opatrascoiu .  its a gap for sure, and whenever we poke a corner with more tests we resolve differences - and that makes the world a better place. :-)  I may bolster this 'generateTests' to create tests with swapped operands for commutative combinations. With comparisons, sort of the same thing needed, eg: each '=' test should get a corresponding '!=' test asserting the opposite boolean result.  Stuff like that.  '<' vs '>' etc.  It might mean a lot of 'obvious' tests, but I'd rather have more than less.

StrayAlien:
I forgot to ask @opatrascoiu - any comments on the way the tests are generated.  It is clear enough? .. 
 or muddy as hell?

StrayAlien:
some more changes here.  New tests for diffs between datetime with zones, the famous 24:00:00 time, plus negative years.  Also changed the generation to auto-create tests with swapped operands for commutative tests where appropriate.

opatrascoiu:
@StrayAlien I don't quite understand why the expected value for `exponent_lhs_number_exp_rhs_number_005` is null

`Table 60: Semantics of exponentiation` does limit the range of values but does nor restrict the exponent to integer. Am I missing something ?

StrayAlien:
Hi @opatrascoiu - Exactly the test I was hoping for comment on.  Thank you!  :-)  My initial feeling was the spec was changed from 'integer' to 'number' because 'integer' is not a supported type in the spec, but 'number' is - so, wasn't 100% sure.  However, in the last hour I've seen the issue raised by Denis Gagne (https://issues.omg.org/issues/DMN13-111) which says it should be any number.  I thought the _intention_ of the change was still to be an integral number exponent but seems that is not the case.  I'm no maths genius, but using an arbitrary non-integral numbers as an exponent is not child's play.  Eg, Java BigDecimal does not support it, though Javascript does.

So, yes, maybe that test is incorrect.  :-(

StrayAlien:
that test is amended now @opatrascoiu - test assertion is to 11 decimal places. Thanks for the feedback and reality check.

opatrascoiu:
@StrayAlien Looks god to me. Thank you!

StrayAlien:
Thanks for checking it out @opatrascoiu. 👍

opatrascoiu:
@StrayAlien @tarilabs I looked again at the latest version. It looks good to me.

opatrascoiu:
@tarilabs This one looks good to merge. Could you double check please?

tarilabs:
During these days I've focused the attention to other PRs and issues; this one and #390 are next on the list, is quite big and I'd like the opportunity to revise it

StrayAlien:
@baldimir - as far as I can tell we're okay here.

opatrascoiu:
Looks good to me.

baldimir:
Thank you. I tested it and as @StrayAlien mentioned, there are some failures with Drools DMN engine, so I need to take a look. I am timeboxing this investigation for the next week, so at latest at the end of next week, I will merge it if there are no new objections. Is that fine for you @StrayAlien and others @dmn-tck/contributors please?

baldimir:
I started reviewing today, I will let you know when I finish. I expect to finish tomorrow.

baldimir:
@StrayAlien thanks for the PR and everyone for the discussion, I merged this.

StrayAlien:
Thank you!!!

PR #409, opened 2021-02-13
bkm typeref was missed in PR #385

StrayAlien:
@opatrascoiu. @tarilabs - Missed one I think.

### DIFF
--- a/TestCases/compliance-level-3/0082-feel-coercion/0082-feel-coercion.dmn
+++ b/TestCases/compliance-level-3/0082-feel-coercion/0082-feel-coercion.dmn
@@ -125,10 +125,10 @@
         </literalExpression>
     </decision>
 
-    <!-- has an explicit context return type but return boolean -->
+    <!-- has an explicit context return type but returns boolean -->
     <businessKnowledgeModel name="bkm_002" id="_bkm_002">
-        <variable typeRef="tNameAndAge" name="bkm_002"/>
-        <encapsulatedLogic>
+        <variable name="bkm_002"/>
+        <encapsulatedLogic typeRef="tNameAndAge" >
             <formalParameter name="nameAndAge" typeRef="tNameAndAge"/>
             <literalExpression>
                 <text>nameAndAge != null</text>
--- a/TestCases/compliance-level-3/0082-feel-coercion/0082-feel-coercion.dmn
+++ b/TestCases/compliance-level-3/0082-feel-coercion/0082-feel-coercion.dmn
@@ -11,9 +11,8 @@
     </itemDefinition>
 
     <itemDefinition name="tDS_001">
-        <itemComponent name="decision_ds_001">
-            <typeRef>string</typeRef>
-        </itemComponent>
+        <functionItem outputTypeRef="string">
+        </functionItem>
     </itemDefinition>
 
     <itemDefinition name="tNameAndAge">
@@ -125,7 +124,7 @@
         </literalExpression>
     </decision>
 
-    <!-- has an explicit context return type but returns boolean -->
+    <!-- has an explicit context return type but return boolean -->
     <businessKnowledgeModel name="bkm_002" id="_bkm_002">
         <variable name="bkm_002"/>
         <encapsulatedLogic typeRef="tNameAndAge" >
@@ -409,7 +408,7 @@
 
     <decisionService name="decisionService_001" id="_decisionService_001">
         <variable name="decisionService_001" typeRef="tDS_001"/>
-        <!-- decision returns number, type expects string -->
+        <!-- decision returns number, functionType typeRef has return type of string -->
         <outputDecision href="#_decision_ds_001"/>
     </decisionService>
 
--- a/TestCases/compliance-level-3/1130-feel-interval/1130-feel-interval.dmn
+++ b/TestCases/compliance-level-3/1130-feel-interval/1130-feel-interval.dmn
@@ -459,7 +459,7 @@
         </tc:testCases>
     </semantic:extensionElements>
     <semantic:decisionService id="_4dacb45e-9b17-48e0-9518-8cd1b47d445b_DS" name="Whole Model Decision Service" triso:dynamicDecisionService="true" triso:wholeModelService="true">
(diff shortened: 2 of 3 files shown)


tarilabs (approved):
Thanks!

opatrascoiu:
Looks good to me too. Thank you!

StrayAlien:
@tarilabs / @opatrascoiu - apols for the commit after approval, but here are two more missed instances of typeRef on a BKM or DS.  The recent range tests had some DS typeRefs of 'Any'.  This may be a moot point.  We've figured that if an invocable has a typeRef it must be a function type.  But the 'Any' thing is interesting.  A function type is compatible with Any, but Any is not a function type.  I figure this can be discussed elsewhere, but removing the 'Any' from the DSs in the range checks as they are not important to the test and we can focus on tests for function types elsewhere.

StrayAlien:
@tarilabs / @opatrascoiu - apols for the commit after approval, but here are two more missed instances of typeRef on a BKM or DS. The recent range tests had some DS typeRefs of 'Any'. This may be a moot point. We've figured that if an invocable has a typeRef it must be a function type. But the 'Any' thing is interesting. A function type is compatible with Any, but Any is not a function type. I figure this can be discussed elsewhere, but removing the 'Any' from the DSs in the range checks as they are not important to the test and we can focus on tests for function types elsewhere.

There is also another instance in the coercion tests - also for a DS.  But, rather than remove the typeRef, as the point of the test is to verify a type mismatch I have changed it to be a function type item definition.

tarilabs (commented):
This PR was evaluated for *BKM* typeRef.
So I do not concur with the following changes which have been added after the previous agreement:
 - db6749fd090f64207d7ef01a109a38a14a54e755

since I am not sure on which premises the variable for BKM representing the function call, can translate to the Decision Service variable?

opatrascoiu:
I did not check all the changes yet. 

However, both BKMS and DS are Invocables and share common properties described in Table 15: Invocable attributes and model associations. The semantics of variable is described here hence the semantics is the same for both.

StrayAlien:
@opatrascoiu / @tarilabs -  I think we have got this PR wrong. 

I think the typeRef for a BKM _AND_ BKM its encapsulatedLogic, Decision Services _AND_ FunctionDefinitions should be all function types.  Let me explain.

The DMN schema says a `FunctionDefinition` expression is a `tFunctionDefinition`.  BKM encapsulatedLogic is also a `tFunctionDefinition`.  A `tFunctionDefinition` is an expression and can have a typeRef, but it also have formal params and a return type.  So, in both bases the type is _actually_ a function type.

Read this text again (Table 15: Invocable attributes and model associations):

> This attribute defines a variable that is _bound to the function defined by the FunctionDefinition_, allowing decision logic to invoke the function by name.

"that is bound to the function defined by the FunctionDefinition".  So, the BKM encapsulatedLogic is a _function_, and thus should be typed as such.  

We've just made changes to put _non-function_ types on the typeRefs on BKMs - effectively expressing the return type, not the function type itself.  I think this is incorrect.  I think we should have put those typeRefs on the _expression_ within the encapsulatedLogic, because the type for the encapsulatedLogic really is a function type just like a FunctionDefinition.

The type of a (box) FunctionDefinition really must be a function type (I'm pretty sure we've all dine it like that) - it is invoked and the invoker must know about params etc.  But, we're typing the encapsulatedLogic of a BKM differently, even though it is the same thing.

Hopefully that makes sense.

If we agree, I think we should remedy this _ASAP_ as the tests are now incorrect.  And it is stuffing up some other tests I am writing on domain correctness with allowed values.

opatrascoiu:
@StrayAlien I agree with you.

I also found something in the spec:

> 7.3.1 Expression metamodel (page 77)
> 
> An instance of Expression references an optional typeRef, which points to either a base type in the default typeLanguage,
> a custom type specified by an ItemDefinition, or an imported type.  The referenced type specifies the Expression's range of possible values.
> 

We should enhance section 7.3.1 with more details about Invocables (functions) - they were added later in DMN hence the gap.

StrayAlien:
Thanks @opatrascoiu - I was hoping you'd agree. 👍 .  Re the spec - yes, the fact that an Invokable and a FunctionDefinition can only really have function types as typeRef should be in the spec.

tarilabs:
> the fact that an Invokable and a FunctionDefinition can only really have function types as typeRef should be in the spec

not sure I share this statement.

I can also have a Decision node, with a LiteralExpression as the decision logic, which has `function(a) a`.
That is a Decision which "_decides_" a function.
I never seen in practical applications, but it's in my understanding, valid.
So far I've left out `typeRef` to this kind of Decision, or alternatively I've marked `typeRef="Any"`. But this is a valid case where I understand it could be as well wired up to a FunctionItem as ItemDefinition, do you agree?

StrayAlien:
Hi @tarilabs - yes, it is valid to have a decision return a lambda or a FunctionDefinition.  Well, we have tests for it in the TCK anyways!  And yes, not sure who'd want to, but, in the hands of developers, anything can happen ....

Above, I wasn't meaning that the typeRef for decision housing a lambda or FunctionDefinition had to be a function type, but, if you associate a typeRef to a invokable/FunctionDefinition element, if should be a function type.  A decision housing a FunctionDefinition (even a typed one) could still be 'Any' and all would be fine.  I was more meaning that invokables and FunctionDefinition can really only ever have a function type, because they are functions and so, the spec should represent that.

Though, the 'Any' thing has been buggin' me a bit, especially when it is on a BKM / DS typeRef.  Like, the tests Simon submitted recently for ranges had Decision Services with a typeRef="Any".

I think maybe a better way to do it is the actual (runtime/inferred) type of a invokable/FunctionDefinition must _conform_ to the type specific by a typeRef (if any).  That could only mean the typeRef could be a function type, or Any.  Which I think is okay.

And by conform, I mean follow the covariant/contravariant rules for function type conformity.
  
If any of that made sense ....

Happy for your thoughts.

StrayAlien:
closing off.  To make life less confusing after the merge of @opatrascoiu PR with similar changes, I'll open another PR with the specific changes.

Issue #386, opened 2020-12-23
Debatable / incorrect test case in CL3/0085-decision-services


opatrascoiu:
The test case below seems to be incorrect. 

```
    <testCase id="007">
        <description>passing a single input decision param with incorrect type</description>
        <resultNode name="decision_007_1" type="decision" errorResult="true">
            <expected>
                <value xsi:type="xsd:boolean">true</value> <!-- DMN13-163 Decision Service output value in the case of single output Decision -->
            </expected>
        </resultNode>
    </testCase>
```
I think the expected result should be **null**. Here is the rationale:
- the execution semantics of a DS is identical to the execution semantics of a FEEL function (see 3.4 and 10.3.2.13.2)
- according to 10.3.2.13.2 if the arguments do not conform to the argument list the result of the invocation is **null**.

In our case the input has a _number_ type that does not conform to _string_ (see 10.3.2.9.2 Type Conformance) and there is not conversion (see 10.3.2.9.4 Type conversions).

StrayAlien:
Hi @opatrascoiu, good observation and TBH, I'm not quite sure where we landed on this.  But, here is a take:

Section 10.3.2.13.1 Built-in Functions says:

> Built-in functions are invoked using the same syntax as other functions (grammar rule 40). The actual parameters must
conform to the parameter domains in at least one signature before or after applying implicit conversions, or the result of
the invocation is null. 

Note the "after" applying conversions.

Then Section 10.3.2.9.4 Type conversions on when implicit conversions occur:

> Binding contexts in which the value of an expression is bound to a variable with associated type information
(e.g. binding actual parameters to formal parameters in an invocation, or binding the result of a decision’s logic
to the decsion’s output variable). The expression is subject to conforms to conversion.

(gotta love the typo in the above - 'decsion').  So the implicit conversion is to null, which conforms to string, hence the invocation proceeds.

Not 100% sure that is the case here, but the tests were thought through pretty thoroughly.  Happy for comments.

opatrascoiu:
@StrayAlien You are right, I completely forgot about the _conforms To_ implicit conversion.

StrayAlien:
Hi @opatrascoiu  - I'd be interested to hear what others think here.  My comments above were more of a (erm) comment thn a statement.  Those implicit conversions do take place and the func is invoked, , but that seems to contradict the "if the arguments do not conform to the argument list the result of the invocation is null" thing.

[...]

Issue #352, opened 2020-01-03
Wrongly named parameters are an error - or ignore them? 


StrayAlien:
Hi all,

We have a number of tests asserting that passing a named parameter to a function that has no matching formal parameter is an error.  I'm not meaning that missing named parameters get a null value, I mean like a typo in a param name - like `sqrt(number: 123)` (correct) vs `sqrt(numbr: 123)` (incorrect) - such that when the function/invocable formal params are scanned for the given param name none is found.

It's kind of the same as for the tests where we assert an error when we pass too many (positional) params.

The trouble is, our 'assert badly named params' tests all are giving false positives at the moment - so they are not actually asserting what they intend on asserting (apologies for that - mea culpa).  The one exception was a test I submitted in suite 0085 (discussed here: https://github.com/dmn-tck/tck/issues/349) which was not a false positive.  That was changed such that the result means badly named params are ignored and do not result in a null function result.

SIDEBAR: the reason for the false positives is that, as @tarilabs kindly pointed out, `sqrt(numbr: 123)` will result in null anyways as a _missing_ `number` param will default to a null value .. and thus the function result will be null.   

IMO wrongly named params should be treated like too many positional params, but it seems maybe other impls do not do same. I'm not religious about it, but think that erring on the side of correctness benefits all.

What do others think?

Greg.

tarilabs:
Hi Greg,
as we discussed at length here #349 the DMN spec is clear on the execution semantic when invoking a function with named parameters, and when named parameters are missing, and as proved here: https://github.com/dmn-tck/tck/issues/349#issuecomment-565977874 and here: https://github.com/dmn-tck/tck/issues/349#issuecomment-566063458 with hand-written execution steps I did targeting your specific reference.

Now you opened a new GitHub issue and I'm happy to help further if you feel it has any value, but rather than a generic textual description, would you be kind to provide a specific, bounded example of what you mean?

I'm just trying to be helpful and to avoid going in circles (https://github.com/dmn-tck/tck/blame/master/README.md#L44), if that helps.

I'll try to give an example first.
You have hinted at calling:
`sqrt(numbr: 123)`
as explained here: https://github.com/dmn-tck/tck/issues/349#issuecomment-566063458
```
sqrt(numbr: 123) // original
sqrt(number: null, numbr: 123) // applies DMN spec 10.3.2.13.5 
null // result, cannot square root a null.
```
so would result in a `null` value (and an error message as you cannot square root the null).

What are you expecting instead?

[...]

PR #412, opened 2021-02-20
DMN1-3 feel is function.

StrayAlien:
All, an initial suite of tests for the new DMN 1.3 is() function.  

This guy is not very well specified at all and it is not very clear what boundaries are on this guy - type the types it can operate on and so forth.  I have raised a issue with DMN seeking clarification.

But, I have some some assumptions that it operates on the types in section in 10.3.2.2 Equality, Identity and Equivalence.  To that end I have no tests covering other types until we know more.

Also, one thing in these tests is the notions that for types with zone/offset info, for the purposes of semantic equivalence we are interested in the _offset_.  The semantic fields given in 10.3.2.2 refer to an optional _offset_. So, two values being equal in all respects except for zone/offset will be semantically equal if they 'normalise' to the same offset value.  

The example in the spec `is(time("23:00:50z"),time("23:00:50+00:00”)) is false` is incorrect I believe.  @tarilabs - I saw an issue raised by you also on that test.  These tests assume that value should be true as `Z` and `+00:00` are the same offset.

Note also the comment i made on the equality tests regarding DMN 'zoned times' and how the time of year nay affect their offset value.  Repeated here:

> As a note, as DMN supports a time with at zone like 10:15:00@Europe/Paris - which is not a hard offset like +01:00 it is possible that some implementations may use the equivalent of now() to get a zoned date time to get an offset from. This is a little dangerous because the meaning of @Europe/Paris and its offset changes during the year. So a test like @"0:15:00+01:00" = @"10:15:00@Europe/Paris" will be correct only for some part of the year. At other times the offset will be +02:00. So, my suggestion is, do not use now() with DMN zoned times, use a standard epoch date like 1970-01-01T00:00:00 to get an offset, then it'll be constant. The spec does not say that, but, maybe it should.

Some of the tests here are purposely the inverse of the results they give in the equality test suite.

I have add some token tests for number/boolean/string more as a means of verifying the impl support those types passing through the is() function.  

All feedback appreciated.

Greg.

### DIFF
--- a/TestCases/compliance-level-3/0103-feel-is-function/0103-feel-is-function-test-01.xml
+++ b/TestCases/compliance-level-3/0103-feel-is-function/0103-feel-is-function-test-01.xml
@@ -1,0 +1,460 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<!-- Contributed to DMN TCK by StrayAlien -->
+<testCases xmlns="http://www.omg.org/spec/DMN/20160719/testcase"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation=""
+           xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+    <modelName>0103-feel-is-function.dmn</modelName>
+
+    <testCase id="number_001">
+        <description>number: is equal</description>
+        <resultNode name="number_001" type="decision">
+            <expected>
+                <value xsi:type="xsd:boolean">true</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="number_002">
+        <description>number: is not equal</description>
+        <resultNode name="number_002" type="decision">
+            <expected>
+                <value xsi:type="xsd:boolean">false</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="boolean_001">
+        <description>boolean: is equal</description>
+        <resultNode name="boolean_001" type="decision">
+            <expected>
+                <value xsi:type="xsd:boolean">true</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="boolean_002">
+        <description>boolean: is not equal</description>
+        <resultNode name="boolean_002" type="decision">
+            <expected>
+                <value xsi:type="xsd:boolean">false</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="string_001">
+        <description>string: is equal</description>
+        <resultNode name="string_001" type="decision">
+            <expected>
+                <value xsi:type="xsd:boolean">true</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="string_002">
+        <description>string: is not equal</description>
+        <resultNode name="string_002" type="decision">
+            <expected>
+                <value xsi:type="xsd:boolean">false</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="date_001">
+        <description>date: is equal</description>
+        <resultNode name="date_001" type="decision">
+            <expected>
+                <value xsi:type="xsd:boolean">true</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="date_002">
+        <description>date: is not equal when year differs</description>
+        <resultNode name="date_002" type="decision">
+            <expected>
+                <value xsi:type="xsd:boolean">false</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="date_003">
+        <description>date: is not equal when months differs</description>
+        <resultNode name="date_003" type="decision">
+            <expected>
+                <value xsi:type="xsd:boolean">false</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="date_004">
+        <description>date: is not equal when days differs</description>
+        <resultNode name="date_004" type="decision">
+            <expected>
+                <value xsi:type="xsd:boolean">false</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="date_005">
+        <description>date: is not equal to UTC zero time of same date</description>
+        <resultNode name="date_005" type="decision">
+            <expected>
+                <value xsi:type="xsd:boolean">false</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="time_001">
+        <description>time: is equal</description>
+        <resultNode name="time_001" type="decision">
+            <expected>
+                <value xsi:type="xsd:boolean">true</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="time_002">
+        <description>time: is equal with same offset</description>
+        <resultNode name="time_002" type="decision">
+            <expected>
+                <value xsi:type="xsd:boolean">true</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="time_003">
+        <description>time: is equal with same timezone</description>
+        <resultNode name="time_003" type="decision">
+            <expected>
+                <value xsi:type="xsd:boolean">true</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="time_004">
+        <description>time: is equal - UTC zone and UTC offset</description>
+        <resultNode name="time_004" type="decision">
+            <expected>
+                <value xsi:type="xsd:boolean">true</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="time_005">
+        <description>time: is not equal - no zone time vs UTC zone time</description>
+        <resultNode name="time_005" type="decision">
+            <expected>
+                <value xsi:type="xsd:boolean">false</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="time_006">
+        <description>time: is equal - @ zone time with same offset zone time</description>
+        <resultNode name="time_006" type="decision">
+            <expected>
+                <value xsi:type="xsd:boolean">true</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="time_007">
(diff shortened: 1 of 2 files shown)


opatrascoiu:
@StrayAlien 
I totally agree with with about adding ZoneID to local times. Using the Epoch approach is a sane, possible approach. However, we should agree on the approach here and in RTF before we merge the PR. Lets discuss in the next meeting.

FYI there is going to be a discussion about date / time aspects in the near future in RTF (e.g. adding partial date / time points). I will propose removing any confusing / redundant references to XML date / time types and use the ISO standard instead.

In my view all the above issues come from the fact that DMN inherited the date / time type system from XML. Things have  changed since then, the PLs evolved in this space (see Java 8).  For example, my preference would be to have only two time types : `local time` and `offset time`. That would make the semantics of relational operators and `is` crystal clear.

StrayAlien:
HI @opatrascoiu .  Thanks. The 'zoned' time thing sure is a little weird isn't it.  Good you think the same on the epoch/zone thing - it not me being weird! :-) Cool - if the zoned time type remains in future DMN specs then maybe best to get into the spec a specific epoch date so we all do the same thing.

Re removing the XML references, I agree it quite likely that making reference to the 'semantic' space of ISO rather than XML would be good, plus dropping the references  to the 'op:' stuff in XPATH.  It seems now that the DMN spec defines equality and arithmetic in its own way which no longer matched the XPATH docs anyways - and that is a very confusing thing.

StrayAlien:
Oh, @opatrascoiu - I was meaning to ask, what is your understanding of the types that is() should operate on?  This is pretty poorly specified atm.

tarilabs:
> The example in the spec `is(time("23:00:50z"),time("23:00:50+00:00”)) is false` is incorrect I believe.  @tarilabs - I saw an issue raised by you also on that test.  These tests assume that value should be true as `Z` and `+00:00` are the same offset.

Yes, my understanding is that `Z` is a synonym for `+00:00`.
That example has been corrected with DMNv1.4 --> https://issues.omg.org/issues/DMN14-86

According to the same, please ref the example table, I do not concur with some of the tests in this PR.

## datetime_014

`is(@"2002-04-02T23:00:00@Australia/Melbourne", @"2002-04-02T23:00:00@Australia/Sydney")` they might be equal, but they are not the same semantic domain equivalent.

The decision logic is allowed, but the test result should be `false`.

## datetime_016

`is(@"2021-04-02T23:00:00@Australia/Melbourne", @"2021-04-02T23:00:00+11:00")` they might be equal, but they are not the same semantic domain equivalent. 

The decision logic is allowed, but the test result should be `false`.

## datetime_017

`is(@"2021-10-02T23:00:00@Australia/Melbourne", @"2021-10-02T23:00:00+10:00")` they might be equal, but they are not the same in the semantic domain.

The decision logic is allowed, but the test result should be `false`.

## named_params_002

Spec chapter `10.3.2.13.5 Positional and named parameters` says:

> If positional, all parameters SHALL be supplied. If named, unsupplied parameters are bound to null.

So the test:
```
is(value1: @"2021-02-13")
// is equivalent to
is(value1: @"2021-02-13", value2: null)
```

The result should be `false`.

I also believe the intention of this test was to be written as `is(value2: @"2021-02-13")` and this would sound more in-line with your comments (thank you for providing those! I really appreciate them)

https://github.com/dmn-tck/tck/pull/412/files#diff-4ecb745cfc2b876338a3a378bab05cfce4fd35b82a93df6531b15dbe2d2b96d9R491-R495

To recap:

1. this test should have for decision logic `is(value2: @"2021-02-13")`
2. the test result should be `false`

## named_params_003

Analogous to above.
The decision logic is allowed, but the test result should be `false`.

## named_params_004

There is no function signature with a named parameter called `value3`.
The decision logic is allowed, but the test result should be `null` (and an error).

## time_006

`is(@"23:00:50@Australia/Melbourne", @"23:00:50+10:00")` they might be equal, but they are not the same element on the semantic domain.

The decision logic is allowed, but the test result should be `false`.

## time_007

`is(@"23:00:50@Etc/GMT", @"23:00:50Z")` they might be equal, but they are not the same element on the semantic domain.

The decision logic is allowed, but the test result should be `false`.

## time_012

`is(@"23:00:50@Australia/Melbourne", @"23:00:50@Australia/Sydney")` they might be equal, but they are not the same element on the semantic domain.

The decision logic is allowed, but the test result should be `false`.

# Conclusions

Thank you, as always, for providing these throughout tests.

In general it seems to me the `is()` function was introduced as an attempt in early DMN spec version to address some problem with date/time/datetime but as also Octavian mention in comment https://github.com/dmn-tck/tck/pull/412#issuecomment-782694100 there are even broader issues about granularities (date with only year+month but no day) which the RTF might consider in the future.

Nevertheless, if we concur on the 6 analysis points above and amend this test PR accordingly, we could merge this.
I'm also happy to discuss in more details the 6 analysis points reported above either in meeting or offline here!

opatrascoiu:
I agree with @tarilabs on  :
      time_006
      time_007
      time_012
      datetime_014
      datetime_016
      datetime_016
      named_params_002
      named_params_003
      named_params_004
AFAIK the purpose of is() is to check the textual equivalence. 

All the components of date and time values (e.g. year and hour components) must be the same for is() to return true. I would add to the list above to discuss:
      dt_duration_002
      ym_duration_002
      zero_duration_001
I think the normalization mentioned in spec for the above durations refers to equals and not for is() to be consistent for all the temporal types. Lets discuss :)

StrayAlien:
hey guys, thanks for the comments.  I really appreciate it.  

First I have to say, the `is()` part of the spec is truly pretty bad.  Like, it seems to clearly be for all types, but the function defn is under "Date and time Functions".  I really don't think anyone could read the spec and have an idea of what is expected here.  

Starting at the end.  @opatrascoiu.  For durations, I think the spec is pretty clear for both - at least as I read it.  The semantic domain is normalized.

> A days and time duration in the semantic domain is a sequence of numbers for the days, hours, minutes, and seconds of duration, normalized such that the sum of these numbers is minimized. For example, FEEL(duration("P0DT25H")) = P1DT1H.

> A years and months duration in the semantic domain is a pair of numbers for the years and months of duration, normalized such that the sum of these numbers is minimized. For example, FEEL(duration("P0Y13M")) = P1Y1M.

////

@tarilabs - re the time zones and semantic equivalence - I do think the above tests with zones are correct.  I read the semantic value for a time zone is its UTC form, not the string value.  

date time:

> A date and time in the semantic domain is a sequence of _**numbers**_ for the year, month, day, hour, minute, second, and optional time offset from Universal Coordinated Time (UTC).

And for time:

> It can be represented by a sequence of _**numbers**_ for the hour, minute, second, and an optional time offset from Universal Coordinated Time (UTC).

Both of the above are not discussing the value, they are referring to the semantic domain.

Thus, the semantic domain for time and date time is all _numbers_.  The semantic domain has a time "offset", not a zone nor its string representation.  In effect, the semantic domain for a time zone string is its offset. 

Happy to discuss, but to me that seems pretty clear.  One has to chew on it for a while, but I had to read it slowly about 100 times!  The semantic equivalence of time and date times uses numeric offsets, not string zones.

//////

The named params thing is something we always see to run into trouble about!  That spec section:

> If positional, all parameters SHALL be supplied. If named, unsupplied parameters are bound to null.

That is specifying what to do for a _unsupplied_ named parameter, not an _expected_ or _unknown_ parameter.  That is, it is about _incomplete_ named params, not _incorrect_ named params. There is a difference. 

In the case of the test `named_params_002`, the named param `value1` is not in the formal parameters of the function so therefore according to :

> If f is not a function, or if the arguments do not conform to the argument list, the result of the invocation is null.

The arguments do not conform to the arguments list - there is no formal param for `value1`. Hence the result is null.  I agree, that if the function took two params and we only supplied one, then the other is null - the spec is clear, and that makes sense.  But, it also makes sense that to supply a named param that is _not_ in the formal params means the arguments are _incorrect_, not _incomplete_.  This is what is being tested.

Thanks again.

StrayAlien:
@tarilabs - yep.  Re named_params_002 and named_params_003.  Correct - my apologies, I had the params for is() as not nullable.  I've updated tests to reflect that.

Re named_params_004 - although, i agree, my personal feeling is this should be null/error as the argument `value3` is not a defined formal parameter, but, as it stands at the moment, other tests are ignoring badly named params.  That was the point of this issue discussion: https://github.com/dmn-tck/tck/issues/352, and also to some degree here: https://github.com/dmn-tck/tck/issues/386.  It is something I have been trying to stress for a while is that there is a difference between _incomplete_ named params, and _incorrect_ named params.  I closed off #352 as it got no comments for 5 months, but I still am of the same opinion.

Or have I missed something?!!

StrayAlien:
HI @tarilabs - happy new year to you there.  

Re named params.  Great.  I'll remove the test.  

Except for the specific case of `"23:00:50Z"` vs `"23:00:50+00:00"` I think you are saying that if a given time or datetime has an @location, then it can only be semantically equivalent with the same time or datetime with the same @location.  Same for offset.  That is, if a time or datetime has an @location, it does not have a semantic equivalent with a numeric offset, and vice versa.  

If so, I'll update the tests.

StrayAlien:
All updated as per comments.

tarilabs:
Happy new 2022 @StrayAlien , thank you for the new commit 2387298

> That is, if a time or datetime has an `@location`, it does not have a semantic equivalent with a numeric offset, and vice versa.

Yes, but very important to remind ourselves is that is from the perspective of the `is()` FEEL builtin semantic, so for this function we need to distinguish from numeric offset of the ISO 8601 just plain standard, from the `@location` which is an extension by the DMN spec itself.

The OMG RTF group concur having approved: https://issues.omg.org/issues/DMN14-86 which contain imho all needed examples.

At the last TCK 2021-12-14 discussed with the group we concurred on the same again.
(Octavian also mentioned it could be worthy in the next OMG RTFs to expand for this behaviour to apply also to Duration; I am neutral on this additional comments, as personally I'd like to see first the RTF convene on the more granular date/times.)

I will mark this as ready for the next milestone based on https://github.com/dmn-tck/tck/pull/412#issuecomment-994004604 and previous @opatrascoiu comment of approval in this https://github.com/dmn-tck/tck/pull/412#issuecomment-986057822

tarilabs (approved):

PR #413, opened 2021-02-21
DMN1-3 exponent. Checking for decimal values.

StrayAlien:
DMN 1.3 Made a small change the exponent spec replacing 'integer' with 'number' for the exponent - the range remains the same but nothing about it being an integral number.

As it was explicit before that an expression like `2 ** 1.5` was not legal, it now is not explicit.  So some tests here to verify that     `2 ** 1.5` returns null, but something `2 ** 1.0` is okay.  Additionally, added a test that the base may also be a decimal like `2.5 ** 3`. 

Comments welcome.


StrayAlien:
All, I realised this test suite is actually redundant.  It is covered by the new 100-arithmetic suite. So, I am cancelling this PR and have added the new tests here to that suite.

PR #420, opened 2021-03-11
Augment TCK Validator detect duplicate id(s) in TCK xmls

tarilabs:
this should automatically perform some of the feedback I've seen in recent PRs :)

### DIFF
--- a/runners/dmn-tck-validation/pom.xml
+++ b/runners/dmn-tck-validation/pom.xml
@@ -10,6 +10,12 @@
   <description>Validate testCase xml files for both DMN model and test case definition.</description>
   <dependencies>
     <dependency>
+      <groupId>org.omg.dmn</groupId>
+      <artifactId>tck-runner</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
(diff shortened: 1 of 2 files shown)


StrayAlien:
Thanks @tarilabs - excellent. I did recently add this my own runner so, hopefully, we'll not see that again from me. :-). Btw, as an aside, the testcases xsd does not enforce unique ids in tests.  Might be a good idea to do so, but .... we'd have a fair few of things to amend.  Many of the ids used in our tests are not valid XML ids.

tarilabs:
https://github.com/dmn-tck/tck/pull/420#issuecomment-797057147
> the testcases xsd does not enforce unique ids in tests. Might be a good idea to do so, but .... we'd have a fair few of things to amend.

It's a valid comment, but IIRC this was discussed waaay back and the decision was not to use schema `ID`, and indeed having it as schema `string`, but avoid duplicates.

> Many of the ids used in our tests are not valid XML ids.

yes, and frankly speaking just adding this check now in the "super partes" validator module I believe is just the easiest.
(there was not even CI integrated back at those days, but nowadays we have :) )

PR #422, opened 2021-03-15
assertions on value domain compliance

StrayAlien:
All,

A suite of tests for assertion on ‘allowed values’ and, I guess what you could call “constrained types”.

To my mind, this is one of the last remaining 'fundamental' areas that is not already in progress and we have not covered in the TCK.  We’ve covered type conformity, but not ‘value domain conformity’.  That is, "what do we do, in all circumstances, when something has allowed values, or is domain constrained".

A ‘constrained type’ (my term, not in spec) is a type that either directly or indirectly has its domain restricted by allowed values (or input/output values in decision tables).

By 'directly, I mean (say) a string type or a date with allowed values in the item definition.

By ‘indirectly’ I mean, for example, a list type where the element type it holds is a string type with allowed values, or a context type with a property that has allowed values, and so on.  Maybe a list of contexts that have a list property with a constrained element type.

So, we have to cover not just (say) assigning a string to a variable with allowed values, we have to consider the validity of entire object structure of lists and contexts - and if somewhere deep inside is a non-domain-compliant value, then the assignment has to be null.

There are a _lot_ of places in the DMN model that can have a typeRef, and thus be influenced by domain restrictions.  Additionally, all scalar types (and speculatively the context type) could have allowed values.

* All TExpressions may have a typeRef for their return value.
* Invokables and FunctionDefinitions have parameters with typeRefs as well as return values with typeRefs.
* A lambda may have typed params, and you can also trick DMN into giving one a return type as well.
* A decision table output clause may have a typeRef as may an outputEntry  
* A decision table output clause may also have ‘output values’.
* A decision table input clause may have ‘input values’.
* A context entry may have a typeRef
* A relation column may have a typeRef
* Invocations have results and parameter binding typeRefs

So, quite a lot really, and every place there may be a typeRef, a value may become null if it is not ‘domain compliant’.

So, taking a very deep breath, here are tests that I hope cover it all - with the exception of external Java calls.  I’m on the fence on that one as I don’t believe a Java callout should even exist in a spec like this.  I’ve also not added domain tests where the allowed values are contexts - though I think it is probably legal (happy for opinions).

Not every place is tested with every scalar type - that'd be too much.  There are a number of simple tests for each scalar type doing compliant/non-compliant assignments to variable and inputs.  But, when we move into functions and params and return types and so on, it is mostly just using a string type with two allowed values “foo”, and “bar” in different guises, like as param types or function return types, or output clause types and so on.

Apologies, I know these tests can be very hard to follow so I have tried to keep them as tight as I can.   Hopefully not too many large test suites to follow this one - maybe just 'inequality'.

So in this suite:

for each scalar type
- type a decision variable and assign it a non/conforming value
- type an input variable and assign it a non/conforming value

Then sanity check the other places with a constrained string type, or some other type directly/indirectly using it.

BKM variable param
BKM variable return
BKM encapsulatedLogic param
BKM encapsulatedLogic result
lambda param
lambda result
function defn param
function defn result
decision service param
decision service result
literal expression
list - boxed and literal
context - boxed and literal
contextEntry
relation
relation column
decision table
decision table output clause
decision table output values
decision table output entry
decision table input values
decision table input/rule
invocation result
invocation param

Comments welcome.

### DIFF
--- a/TestCases/compliance-level-3/0104-allowed-values/0104-allowed-values.dmn
+++ b/TestCases/compliance-level-3/0104-allowed-values/0104-allowed-values.dmn
@@ -59,14 +59,14 @@
     </dmn:itemDefinition>
 
     <!-- a type suitable for a relation - a list of contexts, though
-    each with a contrained property -->
+    each with a constrained property -->
     <dmn:itemDefinition name="tFooBarRelation" isCollection="true">
         <dmn:itemComponent name="col1">
             <dmn:typeRef>tFooBarString</dmn:typeRef>
         </dmn:itemComponent>
     </dmn:itemDefinition>
 
-    <!-- a list with a contrained string type -->
+    <!-- a list with a constrained string type -->
     <dmn:itemDefinition name="tFooBarList" isCollection="true">
         <dmn:typeRef>tFooBarString</dmn:typeRef>
     </dmn:itemDefinition>
@@ -116,7 +116,7 @@
     <dmn:itemDefinition name="tDirectionsString">
         <!-- note this type uses another string type as a base.  We expect
         the allowed values to be 'overridden' here.  That is, the allowed values
-        in tFooBarString will not longer apply -->
+        in tFooBarString will no longer apply -->
         <dmn:typeRef>tFooBarString</dmn:typeRef>
         <dmn:allowedValues>
             <dmn:text>"left", "right"</dmn:text>
@@ -449,7 +449,7 @@
     </dmn:decision>
 
     <dmn:decision name="a_function_typed_lambda" id="_a_function_typed_lambda">
-        <dmn:variable name="a_function_typed_lambda" />
+        <dmn:variable name="a_function_typed_lambda" typeRef="tFooBarFunction" />
         <!-- the literal expression (that returns a lambda) is typed with a function type -->
         <dmn:literalExpression typeRef="tFooBarFunction">
             <dmn:text>(function(a: tFooBarString) a)</dmn:text>
@@ -496,7 +496,7 @@
     </dmn:decision>
 
     <dmn:decision name="a_function_typed_funcdef" id="_a_function_typed_funcdef">
-        <dmn:variable name="a_function_typed_funcdef" />
+        <dmn:variable name="a_function_typed_funcdef" typeRef="tFooBarFunction" />
         <!-- the function definition is typed with a function type -->
         <dmn:functionDefinition typeRef="tFooBarFunction">
             <dmn:formalParameter name="a"/>
@@ -638,7 +638,7 @@
         </dmn:literalExpression>
     </dmn:decision>
 
-    <!-- Decision service accepting an contrained param and echoing it.  The DS variable has no return type -->
+    <!-- Decision service accepting an constrained param and echoing it.  The DS variable has no return type -->
     <dmn:decisionService name="a_decisionservice_with_typed_param" id="_a_decisionservice_with_typed_param">
         <dmn:variable name="a_decisionservice_with_typed_param" typeRef="tFooBarDSFunctionWithTypedParam"/>
         <dmn:outputDecision href="#_ds_echo_from_inputdecision_decision"/>
@@ -826,7 +826,7 @@
 
     <!-- decision with typed literal expression echoing an (untyped) input -->
     <dmn:decision name="decision_literal_expression_001" id="_decision_literal_expression_001">
-        <dmn:variable name="decision_literal_expression_001" />
+        <dmn:variable name="decision_literal_expression_001" typeRef="tFooBarString" />
         <dmn:informationRequirement>
             <dmn:requiredInput href="#_tUntypedInput"/>
         </dmn:informationRequirement>
@@ -836,7 +836,7 @@
     </dmn:decision>
 
     <dmn:decision name="decision_context_expression_001" id="_decision_context_expression_001">
-        <dmn:variable name="decision_context_expression_001" />
+        <dmn:variable name="decision_context_expression_001" typeRef="tFooBarContext"/>
         <dmn:informationRequirement>
             <dmn:requiredInput href="#_tUntypedInput"/>
         </dmn:informationRequirement>
@@ -869,7 +869,7 @@
     </dmn:decision>
 
     <dmn:decision name="decision_list_expression_001" id="_decision_list_expression_001">
-        <dmn:variable name="decision_list_expression_001" />
+        <dmn:variable name="decision_list_expression_001" typeRef="tFooBarList" />
         <dmn:informationRequirement>
             <dmn:requiredInput href="#_tUntypedInput"/>
         </dmn:informationRequirement>
@@ -895,7 +895,7 @@
     </dmn:decision>
 
     <dmn:decision name="decision_relation_expression_001" id="_decision_relation_expression_001">
-        <dmn:variable name="decision_relation_expression_001" />
+        <dmn:variable name="decision_relation_expression_001" typeRef="tFooBarRelation"/>
         <dmn:informationRequirement>
             <dmn:requiredInput href="#_tUntypedInput"/>
         </dmn:informationRequirement>
@@ -932,7 +932,7 @@
     input as an output entry.  The DT has a variable typeRef so we can assert on
     the decision out -->
     <dmn:decision name="decision_dt_expression_001" id="_decision_dt_expression_001">
-        <dmn:variable name="decision_dt_expression_001"/>
+        <dmn:variable name="decision_dt_expression_001" typeRef="tFooBarString"/>
         <dmn:informationRequirement>
             <dmn:requiredInput href="#_tUntypedInput"/>
         </dmn:informationRequirement>
@@ -1032,7 +1032,7 @@
     <!-- decision with a single output decision table having no inputs that just echoes a given
     input as an output entry.  The output clause is constrained by outputValues -->
     <dmn:decision name="decision_dt_expression_005" id="_decision_dt_expression_005">
-        <dmn:variable name="decision_dt_expression_005"/>
+        <dmn:variable name="decision_dt_expression_005" typeRef="tFooBarString"/>
         <dmn:informationRequirement>
             <dmn:requiredInput href="#_tUntypedInput"/>
         </dmn:informationRequirement>
@@ -1071,7 +1071,7 @@
 
     <!-- invocation defines a constrained return type -->
     <dmn:decision name="decision_invocation_expression_002" id="_decision_invocation_expression_002">
-        <dmn:variable name="decision_invocation_expression_002"/>
+        <dmn:variable name="decision_invocation_expression_002" typeRef="tFooBarString"/>
         <dmn:informationRequirement>
             <dmn:requiredInput href="#_tUntypedInput"/>
         </dmn:informationRequirement>
--- a/TestCases/compliance-level-3/0104-allowed-values/0104-allowed-values.dmn
+++ b/TestCases/compliance-level-3/0104-allowed-values/0104-allowed-values.dmn
@@ -536,7 +536,7 @@
     <!-- BKM with a function typed encapsulatedLogic.  Used to assert on return type
     compatibility -->
     <dmn:businessKnowledgeModel name="a_bkm_with_function_typed_logic" id="_a_bkm_with_function_typed_logic">
-        <dmn:variable name="a_bkm_with_function_typed_logic" />
+        <dmn:variable name="a_bkm_with_function_typed_logic" typeRef="tFooBarFunction" />
         <dmn:encapsulatedLogic typeRef="tFooBarFunction">
             <dmn:formalParameter name="a"/>
             <dmn:literalExpression>
--- a/TestCases/compliance-level-3/0104-allowed-values/0104-allowed-values.dmn
+++ b/TestCases/compliance-level-3/0104-allowed-values/0104-allowed-values.dmn
@@ -696,6 +696,39 @@
         </dmn:list>
     </dmn:decision>
 
+    <!-- compliant assignment of single value to 'list' variable type with constrained element type  -->
+    <dmn:decision name="list_003" id="_list_003">
+        <dmn:variable name="list_003" typeRef="tFooBarList"/>
+        <dmn:literalExpression>
+            <dmn:text>"foo"</dmn:text>
+        </dmn:literalExpression>
+    </dmn:decision>
+
+    <!-- non-compliant assignment of single value to 'list' variable type with constrained element type  -->
+    <dmn:decision name="list_004" id="_list_004">
+        <dmn:variable name="list_004" typeRef="tFooBarList"/>
+        <dmn:literalExpression>
+            <dmn:text>"baz"</dmn:text>
+        </dmn:literalExpression>
+    </dmn:decision>
+
+    <!-- compliant assignment of singleton list value to string constrained element type  -->
+    <dmn:decision name="list_005" id="_list_005">
+        <dmn:variable name="list_005" typeRef="tFooBarString"/>
+        <dmn:literalExpression>
+            <dmn:text>["foo"]</dmn:text>
+        </dmn:literalExpression>
+    </dmn:decision>
+
+    <!-- non-compliant assignment of singleton list value to string constrained element type  -->
+    <dmn:decision name="list_006" id="_list_006">
+        <dmn:variable name="list_006" typeRef="tFooBarString"/>
+        <dmn:literalExpression>
+            <dmn:text>["baz"]</dmn:text>
+        </dmn:literalExpression>
+    </dmn:decision>
+
+
     <!-- decision echo value for an input that is a list with a constrained element type -->
     <dmn:decision name="list_input_decision_001" id="list_input_decision_001">
         <dmn:variable name="list_input_decision_001"/>
@@ -757,6 +790,26 @@
         <dmn:variable name="context_004" typeRef="tFooBarListContext"/>
         <dmn:literalExpression>
             <dmn:text>{prop1: ["foo", "foo", "baz"]}</dmn:text>
+        </dmn:literalExpression>
+    </dmn:decision>
+
+    <!-- non-compliant assignment to 'context' variable type with constrained property type
+    The intention here is to show that singleton list conversion/equality does not take place
+    inside a structure -->
+    <dmn:decision name="context_005" id="_context_005">
+        <dmn:variable name="context_005" typeRef="tFooBarContext"/>
+        <dmn:literalExpression>
+            <dmn:text>{prop1: ["foo"]}</dmn:text>
+        </dmn:literalExpression>
+    </dmn:decision>
+
+    <!-- non-compliant assignment to 'context' variable type with constrained property type
+        The intention here is to show that singleton list conversion/equality does not take place
+        inside a structure -->
+    <dmn:decision name="context_006" id="_context_006">
+        <dmn:variable name="context_006" typeRef="tFooBarListContext"/>
+        <dmn:literalExpression>
+            <dmn:text>{prop1: "foo"}</dmn:text>
         </dmn:literalExpression>
     </dmn:decision>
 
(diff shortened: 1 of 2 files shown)


tarilabs:
@StrayAlien again thank you for suggesting these tests.

I've been spending days during this week to look at these tests and I am not yet sure I would concur with _all_ of them, but this was an interesting set of tests to look at.

As a first step, I would like to focus on a bare minimum of corrections which I believe are required to set a baseline.

1. the current .dmn model defines for the default namespace the DMN's namespace, hence any cross-typeRef references such as `tFooBarString` cannot happen this way when the default namespace is set as the DMN's namespace. There are two options to correct for this.
OptionA setting a XML prefix for the model namespace such as `xmlns:montera="..."` and then cross-typeRef reference as `montera:tFooBarString`.
OptionB setting the **XML** default namespace as the model namespace, and deal with xml element with a prefix.
In general most vendors out there set for the convention outline in the OptionB, hence I suggest we stick we that, as other models in the TCK are doing.
I have prepared a PR which does this already, it could be attained easy with a regex such as `(<\/?(?![!]))` and replacing as `$1dmn:` in say VSCode.

2. I am pretty positive it was missing a Decision to be named `context_boxed_002` as in analogous sematically of `context_002` but boxed. I realized this by looking at the test case.

```xml
    <dmn:decision name="context_boxed_002" id="_context_boxed_002">
        <dmn:variable name="context_boxed_002" typeRef="tFooBarContext"/>
        <dmn:context>
            <dmn:contextEntry>
                <dmn:variable name="prop1"/>
                <dmn:literalExpression><dmn:text>"baz"</dmn:text></dmn:literalExpression>
            </dmn:contextEntry>
        </dmn:context>
    </dmn:decision>
```

3. in the "some esoteric stuff" (scary name btw) the Decision `override_004` was meant to be named `override_002` I suppose, again by looking at the test case.

To be clear, I am yet to further analyse some of these tests.

Ref: https://github.com/StrayAlien/tck/pull/6
I hope this PR ☝️  helps in the meantime to set a baseline for further discussions.

StrayAlien:
Hi @tarilabs - thank you for spending the time looking into the PR.  This is certainly a difficult PR to get your head into so your time is appreciated.  Thank you - the errors you noted are valid.  No problems if you feel some test may not be correct - let's discuss.  As usual, my PRs are often 'my take' on things and a starting point for further clarification.  All good.

I've accept the changes above.  No sweat - though, some further discussion of the name spacing thing would be great.  Not quite sure I understand what is expected there.

It seems I had added few extra tests that I had not yet pushed to this PR so I just added them. They are assertions around singleton lists and are fairly straight forward.

re 'esoteric' - getting your head into the DMN spec is not a trivial Sunday afternoon activity.  Some dark corners in there that are not so easy to understand!

Thanks again.

TestCases/compliance-level-3/0104-allowed-values/0104-allowed-values.dmn
```diff
@@ -0,0 +1,1092 @@
...
+
+    <dmn:itemDefinition name="tDirectionsString">
+        <!-- note this type uses another string type as a base.  We expect
+        the allowed values to be 'overridden' here.  That is, the allowed values
+        in tFooBarString will not longer apply -->
+        <dmn:typeRef>tFooBarString</dmn:typeRef>
+        <dmn:allowedValues>
+            <dmn:text>"left", "right"</dmn:text>
+        </dmn:allowedValues>
+    </dmn:itemDefinition>
```
tarilabs, line 124: TCK 2021-07-15
the group is not ootb agreeing that "allowedValues" do any overridden behaviour, instead is restriction on restricitons on the extension hierarchy chain.

But we also need to check in the DMN Specification if there are any explicit mandates about the checking for allowedValues. And potentially something about the XML spec?

tarilabs: (we will also raise this to the RTF group)

TestCases/compliance-level-3/0104-allowed-values/0104-allowed-values.dmn
```diff
@@ -0,0 +1,1092 @@
...
+
+    <!-- decision with typed literal expression echoing an (untyped) input -->
+    <dmn:decision name="decision_literal_expression_001" id="_decision_literal_expression_001">
+        <dmn:variable name="decision_literal_expression_001" />
+        <dmn:informationRequirement>
+            <dmn:requiredInput href="#_tUntypedInput"/>
+        </dmn:informationRequirement>
+        <dmn:literalExpression typeRef="tFooBarString">
+            <dmn:text>tUntypedInput</dmn:text>
+        </dmn:literalExpression>
```
tarilabs, line 835: TCK 2021-07-15 
wfrom the DMN1.3 spec

> If an instance of Expression that defines the output of a Decision element includes a typeRef, the referenced type SHALL be the same as the type of the containing Decision element.

So this DMN is not conformant in the way it is right now.

We will check if we have some general provision in the spec which mandates that for any Expression defining a typeRef, the typeConformance shall be expressed.
Specific example.
For a ContextEntry to specify a typeRef, that the entry value must conform to the typeRef.

Before going forward with this kind of test would be meaningful to clarify the standing ground for type conformance check in the Expression.

We will also check, and eventually raise to the RTF group, if we will see that the DMN specification is missing an "analogous" and corresponding statement for the Expression which are the body of a FunctionDefinition encapsulatedLogic of BKM, where we already agreed is the return type of the function definition.

TestCases/compliance-level-3/0104-allowed-values/0104-allowed-values.dmn
tarilabs: (there could be the need to add something also on the chapter of Equivalence and Conformance)

TestCases/compliance-level-3/0104-allowed-values/0104-allowed-values.dmn
StrayAlien: HI @tarilabs - re the quote above.  That sure feels out of date.  If anything, they should _conform_ I guess.  But, as it is, I think you are saying that if a decision expression has a typeref, then the containing decision must also have that exact typeRef also.  Doesn't seems right (IMO).

TestCases/compliance-level-3/0104-allowed-values/0104-allowed-values.dmn
StrayAlien: Hi @tarilabs @opatrascoiu  - yes, the spec is not very helpful here.  The whole notion of allowedValues is fairly vague. Precisely why I did this suite. 

If it helps, I was going on this in "7.3.3 ItemDefinition metamodel"

> If an ItemDefinition element contains one or more allowedValues, the allowedValues specifies the complete range of values that this ItemDefinition represents.

No notions of extension or hierarchy.

StrayAlien:
@tarilabs - could I ask you to explain a little further the reasoning behind your 'namespace' comments above?  Apologies, I don't quite understand.  The various ways in which naming and references happen across a DRG and imports and the diagrams is a bit of a mind-mincer.  But I thought model-local references do not need to be qualified.  And qualified type references are for imports .. and then they use the name of the import, not definition XML namespaces.

tarilabs:
Wrt

> But I thought model-local references do not need to be qualified.

Agreed, as long as they are consistent with the model default namespace.

> And qualified type references are for imports .. and then they use the name of the import, not definition XML namespaces

Correct, but the DMN-Import again identifies the (target, imported) model default namespace.

image

You are 100% correct when you expect that referencing `typeRef="tPerson"` in a current DMN model, would make reference to an ItemDefinition `tPerson` in the current document. And similarly, when `typeRef="myimport1.tPerson"` in a DMN model is when you have DMN-Imported another DMN, under the DMN-Import-name of `myimport1`.

However to guarantee for backward compatibility (to DMNv1.1) most vendor have settled for OptionB as I described in https://github.com/dmn-tck/tck/pull/422#issuecomment-877004141, that is the XML default namespace ought to be consistent with the _semantic_ DMN namespace (the namespace in the DMN's Definitions element).

Your original PR did not have this consistency, your typeRef were consistent with the semantic DMN namespace, but not consistent with the original XML default namespace.

When you prefix your XML element with the `dmn:` (or alternative prefix) for the namespace of the OMG spec `https://www.omg.org/spec/DMN/20191111/MODEL/`, by side effect, now you can define your XML document default namespace to become consistent with the semantic DMN namespace easily --that is you can set the XML `xmlns=` attribute in root to receive the same value as `<definitions namespace=` value.

I hope I was able to explain clearer?
If not, don't hesitate to let me know, hope this helps!

TestCases/compliance-level-3/0104-allowed-values/0104-allowed-values.dmn
tarilabs: I am still under the impression this DMN model is _not_ conformant based on https://github.com/dmn-tck/tck/pull/422#discussion_r670424829 and our mutual understanding of BKM and its contained typeRefs (in the variable and the body).

> If an instance of Expression that defines the output of a Decision element includes a typeRef, the referenced type SHALL be the same as the type of the containing Decision element.

The spec as it stands right now is quite clear: if a Decision has an expression which defines a typeRef (eg: the Decision's LiteralExpression, the Decision's boxed context, etc) then the same typeRef SHALL be the same as the type of the Decision (in the Decision's variable, that is).

On this premise, one could say for Decision you could just specify the typeRef in the Decision's variable (as specifying the same typeRef in the Decision's boxed expression is redundant).

I am open to revise this perspective, but it would require a change in the spec, for which I'm not sure the rationale... when would you want to specify a _different_ typeRef for the Decision's variable Vs the Decision's boxed expression?

tarilabs:
To recap my understanding on this PR 

1. the DMN model is not conformant due to https://github.com/dmn-tck/tck/pull/422#discussion_r670424829
the easy solution to fix this in the DMN model under test, is to use the typeRef in the Decision's variable only (instead of placing it in the boxed expression and leaving the Decision's variable unspecified).
I am happy to revise with the group if we believe we DMN spec should state something different, but we would also need to articulate on the motivations, as I can't personally see reasons why we would want to specify a different typeRef for the Decision's variable Vs the Decision's boxed expression logic.

2. if I recall correctly in the TCK and RTF (informal) discussion on the topic of https://github.com/dmn-tck/tck/pull/422#pullrequestreview-707289195 , the understanding in that example is that `tDirectionsString` would never conform, because there's no provision in the DMN spec which justify the `allowedValues` override.
Eg: a string that must be either `"left", "right"` cannot be at the same time _ALSO_ a string that must be either `"foo", "bar"`
Yes this could be problematic, as in one modeler might end up defining a type which would never find a valid instance value, but "tool smarts" or interpreters could avoid modeler to step into that situation entirely upfront or emit warnings.

Thank you again for these tests, but I believe we should converge on these 2 points first.

StrayAlien:
Hi @tarilabs - thanks for the comments.  Re 

> If an instance of Expression that defines the output of a Decision element includes a typeRef, the referenced type SHALL be the same as the type of the containing Decision element.

I guess this is subtle, but note that the test does not define a typeRef for the decision, so the decision type is implicitly the same as the expression.  In effect, they are the same type.  And, vague as I am, I think that was likely my angle here.  In practice they should be equal, but if _both_ are provided they _must_ be equal.

If it is the case that "if a decision expression has a typeref, then the containing decision must also have that exact typeRef" then I'm happy to change the test.  If so, I guess the same thing should also apply to invocables like:

> If an instance of Expression that defines the output of a Invocable element includes a typeRef, the referenced type SHALL be the same as the type _return type_ of the containing Invocable's (function definition) typeRef.

//// 

Regarding `tDirectionsString`.  when doing this test I knew it would be discussion point - so I'm glad for the comments, but, my reading of the spec is from this (7.3.3 ItemDefinition metamodel):

> If an ItemDefinition element contains one or more allowedValues, the allowedValues specifies the **complete range of values that this ItemDefinition represents**. If an ItemDefinition element _does not contain_ allowedValues, its range of allowed values is the full range of the referenced typeRef

If it contains allowedValues then those values are the _complete_ range of values that apply.  _Only_ if there is no allowed values does the 'full range' of the referenced typeRef come into effect.  So, an ItemDefinition's own allowed values takes precedence over the referenced type.  The full range of the referenced type is only relevant when the item definition has no allowed values.  This is what this test is to assert.

Personally, I agree with you - it should be cumulative - because that is what makes sense.  But ... as I read it, that is not what the spec says.  It looks like an _override_.  Perhaps that paragraph should have appended to it:

> If an ItemDefinition contains allowedValues, and its referenced typeRef also contains allowedValues, then the ItemDefinition's allowedValues represent a further restriction on the range of values allowed by the referenced typeRef. 

Thanks again for the comments @tarilabs - always insightful.

baldimir:

baldimir:

opatrascoiu:
I agree with @StrayAlien about the way allowedValues are calculated when there is a chain of ItemDefinitions (see the tests for primitive types and the override tests). IMO the standard is clear on this (see 7.3.3 ItemDefinition metamodel as mentioned by @StrayAlien above) 

`If an ItemDefinition element contains one or more allowedValues, the allowedValues specifies the complete range of values that this ItemDefinition represents. If an ItemDefinition element does not contain allowedValues, its range of allowed values is the full range of the referenced typeRef.
`

I believe there is also agreement on how to check the allowedValues : it's a unitary tests that evaluates to a boolean condition.
 
The remaining part is to agree on when the constraints are checked. We should probably discuss it in the next meeting.

We also need to make sure the model is well-formed. For example, when present the typeRefs specified in the DRGElement, variable and enclosed expression are the same. We also need to check if the typeRefs are specified for all input datas and parameters as we agreed a while back. I suggest adding the above conditions in the validation phase.

StrayAlien:
Thanks @opatrascoiu .  Re:

> Also we need to agree on what happens when a nested structure (e.g. list of complex objects) is invalid. I can think of two options: the whole structure is null or only the invalid components are null.

My feeling is it has to be the entire object structure becomes null - it must be viewed as a complete thing - think XSD / JSON schema validation - it is either correct, or not correct.  Otherwise I think we'll run into very hard to track down bugs where suddenly (say) input data has a (coerced) null deeply nested inside simply because it does not match the required complex type.  IMO, all or nothing.

Re:

> We also need to check if the typeRefs are specified for all input datas and parameters as we agreed a while back. 

I don't agree we need to enforce types are specified.  Types are not required in the spec and runtimes need to work within that.  If the type is not specified the runtime must work with what it has.

opatrascoiu:
@StrayAlien Regarding the presence of typeRef for inputs & params, We already discussed this a while back. I believe you were present. I believe we agreed on this one. After the meeting @tarilabs added a validation that breaks the build when typeRef for inputs is missing.

StrayAlien:
Thanks @opatrascoiu. "roger that". I can't say I agree with the notion of making life easier for runtimes to pass tests though.  Quite the opposite I think, the bar should be set (very) high and 'looseness' exercised.  In the end, with a high bar, we all benefit.  If the TCK only has tests that are 'helpful' to runtimes, then no one wins.  </rant>

opatrascoiu:
I started to review the tests. So far I have some comments regarding test lambda_003

The test touches several areas of the DMN model: 2 type definitions and 2 decisions. The execution is equivalent to the following pseudocode:

type tFooBarString = string / "foo", "bar"
type tFooBarFunction = (a: undefined) -> tFooBarString

var a_function_typed_lambda : tFooBarFunction = (function(a: tFooBarString) a)
var decision_lambda_expression_002 = a_function_typed_lambda(tUntypedInput)

When the decision a_function_typed_lambda is executed the value of the expression is (a: tFooBarString) -> tFooBarString and it does not conform to the type of the decision ((a: undefined) -> tFooBarString). See the type conformance of functions (section 10.3.2.9.2 - the FEEL functions follow the "contravariant function arguments and covariant return type" and 10.3.2.13 Function semantics). Hence, the variable a_function_typed_lambda is bound to null. This leads to the invocation to fail hance the variable decision_lambda_expression_002 is bound to null.

I propose to change the definition of tFooBarFunction to (a: tFooBarString) -> tFooBarString. This will allow the execution to lead to the value expected in the test file - 'foo'. I don't think the proposed change preserves the intent of the test.

StrayAlien:
Thanks @opatrascoiu . Appreciate the comment.  I'll dig into it!  Admittedly, the PR is now 2 years old, so I also need to look at it with fresh eyes.  Thanks.

opatrascoiu:
@StrayAlien I am adding more details to clarify my view: the type of _(function(a: tFooBarString) a)_ does not conform to the type of _(function(a) a)_ as per definition of conformsTo for function types (see 10.3.2.9.2 Type Conformance)

(a: tFooBarString) -> tFooBarString <: (a:undefined) -> undefined iff

tFooBarString <: undefind (**covariant** return type part) and
undefined <: tFooBarString (**contravariant** parameter type)

The first one is true (as the target type is missing, hence no restriction on the semantic domain), the second one is false (the target type is a string, hence there is a restriction of the semantic domain) (see 10.3.2.13 Function semantics). 

I think we should add some wording regarding the undefined types in 10.3.2.9.2 to localize the specification in one single place / section.

opatrascoiu:
@StrayAlien It seems to me that some tests do not follow that pattern we agreed in a previous call:
- constraints are checked when a variable is bound to a value
- if a part of a nested value (e.g. instances of complex types such as Person or lists) fails a constraint, the result is null and not a nested value with null parts).

For example, test decision_list_expression_002. 

The evaluation of the decision leads to the evaluation of a list expression that has an element that does not match the allowed values for the element of the list. So far so good. 

However, in this case the constraint check happens at the decision level, hence the whole list expression fails the validation. I think the expected should be null not [null].  

I am missing something?

StrayAlien:
HI @opatrascoiu - thanks for the comments.  I really appreciated it.  I'm still getting my head around the function type comments, so, I'll come back to that one in a moment.  

These typing tests can really do your head in, so, thanks for investing the time.  It is important that we get these right!

// preamble ///

In essence, IMO, typeRefs are like a 'gate' that checks the results of values that pass through them against a given type.  So, a typeRef at the decision level will check the results of the decision expression against that type.   If the decision expression result is a compound (context/list) value then its deeper domain checks are performed and any failure will render the _entire_ result as null. 

Any typeRefs that are explicitly defined _inside_ a decisions expressions are the same.

In a way, it like an onion.  Where an _explicit_ typeRef is defined in a deeply nested expression, that is a type safe 'gate'.  Type safe failure will passes a null 'out'.  

When a value hits a 'gate' it is asserted against the typeRef, and if it is a compound type either the entire value passes or fails.

// the tests //


For `decision_list_expression_002`, it goes like this:

`decision_list_expression_002` has no typeRef and is _untyped_.  It has a list expression that is also _untyped_.  So, neither the decision nor the list asserts a type.

However, the list element is a literal expression that _is_ typed and is expecting a string of either "foo" or "bar", but it gets "baz", so just the list _element_ becomes null.

Therefore the untyped list expression becomes `[null]`.  `decision_list_expression_002` is _untyped_ so therefore has no domain constraints.  Its result is purely the result of the expression, which is `[null]`.

I guess the point is, nothing in the test is typed as `list<string>` in there anywhere so there is no compound type being assert against.

`decision_relationrow_expression_002` is similar, the decision is untyped, and so is the relation.  But the relation column is typed so the null-coercion happens in isolation at the column level.  No compound type involved.

I think `decision_dt_expression_002` and `decision_dt_expression_006` are the same.  

///

Re the function types.  I have to admit, I've having troubles understanding why a return type of 'undefined' does not conform to `tFooBarString`.  Like I said, this can really do your head in!

We do not know the type of 'undefined' so it has no domain and therefore it is not checkable until runtime.  At runtime the domain can be checked against, and, if it fails, then return `null`.  In my mind this is like how we treat 'Any'.  For example, in the type lattice, 'Any' does not comply with 'string'.  But, we've agreed that an 'Any' cannot be checked until its domain becomes known at runtime.  So, in terms of type checking a list signature like:

`list<a: Any>` complies with `list<string>` at _compile-time_, BUT, at runtime if the list has any non-string elements then it will fail and result in `null`.

In essence, 'Any' or undefined ('unknown' as I call it) conforms to everything until such time as the domain is known at runtime.

Discussed some here: https://github.com/dmn-tck/tck/pull/438.  My early assumption was that 'Any' was like Java `Object` where this is not legal:

```
Object o = "foo";
String s = o;  // type error 
```

But, in DMN terms this seems not to be the case.  The domain of an 'Any' can only be known at runtime and thus is effectively excluded from type checking until its domain becomes concrete.  IMO, same for 'undefined'.

Admittedly, that discussion faded away, but, to comply with what @tarilabs  was saying, simply deferring the checking of `Any` (and 'undefined') until runtime was the answer, and, I think the right thing to do.

So, with `lambda_003`, I'm saying it complies because it is uncheckable until runtime, and at runtime, the return type complies with `tFooBarString`, so... all good.

... if that makes any sense ....

baldimir:
Based on the discussion on the meeting, let's read the comments and decide on the next meeting.

StrayAlien:
closing, will raise again as smaller PRs in line with current spec.  Conversations here still very relevant though.

PR #423, opened 2021-03-22
Fix typeRef for bkmn.encapsulatedLogic

opatrascoiu:
@StrayAlien @tarilabs 

Based on the discussions we had as part of PR #385 I believe that:

1. `bkm.variable.typeRef` and `bkm.encapsulatedLogic.typeRef`, when present should be function type
2. `bkm.encapsultatedLogic.expression.typeRef` (function definition body) is the output type of the BKM

Looking forward for your comments.
Octavian

### DIFF
--- a/TestCases/compliance-level-3/0082-feel-coercion/0082-feel-coercion.dmn
+++ b/TestCases/compliance-level-3/0082-feel-coercion/0082-feel-coercion.dmn
@@ -148,9 +148,9 @@
 
     <businessKnowledgeModel name="bkm_004" id="_bkm_004">
         <variable name="bkm_004"/>
-        <encapsulatedLogic typeRef="tNumberList">
+        <encapsulatedLogic>
             <formalParameter name="arg"/>
-            <literalExpression>
+            <literalExpression typeRef="tNumberList">
                 <text>arg</text>
             </literalExpression>
         </encapsulatedLogic>
@@ -158,9 +158,9 @@
 
     <businessKnowledgeModel name="bkm_005" id="_bkm_005">
         <variable name="bkm_005"/>
-        <encapsulatedLogic typeRef="number">
+        <encapsulatedLogic>
             <formalParameter name="arg"/>
-            <literalExpression>
+            <literalExpression typeRef="number">
                 <text>[arg]</text>
             </literalExpression>
         </encapsulatedLogic>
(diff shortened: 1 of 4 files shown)


StrayAlien:
Hi @opatrascoiu - some notes:

* I think the same also applies for the other invokable - Decision Services.  Many of those have a typeRef.  Maybe many are 'Any'.  But as @tarilabs notes, this is moot.  Should that be disallowed?  After all 'Any' is not a function type.  My gut feeling is that `typeRef="Any"` is semantically the same as not having a typeRef at all, so likely should be legal - but, the spec gives no guidance here.  
* Also, some of the 'coercion' tests are set up to catch differences in types - so to keep the intention of the tests it may be better to introduce a function type itemdef (with the current test typeRef as the return type) and have the BKM / DS / Func Defn typeRef point at that new type defn.  Moving the typeRef to the logic level may still keep the test green but the test may no longer be doing what it was intended to do. Though, the new 'allowed values' suite will also catch some of these occurrences.

opatrascoiu:
Hi @StrayAlien 

I checked the DS as well before PR. I did not find any obvious typeRef error.

Regarding Any. It is the upper type in the lattice hence any type conforms to Any; it can hold any value (similar to Object in Java). From the precise modelling perspective using Any is not recommended as it could lead to coercions to null (e.g. when trying to add 2 Anys). 

Missing type is not the same as type being Any. There are places in spec where type conformance is not checked when type is missing.

StrayAlien:
Hi @opatrascoiu. Re "Missing type is not the same as type being Any".  Okay.  Though, yes, any type conform to Any.  But, it that what we're after here?  We're saying the typeRef for invokables and func defn have to be function types.  That is not the same as Any.  Yes, a function type confirms to Any .. but so does a string.  What sense does it make to have (say) the variable on a BKM as 'Any', where the logic _must_ be a function type ..., surely, it implies the BKM variable typeRef must also be a function type.

Re "There are places in spec where type conformance is not checked when type is missing.".  Do you have a page/section?  I'm interested.

tarilabs:
> 1. `bkm.variable.typeRef` and `bkm.encapsulatedLogic.typeRef`, when present should be function type

Hi Octavian, I'm a bit confused on this one.

I previously understood https://github.com/dmn-tck/tck/pull/385#issuecomment-771503773 and we confirmed that 
`bkm.variable.typeRef` is indeed meant to be a function type
but
`bkm.encapsulatedLogic.typeRef`was the return type of the function represented by the bkm.

The linked PR indeed "moves down" the function-output-type _from_ the bkm's variable _to_ the bkm's encapsulated logic, and my understand per last meeting.

It seems to me this PR is proposing some else, can you clarify why the difference, please?
Thanks!

opatrascoiu:
@tarilabs 
Yes, I / we got it wrong last time. TypeRef of variable and encapsulatedLogic should be the same: function type. The type of the body is the output / return type (see my previous comment).

PR #432, opened 2021-04-28
Drools version refresh and Apr 2021 results

tarilabs:


tarilabs (commented):
A Vendor's result refresh, merging.

Issue #430, opened 2021-04-14
Missing typeRefs for InputData


opatrascoiu:
There are several places where InputData does not have a typeRef.

This  makes type inference  and validation of semantics at compile time impossible.

Our users are not allowed to execute above DMs.

I am inclining towards making InputData.variable.typeRef mandatory.

Lets discuss.

StrayAlien:
Hi @opatrascoiu - as much as I sure like the ideas of locking things down for better typing, I think the idea of a 'compiler' and what it can or cannot do is an implementors choice / concern.  The spec does not mention 'compile' so I'm inclined to think that making mods to the spec to cater for compilation is perhaps maybe not in line with the intention there?  Happy for thoughts, but I'd be keeping it 'dynamic' and assume that an implementation may have no 'compiler' at all.

As an aside, we cater for it by having an internal 'unknown' type that everything confirms to, and it conforms to everything.  And thus, runtime checks are required for when the compiler cannot determine typing.  Some limitations apply to its usage and the compiler will complain/err in some situations.  But, in practice it seems to work and some effort / checking is deferred until runtime.

As a suggestion, rather than mandating it in the spec, I'd be voting for a) a suite of tests where the inputData have no typeRefs ... or b) alternatively (and this would be my preference), removing all typeRefs from inputData in the entire TCK _except_ for only those places where it is being tested for specifically.  Like there are some of those in this suite: https://github.com/dmn-tck/tck/pull/422 to catch input data not conforming to 'constrained types' with allowed-values somewhere in them.

Happy for thoughts.

opatrascoiu:
@StrayAlien I think we all agree that DMN / FEEL is a typed language:
- each value belongs to a semantic domain / type
- each DRG element has a variable
- there are conformance rules & implicit conversions

Based on the above type checking is a must. It can be static checking / compile time or dynamic checking / runtime. 

For some DRGElements (e.g. Decisions) the type can be inferred at compile time or runtime. How would you perform type checking for InputData is there is no typeRef ?

StrayAlien:
Hi @opatrascoiu / @tarilabs - my sincere apologies for missing the meeting.  I really really thought it was tonight, not last night.  I was looking forward to it actually.  Some great things to discuss.  Apologies to you and the crew - I'm feeling a little stupid about that one.  Mi culpa.

[...]

PR #436, opened 2021-05-06
Address #430 in TestCases/compliance-level-3/0092-feel-lambda

tarilabs:
ref https://github.com/dmn-tck/tck/issues/430#issuecomment-822363590


tarilabs:
As discussed on the 2021-05-13 TCK meeting, proceeding to merge.

PR #438, opened 2021-05-15
change inputData typeRef to 'string' (from 'Any')

StrayAlien:
Hi @tarilabs / @opatrascoiu - modding the recent inputData typeRef change to the 'get value' function to be 'string'.  Having it as 'Any' will give a `null` result, not what is expected.  Apols, I should have jumped on it earlier.

Why `null`?  When the typeRef was empty the input had no explicit type, but now we have given it 'Any' - so it now has an explicit type.

The parameter domain for the `get value` function is `context, string`.  `Any` does not conform to `string`.  So giving the inputData `Any` does mean it can accept a string value (as `string` conforms to `Any`).  But, in turn, it means it cannot be used where a string parameter is required because `Any` does not conform to `string`.

Yes, the _value_ of the input is a string, but as we've given an _explicit_ type to the inputData that really is `Any` and not based on the runtime content of it.

That is my take - hopefully that makes sense.

Greg.

PS: Octavian, I guess that underlines the differences between not having a type, and having an 'Any' type as per a previous discussion as to why they are not the same thing.

### DIFF
--- a/TestCases/compliance-level-3/0080-feel-getvalue-function/0080-feel-getvalue-function.dmn
+++ b/TestCases/compliance-level-3/0080-feel-getvalue-function/0080-feel-getvalue-function.dmn
@@ -10,7 +10,7 @@
     <description>FEEL built-in function 'get value(m, key)' in unspecified category</description>
 
     <inputData name="input_001" id="_input_001">
-        <variable name="input_001" typeRef="Any"/> <!-- used to externally supply FEEL:string or other type of values defined in the TCK test file -->
+        <variable name="input_001" typeRef="string"/> <!-- used to externally supply FEEL:string or other type of values defined in the TCK test file -->
     </inputData>
 
     <decision name="decision_001" id="_decision_001">


StrayAlien:
Just an additional note here, but this crosses over the issue raised by @opatrascoiu here: https://github.com/dmn-tck/tck/issues/386.  At this point there are some tests that provide incorrect data according to type.  If type checking is enforced for function / invokable arguments the two tests discussed in that PR also need amendment / inclusion, and I support that change.

I guess it is also interesting to consider, in practice, what the `Any` type means on an inputData.   `Any` only conforms to `Any`, so by making inputData typeRefs=`Any` you are effectively limiting the use of the inputData to places that conform to `Any`  .. which is pretty restrictive. 

Anyways, I'd like to fix those two tests, but, there is already some PR activity surrounding 0082-feel-coercion so it'd be great to get these in quickly so we can keep moving:

* https://github.com/dmn-tck/tck/pull/423
* https://github.com/dmn-tck/tck/pull/409

Life gets awkward when there are like 3 or 4 PRs in on the same suite for different reasons.

tarilabs:
Hi @StrayAlien ,

I beg to differ, I do not agree with this PR.

Here:

https://github.com/dmn-tck/tck/blob/7ba1b16823f2dbb3ad03af31690bab257273b7aa/TestCases/compliance-level-3/0080-feel-getvalue-function/0080-feel-getvalue-function-test-01.xml#L132-L134

you are using `input_001` to check for 

```
get value({a: "foo"}, input_001)
```

So if you now change this from `Any` to `string`, you are no longer checking that `get value({a: "foo"}, 12)` is null and an error; but you are checking that `get value({a: "foo"}, null)` is null and an error, which is not the original intent of the test we agreed on?

StrayAlien:
Hi @tarilabs, thanks.  Good catch on that particular test - I can amend to make that test do as it did before with a new input.  But, that is not actually the point of the PR.  I'm more focussed on test `decision_013`.  Apols - I see I did not explicitly mention that in my OP.

The point of the PR is that 'Any' is not assignable to 'string' and hence the test `decision_013` would give an null, and not the expected outcome.   (yes, the `decision_014` test is providing a null result already but for a different reason - like I said, will amend).

The _value_ of in the `input_001 ` for test `decision_013` will be a string, and that's okay because the type of `input_001` is 'Any'.  But, the `get value({a: "foo"}, input_001)` would be null, because the 'Any' type on `input_001` is not assignment compatible with the second argument being a 'string' type.  That is the point here - that when an inputData (or anything else) such as this is explicitly typed as `Any` then it really only can be assigned where the target of the assignment type is 'Any', or no type is specified.

Please also see my second post here re issue https://github.com/dmn-tck/tck/issues/386 and how it relates to this.

Greg.

tarilabs:
> The value of in the input_001 for test decision_013 will be a string, and that's okay because the type of input_001 is 'Any'. But, the get value({a: "foo"}, input_001) would be null, because the 'Any' type on input_001 is not assignment compatible with the second argument being a 'string' type. That is the point here - that when an inputData (or anything else) such as this is explicitly typed as Any then it really only can be assigned where the target of the assignment type is 'Any', or no type is specified.

I do not concur with this statement, but I'm happy if we take the time during the next TCK meeting to discuss this point.

tarilabs:
specifically `decision_013`

```
get value({a: "foo"}, input_001)
```

let's do on paper the exercise.

you are supplying context _c_:
```
{ input_001 : "a" }
```

to the literal expression of `get value({a: "foo"}, input_001)`.

What is `type(e)` when e is the first parameter?
```
TYPE({a: "foo"}) = context<"a":T>
where
TYPE("foo") = string
hence
TYPE({a: "foo"}) = context<"a":string>
```


What is `type(e)` when e is the second parameter?
```
e: input_001 // this is not yet in the semantic domain
e: "a" // this is now a value
TYPE("a") = string
```

So you have correct conformance of the parameters when invoking the function 

image

Hence I don't concur, but as stated earlier, I'm happy to take the time during the next TCK meeting to delve into this.
Hope this help.

StrayAlien:
Thanks @tarilabs - yes, the runtime type is 'string', but the given type is 'Any'.  The type lattice says that 'Any' is not assignable to 'string', and that is what we are trying to do here. 

Worth mentioning, I am not religious about this and defending any view - just "seeking the truth" in DMN terms as I think this is fairly important.

It sounds like you are saying 'Any' has no meaning for assignment compatibility checking and that it should be deferred to runtime checking? 

If so, I'm not sure what the point is of the spec of saying that Any is not assignable to string .. if it is never actually enforced?  

To me, trying to assign a `list<Any>` to a `list<string>` should (IMO) be cause for a 'compiler' error - for those that implement a 'compiler'.  So same applies, to my mind, when assigning an 'Any' to a 'string'.  A bit like this:

image

I get 'Any' is a slightly different animal to other types, but, in terms of assignment compatibility checking are you saying it is to be ignored until an actual value is known (ie, runtime)?  I know you sure know the spec better than I do (meant as a compliment! :-) ) but I don't recall anything in there re same except this `the special type Any, which includes all values in D` (10.3.2.1 Semantic Domain).  But, how to interpret such things is left to the DMN mystics.

All comments appreciated.

Greg.

StrayAlien:
@tarilabs - Thought it through and I'm fine with the 'wait until you actually know the value type' thing.  Closing.

PR #443, opened 2021-05-31
Drools version refresh and May 2021 results

tarilabs:


tarilabs:
A vendor results update, merging.

PR #445, opened 2021-06-01
Fix tck properties file

opatrascoiu:


falko:
LGTM

tarilabs:
As reminded and mentioned in the last TCK meeting 2021-06-10, this property file is used for Vendors to explicitly control properties displayed on the website.

As discussed with Octavian during the same meeting, this did not required an ad-hoc refresh of the website, merging now and it will be accounted at the next refresh.

PR #452, opened 2021-06-22
Introduce new DMN-Scala runner

saig0:
* new test runner for the DMN-Scala engine 1.6.2
* publish test results of the DMN-Scala runner
* add GitHub action for the DMN-Scala runner


falko:
No worries. You can also keep it open as a draft if you like.

PR #458, opened 2021-07-13
Introduce new TCK runner for the Camunda DMN-Scala extension

saig0:
* add new TCK test runner for the DMN-Scala engine 1.6.3
* add test results of the DMN-Scala runner
* add GitHub action for the DMN-Scala runner


saig0:
@tarilabs please have a look.
 
/cc @falko

tarilabs:
Hi, glad to hear Camunda are working with also considering the TCK again :) 

I have a couple of comments, please.

1.
@saig0 @falko for your Vendor ("Camunda") website result publishing, are you planning to submit results via this new runner? If so, can you remove as part of this PR the "second choice" of runner here:
https://github.com/dmn-tck/tck/tree/master/TestResults/Camunda
and for your choice of result submission so _the removal of the "second choice" to be part of this PR_ as well, please?
For each Vendor, we usually asked 1 result submission.
If you could do it as part of this PR, it is appreciated, so we record correctly the intention.
If you have alternative feedback on this point, I'm open to hear it too.

2.
We sticked to the convention of submitting results per each **Vendor**.
Your current designation appears to be the **name of the product**.
e.g.: ` TestResults/DmnScala/1.6.3/tck_results.csv`
it should be
` TestResults/Camunda/1.6.3/tck_results.csv`
please if you can maintain the designation correctly?

3.
GHA
I'm more than happy to keep the "second choice" of the previous Camunda runner as also a GH Action, but if you are actually planning to "deprecate" the previous one I would appreciate if you also removed that --in that case.

I hope that clarifies!
Thanks,

falko:
Hi Matteo,

thank you for your feedback. This submission is not replacing or deprecating the existing Camunda submissions. As stated in the properties, Camunda DMN Scala is a community extension for the Camunda Platform to evaluate DMN decisions based on Camunda's FEEL engine. It's a plugin that extends the capabilities of Camunda Platform to enhance the DMN support. At this point, it can be seen as a community preview to what might become the default engine in future versions. This is similar to how FEEL Scala already became the default FEEL engine after being tested with the TCK. Until then we would like to continue two submissions in parallel also in order to get feedback from the community on the new implementation. Both runners are provided in order to make transparent how the different results can be achieved.

Cheers,
Falko

tarilabs (approved):
Thank you @falko for the clarifications in https://github.com/dmn-tck/tck/pull/458#issuecomment-879198851

A Vendor updating with new runner and results, proceeding to merge as requested.

opatrascoiu:
I would change the name of the folder / runner to dmn-tck-runner-camunda-scala instead of dmn-tck-runner-dmn-scala to avoid ambiguities and specify the provider.

tarilabs:
I believe Camunda calls this solution `DMN-Scala` likewise `jDMN` so to speak, but is a relevant comment if @falko or @saig0 want to reconsider their runner naming, or `dmn-tck-runner-camunda-dmn-scala` in that sense.

falko:
Yes, that makes sense. @saig0 could rename the runner to `dmn-tck-runner-camunda-dmn-scala`?

tarilabs:
(I would suggest accordingly the result directory, too) thank you.

falko:
yes

PR #480, opened 2021-10-31
Update Drools version and submit results

tarilabs:


tarilabs:
A vendor's version update and results, merging.

PR #483, opened 2021-11-08
Context entry with dup keys result not allowed

### LINKED ISSUE #400: Debatable / incorrect test case CL3 / 0057-feel-context
According to DMN 1.3 the keys in a context are distinct.

```
10.3.2.6 Context
A FEEL context is a partially ordered collection of (key, expression) pairs called context entries. In the syntax, keys can
be either names or strings. Keys are mapped to strings in the semantic domain. These strings are distinct within a context.
A context in the domain is denoted using bold FEEL syntax with string keys, e.g. { "key1" : expr1, "key2" : expr2, … }.
```

However, the test below assumes that the keys may be overwritten

```
    <decision name="decision008" id="_decision008">
        <variable name="decision008"/>
        <literalExpression>
            <text>{foo: "bar", foo: "baz"}</text>
        </literalExpression>
    </decision>

    <testCase id="008">
        <description>later duplicate name overwrites previous</description>
        <resultNode name="decision008" type="decision">
            <expected>
                <component name="foo">
                    <value xsi:type="xsd:string">baz</value>
                </component>
            </expected>
        </resultNode>
    </testCase>
```

I think we should keep the test case but change the expected value to null as there is a semantic error.

tarilabs:
results in null and error
see https://issues.omg.org/browse/DMN14-178

Resolves #400

### DIFF
--- a/TestCases/compliance-level-3/0057-feel-context/0057-feel-context-test-01.xml
+++ b/TestCases/compliance-level-3/0057-feel-context/0057-feel-context-test-01.xml
@@ -119,7 +119,7 @@
     </testCase>
 
     <testCase id="008">
-        <description>later duplicate name overwrites previous</description>
+        <description>duplicate context entry keys not allowed (DMN14-178), results in null</description>
         <resultNode errorResult="true" name="decision008" type="decision">
             <expected>
                 <value xsi:nil="true"/>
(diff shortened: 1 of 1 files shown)


StrayAlien:
Thanks for clarifying at RTF.  Nice.  Good to get some closure here!

StrayAlien:
@tarilabs - on your branch, would you mind also changing the test description.  It is still "later duplicate name overwrites previous".  Perhaps it should be "duplicate context keys results in null" or something.

PR #489, opened 2021-12-17
Add minimal typeRefs to allow bottom-up type inference

opatrascoiu:
I believe that a while ago we agreed to include typeRefs for inputs unless the test case needs to pass in data of various types from TCK InputNodes. The idea is to have the minimum type info in the DMN models to infer the types from DMN syntactic constructions only.

I believe the test 0092-feel-lambda is not consistent with the above approach. I added the minimal required typeRefs to allow type inference.

There are also two debatable test cases: 0015 and 0016. They pass-in built-in functions with varArgs. AFAIK the standard does not define type conformance for this type of parameters. My preferred approach is to:
   - comment them out
   - extend the  conformsTo spec to support them
   - review them and add them back after we agree on spec

I did not comment them out in this PR. I added some typeRefs for them as well to give us a hint of the required types when we extend the spec.

### DIFF
--- a/TestCases/compliance-level-3/0092-feel-lambda/0092-feel-lambda.dmn
+++ b/TestCases/compliance-level-3/0092-feel-lambda/0092-feel-lambda.dmn
@@ -4,11 +4,46 @@
              xmlns="https://www.omg.org/spec/DMN/20191111/MODEL/" >
     <description>FEEL lambda</description>
 
+    <!-- function definitions -->
+    <itemDefinition name="lambda_number_returns_number">
+        <functionItem outputTypeRef="number">
+            <parameters name="n" typeRef="number" />
+        </functionItem>
+    </itemDefinition>
+
+    <itemDefinition name="lambda_number_number_returns_number">
+        <functionItem outputTypeRef="number">
+            <parameters name="n1" typeRef="number" />
+            <parameters name="n2" typeRef="number" />
+        </functionItem>
+    </itemDefinition>
+
+    <itemDefinition name="lambda_list_number_returns_number">
+        <functionItem outputTypeRef="number">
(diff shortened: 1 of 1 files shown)


StrayAlien:
Hi @opatrascoiu - happy new year to you.  Re the typeRefs - I recall we were adding types for input datas, not formal params (?). So in that sense, (IMO) I don't think all the typeRefs in this PR are necessary.

The type conformance of built in functions only becomes an issue if you wish to typeRef them - as per above.  Without those typeRefs I don't think it is an issue - they could be called in any of their overloaded guises.

My reasoning for the above stuff if that, even though it has types, DMN is not strongly-typed. It is not Java/etc, maybe more like loose-typescript.  So, types are there to help, but not required. 

As usual, I'm not religious about it, and happy for opinions, but I don't think they are required here.

opatrascoiu:
Here is my rationale for adding the typeRefs for parameters:
- type inference is done bottom-up, from inputs to outputs
- InputData are the inputs for the body / expression of decisions
- Similarly, Parameters are the inputs for the body / expression of function definitions

tarilabs:
I have nothing specific about this PR either as Greg, but since ItemDefinitions have interdependencies, this PR need shall fix the DMN model (default) namespace, per https://github.com/dmn-tck/tck/pull/422#issuecomment-877004141 point 1.

As a reminder, that is to ensure backward compatibility but most importantly that the ItemDefinition are defined in the correct namespace.

At the time of writing this comment, the default namespace is the OMG's DMN namespace, hence the ItemDefinition gets defined not in the DMN semantic namespace of the author, but in the OMG's namespace.

The fix proposed in https://github.com/dmn-tck/tck/pull/422#issuecomment-877004141 point 1 should apply easily here, just a matter of regex in a text editor

Hope this helps!

PR #491, opened 2021-12-17
jDMN results for December 2021

opatrascoiu:


tarilabs:
A vendor's result refresh. Merging.

PR #492, opened 2021-12-27
Update Drools version and submit results

tarilabs:


tarilabs:
Vendor submitting update and result refresh, merging.

PR #495, opened 2022-01-12
corrections and assertions on invalid params

StrayAlien:
As per recent discussions that an invocation with params that do not conform to the to-be-invoked-function formal param types, or has unknown named params, will 'short-circuit' to null and the actual target of invocation will never be executed ...

Here are some fixups in the TCK to reflect that.  

As a boxed Invocation is really a function call with named params and therefore (IMO) also subject to the same conditions as any other invocation with named params I've added one extra test to assert that same null result.

All comments and feedback welcome. 

PS: I apologise for the 'noise' with the spaces but I have git config set at `core.autocrlf input` as per guidelines for mac/linux.

### DIFF
--- a/TestCases/compliance-level-3/0082-feel-coercion/0082-feel-coercion-test-01.xml
+++ b/TestCases/compliance-level-3/0082-feel-coercion/0082-feel-coercion-test-01.xml
@@ -71,7 +71,7 @@
             </expected>
         </resultNode>
     </testCase>
-<!-- Ref: https://github.com/dmn-tck/tck/pull/238#issuecomment-497744315 
+<!-- Ref: https://github.com/dmn-tck/tck/pull/238#issuecomment-497744315
      Expected result is a wrapped element not justified with DMN spec
     <testCase id="decision_006">
         <description>decision returns value but type is list - coercion to list</description>
@@ -132,11 +132,10 @@
     </testCase>
 
     <testCase id="decision_bkm_002">
-        <description>pass non-conforming type to bkm - bkm
-            input param is null coerced</description>
-        <resultNode name="decision_bkm_002" type="decision">
+        <description>pass non-conforming type to bkm - bkm is never invoked</description>
+        <resultNode name="decision_bkm_002" type="decision" errorResult="true">
             <expected>
-                <value xsi:type="xsd:boolean">false</value>
+                <value xsi:nil="true"/>
             </expected>
         </resultNode>
     </testCase>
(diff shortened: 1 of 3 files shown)


StrayAlien:
bump

tarilabs (commented):
I do recall about some discussion for this topic, but if the memory served me well, we didn't have a 1:1 equivalence (yet?) in the spec between a boxed invocation and purely FEEL based one.

So @baldimir what needs to be done here is to check the proposed changes, accordingly to the semantic defined in the spec.

About the whitespace "noise" typically you can instruct Git to maintain the original source.
Otherwise, you can consider mass-swipe the TCK repo for Linux CR formatting, and then have the 1 single only commit which is changing the formatting, in a repo `.git-blame-ignore-revs` in the root.
ref https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/

baldimir:
I couldn't find anything about the behaviour in the spec, but I am surely missing something as I am still learning the spec. @StrayAlien @tarilabs or other @dmn-tck/contributors, could you please help me to find a place in the spec where to look? E.g. I was checking 7.3.7 Binding metamodel - there is written: 

"When the Invocation element is executed, each InformationItem element that is referenced as a parameter
by a binding in the Invocation element is assigned, at runtime, the value of the bindingFormula."

This doesn't say anything what should happen when there is a parameter that is not referencing any InformationItem. Does this binds by default to null in general in the spec in similar cases elsewhere? For the other cases with BKMs I couldn't find information how it should look.

opatrascoiu:
There are a few sections that deal with invocations:

10.3.2.13.2 User-defined functions
10.3.2.13.1 Built-in Functions
10.3.2.9.4 Type conversions

The ambiguity (result is null or body is invoked with null arguments) was created by the fact that 10.3.2.9.4 does not clearly defines the concept of 'implicit conversion'.

There is an issue with RTF to handle this https://issues.omg.org/browse/DMN15-44 and the implicit conversion date -> date and time.

StrayAlien:
10.2.1.3 Boxed Invocation

> Formally, the meaning of a boxed invocation is given by the semantics of the equivalent textual invocation, e.g., function-valued expression(parameter1: binding expression1, parameter2: binding expression2, ...).

So, a boxed invocation is semantically the same as a function call with named arguments.  Hence the reason for this PR.  If invalidly typed or unknown param names are provided then it is an invalid invocation, hence the null shirt-circuited null result.

As I recall, we've updated a number of other tests (with literal expressions passing bad params) to ensure they return a short-circuit null, so, given boxed follows the same semantics, just doing the same here.

By "short-circuit", I mean the function is never invoked at all - it is the runtime determining it is invalid invocation and simply returning null.

Additionally, it is not only about 'implicit conversion' but also about the attempt to bind to an unknown formal parameter.  

.. section 7.3.7 Binding metamodel

> The parameter names in the Binding elements SHALL be a subset of the formalParameters of the calledFunction.

So, using an invalid name as a boxed invocation binding, or as a function call named argument renders the invocation invalid.  Hence the 'short-circuit' null.

opatrascoiu:
@baldimir @StrayAlien As promised I reviewed the current changes. They look fine to me.

Thanks again for contribution @StrayAlien

PR #511, opened 2022-09-19
update to Java 11 version

gautric:
- updating github action
- updating pom.xml


tarilabs:
I believe the TCK worked already with JDK11+.
(IIRC I've always checked with 1.8, 11 and 17).

This PR would bump min Java requirement to 11, but not sure why?

baldimir:
I personally agree with bumping the default to 11 as Java 8 very old now.

tarilabs (commented):
https://github.com/dmn-tck/tck/pull/511#issuecomment-1268295797 was provided.

baldimir:
Also this PR is not just about raising the minimal Java version, but actually running the PR checks on Java 11. That is a good thing in my opinion.

baldimir (approved):

baldimir:
Hi @gautric, thank you for the contribution. I am sorry for the delays. Could you please rebase your PR?

baldimir:
We discussed this on a TCK meeting and there are no objections to this. I will ping contributors to find out last time if anyone has objections. If not and it is rebased, I will merge this.

baldimir:
One question on the meeting about this was that for the marshaller module, there would need to be dependencies checked, but there is already a profile in the pom.xml for java 9+ that gets automatically enabled, that adds required dependencies. So this is not a problem.

PR #521, opened 2022-10-25
Update Camunda test results

saig0:
* update the test results of Camunda
* update the test results of Camunda-DMN-Scala


saig0:
@tarilabs here are the new test results from the Camunda runners. Please merge. :cookie: 

cc: @falko

falko:
thx @saig0

PR #524, opened 2022-10-26
Add minimal typeRefs to allow bottom-up type inference

opatrascoiu:
This is a replacement of  https://github.com/dmn-tck/tck/pull/489 that was closed automatically when repo was refreshed.


opatrascoiu:
I suggest to have another PR to fix the namespace issues raised by @tarilabs as there are other test files with the same problem.

baldimir:
Hi @dmn-tck/contributors, could you please review?

baldimir:
@StrayAlien @tarilabs what do you think about this PR please? I see you commented the original one that was closed.

baldimir:
Thanks @opatrascoiu for the tests. I personally think we should have the namespaces fixed in this, so each code change gets merged as correct. The other cases that are already in the repository should be fixed separately. I can create a GitHub issue for that and take a look. Please let me know if that makes sense.

opatrascoiu:
@baldimir @tarilabs 

Firstly, I think a MR should be focused one single change at a time to avoid complexity, in this case extra typRefs. 

Secondly, we need to discuss the namespace issue in more detail.  In DMN 1.1 the types were referenced by constructions like prefix:name hence the comment is valid for DMN 1.1. However, this has changed in DMN 1.2 to prefix.name where the prefix is optional:

```
Type Ref: definition
A pointer to a specification of the datatype of the possible outcome values for this Decision, either an ItemDefinition,
a base type in the specified expressionLanguage, or an imported type.

ItemDefinifion definition:
An ItemDefinition element SHALL have a typeRef, which is a QName that references, by namespace prefix and
local name, either an ItemDefinition in the current instance of Definitions or a built-in type in the specified
typeLanguage or a type defined in an imported DMN, XSD, or other document. In the latter case, the external
document SHALL be imported in the Definitions element that contains the instance of ItemDefinition, using
an Import element. For example, in the case of data structures contributed by an XML schema, an Import would be
used to specify the file location of that schema, and the typeRef attribute would reference the type or element definition
in the imported schema. If the type language is FEEL the built-in types are the FEEL built-in data types: number, string,
boolean, days and time duration, years and months duration, time, and date and time.
```

There is an error is spec: typeRef is not an XML QName anymore due to the new syntax. The question is if the prefix is needed when the type is defined in the same model. My reading of the above paragraph is that the prefix if not needed. For types defined outside an import and prefix are needed. Same approach with other languages.

All the tests we have seem to take this approach, including the example from the DMN spec (see 0004-lending.dmn).

I suggest:
1: we merge this PR as we agreed with the existing changes in PR. 
2. discuss the namespace issue and if we agree that the prefix is needed, clarify that in spec and change all the test filles accordingly.

StrayAlien:
Hey all, I'll "not die on this hill" as they say, but, again, my personal feeling is that the tests should be absent of any type information unless the test is _specifically _ asserting on the typed aspects of the spec.  IMO, a test should have the minimum in it to permit making its assertion.  Types cloud things and add noise if not needed.  Additionally, these tests in this TCK here are not to assist runtimes with anything - the spec has types as optional and I feel we should treat them as such - unless, of course, the test is asserting on types.

The test changed in the PR are already in master, they are valid tests without the type info and clearly the type info is not required for the assertions being made.  As runtimes work without them as they are, why add the extra noise?

But, no religion here - if you guys feel it is important - okay, but, I don't agree sorry.

falko:
@saig0 do you have an opinion on this?

baldimir:
I agree with @StrayAlien. When types are optional and the changed tests are valid tests without the type, we shouldn't change them as that would change the "focus" of the tests. If there are some type related compatibility tests needed, we could introduce a new set of tests. @opatrascoiu what do you think please?

opatrascoiu:
I strongly disagree with types being optional. AFAIK (from RTF calls) in DMN types are optional so that partial models can be saved (see compliance levels). We spent a lot of time and effort to add a type lattice and define the conformance & conversion rules. 

I am not a fan of dynamic semantics as it increases the risk of breaking at runtime (think about use cases dealing with transactions worth millions or sending a satellite in space). 

Historically, all the main stream languages, starting with Fortran, Cobol and ending with Java and C# have static semantics. Even Python added type annotations to reduce the risk, although too late.

StrayAlien:
Hi @opatrascoiu - you sure are closer to the RTF than I am, and it is an interesting notion that the types are optional to support partial models.  But, that is not how the spec presents itself (IMO).  I'm not sure that comparing DMN/FEEL to Java etc is useful here - DMN/FEEL is its own thing and even if other languages support certain features that does not mean intrinsically that DMN/FEEL must also.  Just because types are good does not mean DMN/FEEL mandates them.  You're reading this on a page that is likely powered by Javascript .... and it has no typing. And regardless how one feels abut dynamic languages I'm not sure that comes into play here either.  I also like strong typing - preferring TypeScript to JavaScript any day, and I cut my teeth on C / Java.  I like types.

I agree the spec is pretty poor in many cases and maybe it is here as well, but, I don't see anything that says a complete/canonical/executable/correct model 'SHALL' include types.  I only see the opposite.  Like, say, the given examples for the `sort()` function do not include types.   The FEEL grammar says they are optional - and I'm not sure the FEEL grammar can be 'partially correct' to support model interchange (or can it????).  Surely a FEEL expression is either right or wrong (??).

To my eyes, the type lattice is there to support types _when present_ and to facilitate 'instance of'.  

So, speculatively, if a correct model 'SHALL' have types and is thus strongly typed ... and types are only optional to support interchange of incomplete models, then I guess that means in a complete/correct model every place in the spec / schema that says a typeRef is optional SHALL have a type?  .... and all FEEL expressions that could define type SHALL define types?  Like every `function()` in code SHALL define types?

If the above is not the case (not every place that may have a typeRef _must_ have a typeRef), then how does one know _which_ typeRefs to leave off?  Surely they are either optional or they are not optional. 

Get my drift?  If DMN/FEEL is to be strongly typed then everything needs to be typed, and if not everything is typed how does the spec let us know _which_ things to type and which not to type?

My take: types are optional.   They are _helpful_ and surely _desirable_, but optional.

Btw, I mean no offence with all this.  Not trying to be stubborn or whatever, but, just expressing an opinion.  Btw, if the spec does disagree with the above, cool.  Lemme know.  It would not be the first time I am wrong!

baldimir:
Getting this merged as agreed on the TCK meeting. The type definitions are an optional extra information that can be ommited by vendors in their implementations.

PR #527, opened 2022-11-17
Update from DMN 1.3 to DMN 1.4 namespaces for DMN/FEEL

SimonRinguette:
As discussed on the 2022-11-17 meeting I've updated the test cases from DMN 1.3 to DMN 1.4


falko:
@saig0 FYI

StrayAlien:
hey @baldimir / @falko - as a suggestion here but happy for opinions (and disagreement!).  Perhaps we tag the current master and then get this into master so we can move forward with 1.4.  I see no breaking changes in the 1.4 changebar, and a number of small corrections in the new spec are actually already present in the TCK tests.

I already have initial TCK tests suites ready for PR for all the new 1.4 functions and will be doing suites for the new boxed expressions in the coming days.  

I'm suggesting because, rather than get into PR hell with PRs on PRs while we still have existing PRs for 1.3 outstanding etc it might just make life easier. 

But, happy to work how you guys feel is best.

falko:
If it makes your contributions easier, I'm all for it.

baldimir:
Yes, if we move the spec version, we need to make a tag first, so the 1.3 tests can be found historically. For the old PRs - I feel that we should close them as they are stuck. If some of those things are needed, they can be reopened again when we have 1.4 on the master.

StrayAlien:
hey @baldimir - some of those older PRs are awaiting RTF clarification so maybe they could stay.  The trouble with deleting PRs is we lose the chat history - even if it is often very confusing!

But, getting 1.4 into master is something that can be discussed with the group.  I suspect vendors are already in production with the 1.4 impls based on the current spec, so IMO not so big of an issue.  If the spec changes people will have to change their impls anyways, and we'll have to changes some tests.  I don't see too much controversial stuff in the new spec though, it has created some inconsistencies (https://github.com/dmn-tck/tck/discussions/504) and has some of the the usual silly not-quite-thought-out gaps in it, but, all fairly small.

PR #533, opened 2022-11-30
Refresh webpage 2022-11.

baldimir:


tarilabs (approved):
as suggested in chat, you might prefer to delete detail_ and overview_ file locally, before generating the report with `mvn` for the generate-website, and then submit PR.

That way only latest report for each entry+vendor will be kept, with no dangling/deep pages on the docs/

baldimir:
@tarilabs redid it like that.

Issue #536, opened 2023-01-16
Clarify FEEL `null` handling in `FilterExpression`


nikku:
### Describe the context

The [DMN 1.4 specification](https://www.omg.org/spec/DMN/1.4/Beta1/PDF) gives users some pointers on how FEEL should handle invalid inputs and data formats (cf. [additional context](#user-content-additional-context)). From what it reads to me the way is `null` cohersion. Invalid (incompatible operations, access, or invocations) yield `null`. That intention is not explicitly expressed though (you need to read between the lines, interpret what is actually specified.

`null` cohersion around `FilterExpression` is not clearly specified and may lead to engines behaving slightly different:

```
// given
list = [
  { a: 1 },
  { a: 2 },
  { b: 1 }
]

// when
result = list[ a < 3 ];
```

### Expected Behavior

```
// then (?)
result = [ 
  { a: 1 },
  { a: 2 }
]
```

### Alternative Interpretations

To be clear there exists a number of different possible interpretations:

#### strict `null` handling, blowing up

```
a is not present on list entry `{ b: 1 }`
```

#### strict type format handling

Coherse to empty list as not all of the list members match `{ b }`.

```
result = []
```

### Additional Context <a name="additional-context" />

I'm aware of the following cases of `null` cohersion, explicitly defined by the spec:

#### Access function with wrong arguments

```
date(1, 2, 3, 4, 5, 6) = null
```

#### Unboxing of single value lists

```
["a"] + "a" = "aa"
```

----

Related to #537

barmac:
I brought this issue up in the DMN 1.5 RTF meeting today. We looked together into the specification and agreed that it's sufficiently specified. On [page 109](https://www.omg.org/spec/DMN/1.4/Beta1/PDF#page=125), one can read:

> The filter expression is evaluated for each item in list, and a list containing only items where the filter expression is **true** is returned. E.g:
[ {x:1, y:2}, {x:null, y:3} ][x < 2] = [{x:1, y:2}]

So for the third element, the evaluated filter is:
```feel
{ b: 1 }.a < 3

// which is equal to
null < 3

// which results in
null
```

As `null != true`, the filter expression should drop the item.

As a follow-up, we should create a PR with a test case which ensures the behavior above between the implementations.

nikku:
Thanks for clarifying @barmac.

I can see if I can add a TCK test case to cover two topics that you mention:

```
// access of undefined members in a Context
{ x: 1 }.y = null
```

```
[...]

PR #540, opened 2023-01-19
test(feel): verify non-member filter expression handling

nikku:
Related to https://github.com/dmn-tck/tck/issues/536


TestCases/compliance-level-3/0110-feel-filters/0110-feel-filters.dmn
```diff
@@ -0,0 +1,80 @@
...
+  <decision name="Decision1" id="d_Decision1">
+    <variable typeRef="boolean" name="Decision1"/>
+    <literalExpression>
+      <text>[ 1, 2, 3, 4 ][ item > 2 ] = [ 3, 4 ]</text>
+    </literalExpression>
+  </decision>
+  <decision name="Decision2" id="d_Decision2">
+    <variable typeRef="boolean" name="Decision2"/>
+    <literalExpression>
+      <text>[ { x: 1, y: 2}, {x:2, y:3} ][ x = 1 ] = [ { x: 1, y: 2 } ]</text>
```
barmac, line 12: Not sure if we need the first two as they are already covered in https://github.com/dmn-tck/tck/blob/master/TestCases/compliance-level-3/0069-feel-list/0069-feel-list.dmn#L183

However, I couldn't find existing test cases for the rest of the expressions.

opatrascoiu:
What is the point of creating a new test file and not using the existing one for lists ?

I would prefer to see all the tests about list / filters in one single place - its easier to figure out the coverage.

nikku:
> What is the point of creating a new test file and not using the existing one for lists ?

There is no specific purpose beyond "testing filters" rather than lists.

I was not aware that "list" tests filters, too.

nikku:
And if you check our existing tests you'll see that we have a dedicated `paths` spec, too.

TestCases/compliance-level-3/0110-feel-filters/0110-feel-filters.dmn
nikku: I was just following along existing good practices to test the actual examples, in a single place. Happy to remove what has already been covered elsewhere.

However I'd suggest to introduce explict specs for filters, as we have for `path`: https://github.com/dmn-tck/tck/pull/540#issuecomment-1396727466

opatrascoiu:
There is a test file for paths as well 0090. Did you look at them ?

nikku:
> There is a test file for paths as well 0090. Did you look at them ?

Yes.

opatrascoiu:
Than why not adding the extra tests in the same file? Same as for lists - one place.

nikku:
I'm happy to put tests in the places that you find most reaonable.

The question is though, what are we testing? Are we testing lists, or are we testing filters?

The section + tests I'm referring to is called `lists and filters`, hence I grouped all tests into a single filters file:

image


----

How I understand you:

* Add filters test from that section to lists spec
* Add path test from that section to path spec

I can do that, if you confirm.

baldimir:
Based on the discussion on the TCK call, it can be in a separate test file. Filters are a bigger feature that could be separate. @dmn-tck/contributors please review.

opatrascoiu:
I have a few comments:

1. I think that

// non member path filtering
[ { x: 1 }, { y: 2 } ].y = [ 2 ]
[ { x: 1 } ].y = [ ]

are not correct, they should be

// non member path filtering
[ { x: 1 }, { y: 2 } ].y = [ null, 2 ]
[ { x: 1 } ].y = [ null ]

According to 10.3.2.5 Lists and filters (see last paragraph) the selection is evaluated for every element in the list and the result is collected in a list.

2. The above tests are not related to filters, they are about lists & selection / paths. They should not be in a test file called filter-test. 

I strongly recommend moving all the tests from this file in 0069-feel-list  to consolidate all the tests about lists (index and selection) in the same place.

3. The tests use a pattern that is not consistent with the rest of the test cases: an expression is evaluated than compared with the expected value in a decision. I recommend changing the tests by moving the second operand of the equality in the test file (xml file) as expected value. The test becomes more readable: it is more obvious what syntactic construction is under test.

nikku:
@baldimir Thanks for the reminder. This is on my TODO-list. I just was not able to get into it.

nikku:
@opatrascoiu Regarding the following review hint:

> The tests use a pattern that is not consistent with the rest of the test cases: an expression is evaluated than compared with the expected value in a decision.

The reason I did not do this is that it is much harder to write tests that way, and also to read them for outsiders (as XML). But happy to perform that change if you think it benefits the team.

opatrascoiu:
@nikku Apologies fir delay. Thank very much for the contribution. They look good to me.

baldimir (approved):

PR #545, opened 2023-02-15
[DMN 1.5] Add tests for equality operators in positive unary tests to the TCK suite

kbowers-ibm:
Adding tests for positive unary tests.

### DIFF
--- a/TestCases/compliance-level-3/1131-feel-unary/1131-feel-unary-test-01.xml
+++ b/TestCases/compliance-level-3/1131-feel-unary/1131-feel-unary-test-01.xml
@@ -5,33 +5,113 @@
         <label>Compliance Level 3</label>
         <label>FEEL unary</label>
     </labels>    
-    <testCase id="number_01">
+    <testCase id="number_not_equal">
         <description>number: e1 in unary tests</description>
-        <resultNode name="number_01" type="decision">
+        <resultNode name="number_not_equal" type="decision">
             <expected>
                 <value xsi:type="xsd:boolean">true</value>
             </expected>
         </resultNode>
     </testCase>
-    <testCase id="number_02">
+    <testCase id="number_not_equal_a">
         <description>number: e1 in unary tests</description>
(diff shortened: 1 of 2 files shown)


baldimir:
Thanks Kennedy. If the group agree to merge this, it needs to be done after the current master is branched for DMN 1.4.

baldimir:
@dmn-tck/contributors please review

SimonRinguette:
Tested on the Trisotech engine and it works. 

Not sure why this would go to `dmn-tck:dmn1.5` instead of master. because its DMN 1.3

opatrascoiu:
I think is because support for the operators = and != in simple positive unary test (before endpoints) was added in DMN 1.5 https://issues.omg.org/browse/DMN15-108

SimonRinguette:
Make sense

dsntk:
There are already tests for **1131** in main. Test case directory and file names have to be adjusted. Otherwise existing tests will be overridden.

baldimir:
I will fix it. Thanks for noticing this.

opatrascoiu:
The tests look good to me apart from the filename (1131 is already taken by another test case)

StrayAlien:
I think these cases are already covered in "0072 feel in tests".

opatrascoiu:
@StrayAlien The == and != are not covered as they were introduced in DMN 1.5. How about we add tests for == and != in 0072 to keep all trhe in tests in one place?

StrayAlien:
Thanks @opatrascoiu .  That what this current PR is for: https://github.com/dmn-tck/tck/pull/668.   It asserts on those operators for FEEL types - except list because this is still uncertain I think.  (as well as uncommenting some null tests we had in there to see if they now are valid)

opatrascoiu:
I checked the tests in PR #668 and they cover the ones here.

@kbowers-ibm @baldimir I think it makes sense to close this MR and focus the review on the one mentioned by @StrayAlien as it groups the test cases for unary tests in an already existing test file.

PR #547, opened 2023-02-16
Update from DMN 1.4 to DMN 1.5

falko:
The OMG Architecture Board has approved DMN 1.5. It will still take a while for the new version to be published but its contents are now sealed and already available to OMG Members to implement in their tools.


baldimir:
This is replaced by https://github.com/dmn-tck/tck/pull/650

PR #549, opened 2023-03-03
Update Drools results to 8.34.0.Final.

baldimir:


tarilabs (commented):
looks as I would expected for a version update 👍

PR #569, opened 2023-06-22
Add zeroes before decimal points

dmntk:
# Add zero before decimal point

### Reason 1 (general)
Omitting leading zeros for type **xsd:decimal** is correct, but not for FEEL **number** when converted to string.

### Reason 2 (DMNTK specific)
Test runner for DMNTK validates both, the returned value **and** the string representation of this value, so this test file must be tweaked after each update from master TCK repository to make the test passing.

## Change proposal
Add leading zeroes to expected decimal values for rates in test file.

## Change risks
I guess that other runners validate only the value, so nothing should change for them. When loading decimal value without leading zero in Java works, then loading the same value with a leading zero should make no harm.

Please share your thoughts ;-), comments welcome!

### DIFF
--- a/TestCases/compliance-level-3/0014-loan-comparison/0014-loan-comparison-test-01.xml
+++ b/TestCases/compliance-level-3/0014-loan-comparison/0014-loan-comparison-test-01.xml
@@ -29,7 +29,7 @@
 							<value xsi:type="xsd:string">Oceans Capital</value>
 						</component>
 						<component name="rate">
-							<value xsi:type="xsd:decimal">.03500</value>
+							<value xsi:type="xsd:decimal">0.03500</value>
 						</component>
 						<component name="points">
 							<value xsi:type="xsd:decimal">0</value>
@@ -43,7 +43,7 @@
 							<value xsi:type="xsd:string">eClick Lending</value>
 						</component>
 						<component name="rate">
-							<value xsi:type="xsd:decimal">.03200</value>
+							<value xsi:type="xsd:decimal">0.03200</value>
 						</component>
 						<component name="points">
 							<value xsi:type="xsd:decimal">1.1</value>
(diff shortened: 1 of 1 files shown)


dmntk:

StrayAlien:
Hi @dmntk - no leading zeroes is valid for a FEEL number - (if I understand the grammar correctly!).  I don't think the string representation of a number is relevant here.  The test result expects a number and the runtime should return a number.  (The spec does not define what a 'number' looks like in a return result - but, in the TCK we do make some assumptions about max decimal points and so on)

If the runtime returns a 'canonical' version of a number with leading zeroes, all well and good, but, it is still a number.  I don't see any reason to consider a number return value as being a 'canonical' string value and basing a comparison on that - with or without leading zeroes that value is the same.  The test runner (IMO) should just concern it self with whether it got the correct number back or not.

dmntk:
Hi @StrayAlien - you are right, the grammar allows the number without leading zeros:
```
35. numeric literal = [ "-" ] , ( digits , [ ".", digits ] | "." , digits ) ;
```
but this the other end of conversion: string -> FEEL number.
Specification says nothing about the other end: FEEL number -> string, except removing trailing zeros.

So you are right, there is no need to do any changes from the formal point of view.
Couldn't we change it just to keep the numbers to look like the children learn it in primary school?

opatrascoiu:
@dmntk @StrayAlien I think we should keep them as they are, as the XML standard supports the syntax. We will lose test coverage at the TCK schema level if we change them. We already have test cases where 0 is present.

My proposal is "close, no change".

dmntk:
@opatrascoiu I give up an accept the decision ;-). Where can I find the coverage tests for TCK schema? I guess the test cases are the input, but where is the generated coverage report. I am curious because I would like to improve my development process in DMNTK ;-)

PR #570, opened 2023-06-22
Change zero representation for durations

dmntk:
The string representation of zero value for days and time duration is **PT0S**.
The string representation of zero value for years and months duration is **P0M**.

So the proposal is to adjust these values accordingly in the test file.

@dmn-tck/contributors comment welcome.

### DIFF
--- a/TestCases/compliance-level-3/0100-arithmetic/0100-arithmetic-test-01.xml
+++ b/TestCases/compliance-level-3/0100-arithmetic/0100-arithmetic-test-01.xml
@@ -91,7 +91,7 @@
         <description>multiply_lhs_number_by_rhs_dtDuration_006</description>
         <resultNode name="multiply_lhs_number_by_rhs_dtDuration_006" type="decision" >
             <expected>
-                <value xsi:type="xsd:duration">P0D</value>
+                <value xsi:type="xsd:duration">PT0S</value>
             </expected>
         </resultNode>
     </testCase>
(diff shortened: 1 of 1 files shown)


StrayAlien:
Hi @dmntk - The 'canonical' string representation of all those flavours of  zero duration may be different to the actual input durations.   The duration `P0D` is a valid representation for a zero duration, but, 'by convention' , when representing it as a string, it is `PT0S`.  There are many ways to express a zero duration as input, but, we have to decide how to 'canonically' represent a zero duration when expressed as an output a string.

I say 'by convention' because (as I recall) ISO8601 gives no guidance on a canonical representation of a zero duration, so, here in the TCK, the vendors have just landed on `PT0S/P0M` - as good as any I guess.  I could be wrong re ISO8601, but, that is from memory.

The spec tells us nothing on how to represent data as strings.  So, whatever we have here is where the various vendors have landed.   To my mind, it would be a good thing to get into the spec.

My feeling is that as each of the input zero durations is valid FEEL and a valid duration, and also help exercise different runtime aspects of representing zero durations, the tests should remain as-is.

Happy for others to comment.

opatrascoiu:
Hi @dmntk According to the XML spec both values are the same

<value xsi:type="xsd:duration">P0Y</value>
<value xsi:type="xsd:duration">P0M</value>

hence, normalizing them reduces the coverage at the TCK schema level.

Hence, my proposal is "close, no change".

I am in favor of normalizing the output of the FEEL string() function though. This is an area we discussed in the past, and we should probably revisit it.

dmntk:
Hi @opatrascoiu You are right, both values are the same according to the XML spec, so why not to stick to the same value in all test cases? Are there any test examples that generate P0Y or P0M depending on different input or different FEEL expression? If so, then let us revisit them. My intention in this change was the unification across all test cases.

PR #593, opened 2023-06-24
DMN 1.4 - 1142-feel-round-down-function

StrayAlien:


opatrascoiu:
The issue with the  "Scale is in the range [−6111 ..6176]" has been clarified in RTF. The tests look good to me.

SimonRinguette:
RTF issue: https://issues.omg.org/issues/INBOX-1769

StrayAlien:
commented out test 6 as per request

SimonRinguette:
The 11b test that call using a named parameter (only one) should have also been removed with test 6 because it is the same test but with named parameter

StrayAlien:
11b commented out

PR #599, opened 2023-06-24
DMN 1.4 - 1148-feel-now-function

StrayAlien:
not a lot to test here.  We can't mock the current date/time, so, just asserting on return type. :-(


opatrascoiu:
@StrayAlien Looks good to me. Thank you!

PR #600, opened 2023-06-24
DMN 1.4 - 1148-feel-today-function

StrayAlien:
not a lot to test here, so asserting on return type.


StrayAlien:
duplicate test number.  closing. redoing.

PR #601, opened 2023-06-24
DMN 1.4 - 1150-boxed-conditional

StrayAlien:


StrayAlien:
Thanks @opatrascoiu - fair call.  I guess similar tests could be added to the other new boxed expression as well.  I was going to leave the typeRef checking as a follow on PR to the 'domain' checking PR, but, it'll not hurt to get them in here.  Thanks.

SimonRinguette:
I validated these tests

baldimir:
If we need more tests, they could be created as a separate PR to not hold these.

PR #602, opened 2023-06-24
DMN 1.4 - 1151-boxed-filter

StrayAlien:


SimonRinguette:
Test 003 should be auto-boxed since `"not a list"[true]` is valid in FEEL.

Test 006 should be auto-boxed since `1[item  >=3]` is valid in FEEL.

baldimir:
After the discussion on the TCK meeting, let's comment the tests 003 and 006 out. When we have clarification in the RTF or changes in the spec, covering the cases, we can enable them again.

SimonRinguette:

StrayAlien:
tests commented out as per request

PR #604, opened 2023-06-24
DMN 1.4 - 1152-boxed-for

StrayAlien:


opatrascoiu:
@StrayAlien I think test 007 is debatable. 

The intention is to test when the input is an empty list:  `<description>Will return false for empty list</description>`

However, decision007 contains an error, the satisfies condition is not a boolean expression:  `<satisfies><literalExpression><text>"not a boolean"</text></literalExpression></satisfies>`

I suggest replacing it with  `<satisfies><literalExpression><text>true</text></literalExpression></satisfies>`

Apart from that, the tests look good to me.

Thank you again !!

SimonRinguette:

StrayAlien:
commented out test 3/4 for undecided list coercion. Also changed test 7 to reflect @opatrascoiu  comment re 'two things wrong in the same test'

PR #605, opened 2023-06-24
DMN 1.4 - 1154-boxed-every

StrayAlien:


SimonRinguette:
@StrayAlien Thank you for the tests.

I would argue that the following tests are not right:

003 - Will return null when 'in' expression is not a list
004 - Will not coerce single value 'in' expression to list - and return null

I believe that that rule for singleton lists (When the type of the expression is T and the target type is List<T> the expression is converted to a singleton list. ) applies.

The results still return null because the satisfies does not return a boolean value.

I would suggest changing them to:

```
   <decision name="decision003" id="_decision003">
        <variable name="decision003"/>
        <every iteratorVariable="element">
            <in><literalExpression><text>true</text></literalExpression></in>
            <satisfies><literalExpression><text>element</text></literalExpression></satisfies>
        </every>
    </decision>

    <decision name="decision004" id="_decision004">
        <variable name="decision004"/>
        <every iteratorVariable="element">
            <in><literalExpression><text>1</text></literalExpression></in>
            <satisfies><literalExpression><text>element &lt;  2</text></literalExpression></satisfies>
        </every>
    </decision>
```

But these would now return true results.

SimonRinguette:

StrayAlien:
commented out test 3/4 for undecided list coercion. Also changed test 7 to reflect @opatrascoiu comment re 'two things wrong in the same test'

Issue #539, opened 2023-01-16
Clarify `null` handling in `FunctionInvocation`


nikku:
### Describe the context

The [DMN 1.4 specification](https://www.omg.org/spec/DMN/1.4/Beta1/PDF) gives users some pointers on how FEEL should handle invalid inputs and data formats (cf. [additional context](#user-content-additional-context)). From what it reads to me the way is `null` cohersion. Invalid (incompatible operations, access, or invocations) yield `null`. That intention is not explicitly expressed though (you need to read between the lines, interpret what is actually specified.

`null` cohersion / handling of non existing functions `FunctionExpression` is not clearly specified and may lead to engines behaving slightly different:

```
// given
value = non_existing_function(10)
```

### Expected Behavior

```
// then (?)
value = null
```

### Alternative Interpretations

To be clear there exists different possible interpretations:

#### strict `null` handling, blowing up

```
`non_existing_function` is not a function`
```

### Additional Context <a name="additional-context" />

I'm aware of the following cases of `null` cohersion, explicitly defined by the spec:

#### Access function with wrong arguments

```
date(1, 2, 3, 4, 5, 6) = null
```

#### Unboxing of single value lists

```
["a"] + "a" = "aa"
```

----

Related to #536

StrayAlien:
Hi @nikku - for sure the 'error handling' aspects of the spec leave it all pretty open and up to the impl.  What is certain is that invoking a non-existent function is a null.  Now, how an impl chooses to report that either at compile time (if indeed, it has a compiler) or runtime is not defined.  

Personally, I think that given the broad range of possibilities for implementors to operate as a pure 'dynamic runtime' engine with no compiler to a more advanced 'strict compiler' model would make it very hard for the spec to encompass error codes and messages and so on.

I think it is more a point of differentiation between engines how they handle errors - if (in a perfect world) we all executed in the same way, then you get choice as to how helpful a vendor's toolset is in model authoring or execution.

As a dev, I personally favour the strict compiler approach :-) ... but, others may not.

In TCK terms, a test execution is considered to have an 'error' if any error (such as a missing function), or a type null-coercion takes place.  So, loosely, a test can detect an 'error' but it is not very specific.

nikku:
> What is certain is that invoking a non-existent function is a null. 

@StrayAlien If this is certain then you answered the main question.

I'd be happy to create a TCK test covering that behavior.

```
nonExistingFunction(1) = null
```

[...]

PR #612, opened 2023-07-03
test: verify non-existing function invocation

nikku:
Yields `null`.

----

Closes https://github.com/dmn-tck/tck/issues/539


StrayAlien:
looks good to me @nikku .  Thanks.  I really thought we'd have this covered already - even accidentally - but, I don't think we do.  Good catch.  Though, just a suggestion, perhaps we could take the opportunity to examine some other simple related cases here?  

Say ...

* `null()` (returns null)
* `"somefunc"()` (returns null), where a function called `somefunc` is in scope 
* and similar to the above, invoking non functions as though they were functions.  The grammar says  function invocation is `function invocation = expression , parameters` where expression can be (well) any expression including literals.  So perhaps we could exercise a few, like `123()`, `@"2023-11-11"().  Perhaps even a double invocation like `func()()` where `func` exists in scope as a function.

Happy for comments.

nikku:
@StrayAlien I'll add the cases you mentioned.

StrayAlien:
Thanks @nikku .  I know they might seem a little (perhaps) silly, but, we're here to catch as many angles as we can.  Maybe a new runtime would not actually do `123()`, who knows really. Perhaps the mixup of literal vs identifier for the `"some_func"()` thingo might catch people, but, hey, I wish I has a € for every time some silly test saved my *ss. :-)  Easy to pass tests are 'free', but they're still tests ... 👍

nikku:
@StrayAlien There is nothing better than testing different (edge) cases. So no, testing such cases does not sound silly at all.

PR #616, opened 2023-07-20
Report Update to 1.4

SimonRinguette:
@baldimir I've updated the report to display that its on DMN 1.4 and archived the DMN 1.3 results.


baldimir:
Thank you Simon for this. Merging.

PR #618, opened 2023-08-04
Add tests for 'matches' function

opatrascoiu:


SimonRinguette:
The tests `fn-matches-32`, `fn-matches-33`, `fn-matches-34` leverage XQuery 3.0 and the DMN spec says that we should use only XQuery 1.0:

```
pattern, replacement, and flags SHALL conform to the syntax and constraints specified in clause 7.6 of XQuery 1.0 and XPath 2.0 Functions and Operators. Note that where XPath specifies an error result, FEEL specifies a null result. 
```
Your `fn-matchesErr-1` test case check for a 'p' flag that is invalid, in theory the q flag should also be invalid.

Personally, I don't mind opening a RTF issue to allow support for XQuery 3.

`K2-MatchesFunc-16`... this one my compiler does not seem to think its valid and returns null. Since its a pattern with an interpretation that is unknown, should we consider that a valid test?

Finally, the dmn model has a `K2-MatchesFunc-15` decision not referred by the test cases. This one does not compile. It has a comma between the two matches statement that seems out of place.

Otherwise, GREAT test file and would merge.

opatrascoiu:
Thanks @SimonRinguette for the comments. I will look into it in the near future.

opatrascoiu:
@SimonRinguette I committed the following changes to address the comments:
- commented out fn-matches-32, fn-matches-33, fn-matches-34 - will raise a DMN issue to upgrade to XQuery 3.0
- left fn-matchesErr-1 in as the test case as AFAIK it is valid both for XQuery 1.0 and XQuery 3.0
- commented out K2-MatchesFunc-1 as it is debatable at this stage
- added missing decision for K2-MatchesFunc-15

Please review.

PR #620, opened 2023-10-17
DMN 1.5 1155-feel-range

SimonRinguette:
Additional test coverage around ranges


StrayAlien:
@SimonRinguette - range function is 1.5, not 1.4?

SimonRinguette:
Moved to DMN 1.5 since we are using the range function in the test case

StrayAlien:
Simon, the new grammar 66 in the 1.5 beta spec seems not to permit unary ranges as string ranges.  From what I can see, (please correct me if wrong), but, all range literals but have an opening, a '..' and a closing.  In some circumstances an opening or close endpoint may be omitted, but the unary style of `< 10` is not permitted as a range literal.

AFAICT, a number of these test are using unary ranges as literals.

I kind missed this PR and have done a separate effort and suite of tests to test range literals, perhaps let's discuss at the meeting.

opatrascoiu:
There are 2 PRs that cover the range function with different file names. I suggest we drop one and include tests in the other.

SimonRinguette:
Closing this PR, you are right that there is no `<10`. I realised that when looking at Greg's PR. I had not double checked the test cases, they were prepared by someone on my team. we should use the other test cases for ranges.

PR #626, opened 2023-11-12
Invalid model.  Variable names must match the decision name.  Fixing …

StrayAlien:
just changing the names on the decision variables to match the decision names.  Otherwise, it is not a valid model.


opatrascoiu:
@StrayAlien Well spotted! I added a PR https://github.com/dmn-tck/tck/pull/627 to check these constraints during validation.

PR #630, opened 2023-11-22
Add extensions at the attribute level

opatrascoiu:


StrayAlien:
Apologies to all if I missed a conversation at a recent meeting - but, what is the objective here?  There is no description.

opatrascoiu:
@StrayAlien As you probably know, in DMN / TCK there is support for extensions at the tag and attribute level that can be used my vendors to support certain use cases. 

This PR is to add support for extensions at the attribute level in TCK.

baldimir:
Agreed on the meeting to merge this. Merging.

PR #660, opened 2024-05-15
DMN 1.5 - date increments in loops

StrayAlien:
Dates may now be used by loops.  Some simple tests for prove this is possible and that the increment is by a single day

### DIFF
--- a/TestCases/compliance-level-3/0084-feel-for-loops/0084-feel-for-loops-test-01.xml
+++ b/TestCases/compliance-level-3/0084-feel-for-loops/0084-feel-for-loops-test-01.xml
@@ -218,4 +218,110 @@
         </resultNode>
     </testCase>
 
+    <testCase id="decision_019">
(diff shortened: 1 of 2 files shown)


baldimir:

SimonRinguette:
Reviewed and test pass on the Trisotech Engine

StrayAlien:
Thanks @SimonRinguette .  Apols, I have a few more sanity check to throw in here.  Apols. On their way.  Assertions that the other date time  type are _not_ permitted.

StrayAlien:
Btw, as a small note here.  These tests did not have any actual ranges as iterations contexts.  I have added some as the spec does refer to 'ranges' of numbers and dates, but, I have a feeling the absence of actual ranges in these tests was the result of some early discussion whereby it was agreed that the spec did not permit 'range' types here.  At any rate, the tests now have actual range values.

opatrascoiu:
Looks good to me.

Issue #663, opened 2024-05-24
breaking change?:  what do do with equality range equality? What is 'undefined'?


StrayAlien:
Hi all,

We have tests in 0068-feel-quality that assert stuff like this `(<= 10) = (null..10]` is `true` for ranges.  Before 1.5, the missing endpoint on `<=10` was null.  Now it is 'undefined'.  Ironically, the word 'undefined' is not actually defined in the spec so, what it means is unknown.  It is a value?  Not sure.  It give no guidance as what happens when something is undefined.

But ... the undefined thing for ranges raises some questions.

Is `(<= 10) = (null..10]` now `false`?

 What is the value of `<=10.start` ? I think it used to be `null`, but now it is undefined.  Is accessing that `start` property now illegal?  If no ... (here we go) .... is that result of  `<=10.start` actually an now error condition and therefore returns .... `null` ...

See what I mean?  I have a feeling that perhaps `(<= 10) = (null..10]` is now supposed to be `false` ... but `<=10.start` might still return `null`.

opatrascoiu:
Not sure where the confusion is coming from. The operational semantics of unary tests and ranges are well defined in Table 55: Semantics of decision table syntax. The ones for the relational operators in Tables 52, 53, and 54.

Based on the above <=5 and (null, 5] evaluate to different results hence they are not equivalent, hence cannot be equal.

StrayAlien:
@baldimir - that was my take.  BUT ... just noting that `<=10.start` returns `null` not because the start value is actually `null`, but because accessing `start` is an error condition because its value is "undefined" (whatever that is ...).  

Furthermore to the range equality discussion we also have this: https://github.com/dmn-tck/tck/issues/667.  Whereby ranges can now be represented by (say) `=10` and `!=10`.  

But, I think I may have the answer for equality .... strictly speaking, we can go back to this definition from the spec table 53:

>  the ranges must specify the same endpoint(s) and the same comparison operator or endpoint inclusivity flag.

I think what we've (I!) have been missing here is the "the same comparison operator" thing.

I think this tells use that a range with a comparison operator could _never_ be equal to  a range without a comparison operator.  So, `<=10` cannot be equal to anything except `<=10`.  Period.

Though, as "comparison operator" is not exposed as a range property it cannot be taken into account and thus you get the following strange situation in feel:

| Var | expression | start included | start | end | end included |
| ------------| ------------- | ------------- | ------------- | ------------- | ------------- |
[...]

PR #668, opened 2024-05-24
DMN 1.5 - feel 'in' additions for '=' and '!='.  Plus ...

StrayAlien:
uncommenting null endpoint ranges tests as per: https://github.com/dmn-tck/tck/pull/393#issuecomment-1396960083.

The '=' and '!=' get exercised for each data type.

Those null range tests were commented out due to pre-1.5 ranges not permitting expressions as endpoints.  They may or may not be correct in the light of the new changes.

It is not clear to me whether `[null..10]` is the same as `<=10`, or a null endpoint should now be treated as an error condition.

### DIFF
--- a/TestCases/compliance-level-3/0072-feel-in/0072-feel-in-test-01.xml
+++ b/TestCases/compliance-level-3/0072-feel-in/0072-feel-in-test-01.xml
@@ -29,10 +29,6 @@
     list, context we just test for in scenario 001 and 011 and make sure they can be treated as endpoints in a
     list or tested for equality.
 
-    each of the range tests has 5 asserts for before/on/in/on/after endpoints
-
-    012 is actually covered by 001.  So .. so 012.
-
     -->
 
     <testCase id="number_001">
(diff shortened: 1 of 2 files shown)


opatrascoiu:
I believe some tests involving ranges with null endpoints do not follow the semantics of DMN 1.5. For example,

> null_001_a:  5 in (null..10]

According to  `10.3.2.7 Ranges` the operational semantics is delegated to the semantics of existing relational and boolean operators:

`The semantics of comparison expressions involving ranges (grammar rules 49c and 49d) is defined in Table 55, Table 54, Table 52, and Table 50.`

According to these tables

5 in (null..10] <=> 5 > null and 5 <= 10 <=> null and true <=> null (3-value logic)

According to Table 55,  condition 5 in (null..10] is not equivalent with 5 in <=10.

StrayAlien:
I have to admit ... I am now really pretty confused at to how null in ranges and the 'unary comparison' style psuedo-ranges are supposed to behave. And what is equal to what, and what 'undefined' is.  And whether attempting to get a value that is undefined actually should be an error .. and give null .. etc.

Like, we have tests in the equivalence suite asserting that `(<= 10) = (null..10]`.  In 1.4, the start range of `<=10` was null.  Now it is 'undefined' .. whatever that is.

Raised: https://github.com/dmn-tck/tck/issues/663 .. to discuss.

opatrascoiu:
@StrayAlien I checked the latest status of the PR. We are very close to merging :)

I believe the expected value for tests `null_001_a` and `null_001_b `should be `null` based on the rationale above: https://github.com/dmn-tck/tck/pull/668#issuecomment-2145284047

All the others look good to me.

SimonRinguette:
I agree with Octavian interpretation of the result of `null_001_a` and `null_001_b`. Ok with the other changes

baldimir:
We can comment out the tests we don't have agreement on and merge.

baldimir:
@StrayAlien please review.

baldimir:
"undefined" term is not properly defined in the specification. It may need more discussion in the RTF group. The current status of the spec implies meaning described in the https://github.com/dmn-tck/tck/pull/668#issuecomment-2145284047.

StrayAlien:
Tests commented @opatrascoiu  / @baldimir .  Feel free to merge if we all agree that was the blocker.

StrayAlien:
Apols - I'm confused.  I thought the request was to comment out those two tests.  They are the only two test mentioned here.  Seems maybe the request is just to make them null?  Not comment them out? I thought that was what @baldimir was requesting above.  That's what I had in my notes form the meeting.  

Also, as we've got two PRs on the same set of tests.  This PR being older than the #671 , I think we've got some wires crossed (well, we == me) when commiting.  I suggest we fix / merge this one first then I'll sort out the other.

So, just update those two tests to null is it?

StrayAlien:
@opatrascoiu .  done.

Issue #667, opened 2024-05-24
Spec permits ranges to use new `=` and `!=` comparison operators - but it doesn't make much sense (to me)


StrayAlien:
Section "10.3.2.7 Ranges"  permits ranges to be represented syntactically :

> as a comparison operator and a single endpoint (grammar rule 7.a.)

Grammar rule 7.a now includes the `=` and `!=`  operators.  Meaning that `=10` is (according to the spec) actually a range.  Same for `!=10`.

Single endpoint ranges for stuff like (say) `<=10` make sense as the start endpoint is unbounded and now 'undefined' (refer also questions in https://github.com/dmn-tck/tck/issues/663. But, an equality or inequality comparison represented as  a range would need to be something like `[10..10]` and `(10..10)` respectively.  The endpoint is not undefined and thus `=10.start` would have to be `10` .. which is not what the spec says.

So ... hope that makes sense.  

Kind of a bit confusing all around.

StrayAlien:
@opatrascoiu ... also something confusing re ranges.  Got any thoughts?

opatrascoiu:
@StrayAlien  

According to the spec both equality and inequality are comparison operators (see page 124 and the grammar) hence it makes sense to support them in the simple positive unary tests. After all, they are allowed in the comparison expressions.

I think that from the semantic point of view (basic maths) `=10` is equivalent to `>=10 and <=10` hence it is equivalent to the notation `[10..10]`. 

`!= 10` is equivalent to `<10 or >10` hence cannot be mapped to `(10..10)` which is equivalent to `>10 and <10`. I think that in this case the range properties (start, end & flags) cannot be defined.

StrayAlien:
Thanks @opatrascoiu - got to admit, I'm still not sure how to cover this.  Or even if it is possible.  The range thing is just ... well .. odd.  I guess it folds in this issue also: https://github.com/dmn-tck/tck/issues/663.   Kind of the same, so, I'll leave this ticket open for reference purposes just now, but continue on in that other issue.

gitgabrio:
@StrayAlien @opatrascoiu  @baldimir 
We are hitting the same issue.
While it would make sense to translate "=10" to a range that starts and ends at 10 (regardless of its meaning) it is pretty hard to understand what/if, should be the result of !=10.
So, any idea on how to deal with that ?

PR #671, opened 2024-07-25
DMN 1.5 - extra assertions for the range properties for `=` and `!=`.  

StrayAlien:
I have changed the structure of the tests to make then less spread out.  In effect, each range test asserts on a context of its property values.  Keeps stuff together.  All the previous test assertions are still here.

The values I am asserting on here are:

| expression | start included | start | end | end included |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| =10  | true | 10 | 10 | true |
| !=10  | false  | 10 | 10 | false |

This is not covered in the spec, so, comments welcome.

### DIFF
--- a/TestCases/compliance-level-3/0074-feel-properties/0074-feel-properties-test-01.xml
+++ b/TestCases/compliance-level-3/0074-feel-properties/0074-feel-properties-test-01.xml
@@ -407,268 +407,234 @@
         </resultNode>
     </testCase>
 
+<!--    <testCase id="range_001">-->
+<!--        <description>range - start</description>-->
+<!--        <resultNode name="range_001" type="decision">-->
+<!--            <expected>-->
+<!--                <value xsi:type="xsd:decimal">1</value>-->
+<!--            </expected>-->
+<!--        </resultNode>-->
+<!--    </testCase>-->
+
     <testCase id="range_001">
-        <description>range - start</description>
+        <description>range - [1..10]</description>
         <resultNode name="range_001" type="decision">
             <expected>
-                <value xsi:type="xsd:decimal">1</value>
+                <component name="start included">
+                    <value xsi:type="xsd:boolean">true</value>
+                </component>
+                <component name="start">
+                    <value xsi:type="xsd:decimal">1</value>
+                </component>
+                <component name="end">
+                    <value xsi:type="xsd:decimal">10</value>
+                </component>
+                <component name="end included">
+                    <value xsi:type="xsd:boolean">true</value>
+                </component>
             </expected>
         </resultNode>
     </testCase>
 
     <testCase id="range_002">
-        <description>range - end</description>
+        <description>range - (1..10]</description>
         <resultNode name="range_002" type="decision">
             <expected>
-                <value xsi:type="xsd:decimal">10</value>
+                <component name="start included">
+                    <value xsi:type="xsd:boolean">false</value>
+                </component>
+                <component name="start">
+                    <value xsi:type="xsd:decimal">1</value>
+                </component>
+                <component name="end">
+                    <value xsi:type="xsd:decimal">10</value>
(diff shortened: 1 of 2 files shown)


SimonRinguette:
@StrayAlien I could see `=10` as a range but `!=10` is two ranges: `<10 and >10`. The way you have it there `(10..10)` can't return true and does not represent `!= 10`.

StrayAlien:
Hi @SimonRinguette ... exactly, and now we're seeing some of the weirdness of it.  :-). Two ranges?

A little more here: https://github.com/dmn-tck/tck/issues/667

> FEEL supports a compact syntax for a range of values, useful in decision table test cells and elsewhere. Ranges can be syntactically represented either:
a) as a comparison operator and a single endpoint (grammar rule 7.a.)
b) .. etc

If a range can be expressed as a "comparison operator and a single endpoint", then by implication a "comparison operator and a single endpoint" _should_ be able to be expressed as a range?

opatrascoiu:
@StrayAlien The paragraph that you mentioned talks about **syntactic** representation (e.g. one object in the AST that contains the operator and the endpoint). The same section contains also references at the end to several tables that describe the **semantics** of ranges: boolean lambdas.

What the spec does not say in examples is the values of range properties for the != range (e.g. `start included` and `start`).

 I don't think it makes sense to evaluate `(!=10).start` and` (!=).end` and hence treat them as errors (e.g. return null).

StrayAlien:
Thanks @opatrascoiu - if that is the case - cool - we should get it into the spec - perhaps the `!=` comparison operator makes no sense to be described as a range.

baldimir:
From discussion on the meeting: (!=10) should be raised as a ticket to DMN OMG, asking for clarification or modification to the spec.

opatrascoiu:
@StrayAlien I suggest commenting out test range_012 until the matter is clarified by RTF.

The rest of the other tests look good to me.

SimonRinguette:
Also good with the test cases excluding ` range_012`

baldimir:
@StrayAlien could you please take a look, based on the last comments from Simon and Octavian?

baldimir:
We can comment out the tests we don't have agreement on and merge.

StrayAlien:
Commented out range_012.  @baldimir  / @opatrascoiu .. it this was the blocker, feel free to merge.

opatrascoiu:
I believe that the comment https://github.com/dmn-tck/tck/pull/671#issuecomment-2489610645 referred to removing

    <testCase id="range_012">
        <description>range - (!=10)</description>
        <resultNode name="range_012" type="decision">
            <expected>
                <component name="start included">
                    <value xsi:type="xsd:boolean">false</value>
                </component>
                <component name="start">
                    <value xsi:type="xsd:decimal">10</value>
                </component>
                <component name="end">
                    <value xsi:type="xsd:decimal">10</value>
                </component>
                <component name="end included">
                    <value xsi:type="xsd:boolean">false</value>
                </component>
            </expected>
        </resultNode>
    </testCase>

from the first commit in your PR. The second commit of your PR reverts the test approach (without components) and removes

    <testCase id="range_012">
        <description>range - start: GE unary syntax</description>
        <resultNode name="range_012" type="decision">
            <expected>
                <value xsi:type="xsd:decimal">10</value>
            </expected>
        </resultNode>
    </testCase>

I prefer the testing approach with components - it makes the test coverage more readable. 

I suggest reverting the second commit in your PR and comment out test range_012 with components.

StrayAlien:
hey @opatrascoiu  - yeah - my fault - apologies - there is two different PRs on the same test suite and this one is later with the 'context' thing..  Looks like I got some wires crossed.  I suggest we sort the other one, then come back to this.

StrayAlien:
okay, reverted then re-applied.  Hopefully okay now.

PR #674, opened 2024-08-06
DMN 15b - no-name imports

StrayAlien:
Hold onto your hats.  Tests for the new "no-name" import feature of DMN 1.5b.

Import tests can be very confusing, so, I have broken these up into lots of small models with an associated test suite comprising of (generally) just one assertion.

Here is the new spec text:

> Multiple imports with empty import names are allowed in the default namespace and their precedence is resolved according to their definition order. When the import name attribute is an empty string, the elements are imported in the default namespace of the model. 

> When a name collision occurs between an element in the default namespace and an imported element, the imported element does not replace the one already in the default namespace while the elements without name collision are imported.

Some things to note:

* "name collision" is not just _any_ name - itemDefinitions/imports can share names with DRG elements.  That has been exercised here.
* I think this should also be "id collision" as ids should be unique within a model AFAIK.  So, if an imported model has an id that conflicts with the importing model then it should also not be imported (IMO).  I wanted to test this, but, in strictness, the spec does not mention it.  I think we should test it ... somehow.
* some test models have "dependency loops" in that a model may directly or indirectly import its own namespace.  We can really test it, but, we can put it in tests as a sanity check for runtimes.
* tests here assert stuff about direct imports as well as imports of imports.

Each test suite has some small notes in it describing the model structure and expectations.

Comments welcome.

### DIFF
--- a/TestCases/compliance-level-3/1158-noname-imports/1158-noname-imports-001-A.dmn
+++ b/TestCases/compliance-level-3/1158-noname-imports/1158-noname-imports-001-A.dmn
@@ -1,0 +1,59 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<definitions namespace="http://www.montera.com.au/spec/DMN/1158-noname-imports-001-A"
+             name="1158-noname-imports-001-A"
+             id="_i9fboPUUEeesLuP4RHs4vA"
+             xmlns="https://www.omg.org/spec/DMN/20230324/MODEL/"
+>
+    <import namespace="http://www.montera.com.au/spec/DMN/1158-noname-imports-001-B"
+            name=""
+            importType="https://www.omg.org/spec/DMN/20230324/MODEL/"
+    />
+
+    <!-- references an informationRequirement from an import  -->
+    <decision name="decision001" id="_decision001">
+        <variable name="decision001"/>
+        <informationRequirement>
+            <requiredDecision href="#_model_b_decision001"/>
+        </informationRequirement>
+        <literalExpression>
+            <text>model_b_decision001</text>
+        </literalExpression>
+    </decision>
+
+    <!-- references an knowledgeRequirement from an import -->
+    <decision name="decision002" id="_decision002">
+        <variable name="decision002"/>
+        <knowledgeRequirement>
+            <requiredKnowledge href="#_model_b_bkm001"/>
+        </knowledgeRequirement>
+        <literalExpression>
+            <text>model_b_bkm001()</text>
+        </literalExpression>
+    </decision>
+
+    <!-- references an inputData from an import -->
+    <decision name="decision003" id="_decision003">
+        <variable name="decision003"/>
+        <informationRequirement>
+            <requiredInput href="#_model_b_input001"/>
+        </informationRequirement>
+        <literalExpression>
+            <text>model_b_input001</text>
+        </literalExpression>
+    </decision>
+
+    <!-- uses an imported "string" typeRef - two tests will provide a string and a number
+    as input to assert the correct type is being used (we'll see a null for a number value) -->
+    <decision name="decision004" id="_decision004">
+        <variable name="decision004" typeRef="typeRefA"/>
+        <informationRequirement>
+            <requiredInput href="#_model_b_input001"/>
+        </informationRequirement>
+        <literalExpression>
+            <text>model_b_input001</text>
+        </literalExpression>
+    </decision>
+
+
+</definitions>
+
--- a/TestCases/compliance-level-3/1158-noname-imports/1158-noname-imports-001-B.dmn
+++ b/TestCases/compliance-level-3/1158-noname-imports/1158-noname-imports-001-B.dmn
@@ -1,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<definitions namespace="http://www.montera.com.au/spec/DMN/1158-noname-imports-001-B"
+             name="1158-noname-imports-001-B"
+             id="_i9fboPUUEeesLuP4RHs4vA"
+             xmlns="https://www.omg.org/spec/DMN/20230324/MODEL/"
+>
+    <itemDefinition name="typeRefA">
+        <typeRef>string</typeRef>
+    </itemDefinition>
+
+    <inputData name="model_b_input001" id="_model_b_input001">
+        <variable name="model_b_input001" typeRef="Any"/>
+    </inputData>
+
+    <!-- decision just returns its own name -->
+    <decision name="model_b_decision001" id="_model_b_decision001">
+        <variable name="model_b_decision001"/>
+        <literalExpression>
+            <text>"model_b_decision001"</text>
+        </literalExpression>
+    </decision>
+
+    <!-- BKM just returns its own name -->
+    <businessKnowledgeModel name="model_b_bkm001" id="_model_b_bkm001">
+        <variable name="model_b_bkm001"/>
+        <encapsulatedLogic>
+            <literalExpression>
+                <text>"model_b_bkm001"</text>
+            </literalExpression>
+        </encapsulatedLogic>
+    </businessKnowledgeModel>
+
+
+</definitions>
+
--- a/TestCases/compliance-level-3/1158-noname-imports/1158-noname-imports-001-test.xml
+++ b/TestCases/compliance-level-3/1158-noname-imports/1158-noname-imports-001-test.xml
@@ -1,0 +1,68 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<!-- Contributed to DMN TCK by StrayAlien -->
+<testCases xmlns="http://www.omg.org/spec/DMN/20160719/testcase"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation=""
+           xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+    <modelName>1158-noname-imports-001-A.dmn</modelName>
+
+    <!-- model A uses a number of no-name imports from model B -->
+
+    <testCase id="001">
+        <description>will import a decision</description>
+        <resultNode name="decision001" type="decision">
+            <expected>
+                <value xsi:type="xsd:string">model_b_decision001</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="002">
+        <description>will import a BKM</description>
+        <resultNode name="decision002" type="decision">
+            <expected>
+                <value xsi:type="xsd:string">model_b_bkm001</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="003">
+        <description>will import an inputData</description>
+        <inputNode name="model_b_input001">
+            <value xsi:type="xsd:string">model_b_input001</value>
+        </inputNode>
+        <resultNode name="decision003" type="decision">
+            <expected>
+                <value xsi:type="xsd:string">model_b_input001</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="004">
+        <!-- imported typeRef is a string, we test a string against it -->
+        <description>will import a typeRef</description>
+        <inputNode name="model_b_input001">
+            <value xsi:type="xsd:string">a string</value>
+        </inputNode>
+        <resultNode name="decision004" type="decision">
+            <expected>
+                <value xsi:type="xsd:string">a string</value>
+            </expected>
+        </resultNode>
+    </testCase>
+
+    <testCase id="004_a">
+        <!-- imported typeRef is a string, we test a number against it -->
+        <description>will import a typeRef</description>
+        <inputNode name="model_b_input001">
+            <value xsi:type="xsd:decimal">1234</value>
+        </inputNode>
+        <resultNode name="decision004" type="decision">
+            <expected>
+                <value xsi:nil="true"/>
+            </expected>
+        </resultNode>
+    </testCase>
+
+
+</testCases>
--- a/TestCases/compliance-level-3/1158-noname-imports/1158-noname-imports-002-A.dmn
+++ b/TestCases/compliance-level-3/1158-noname-imports/1158-noname-imports-002-A.dmn
@@ -1,0 +1,59 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<definitions namespace="http://www.montera.com.au/spec/DMN/1158-noname-imports-002-A"
+             name="1158-noname-imports-002-A"
+             id="_i9fboPUUEeesLuP4RHs4vA"
+             xmlns="https://www.omg.org/spec/DMN/20230324/MODEL/"
+>
+    <import namespace="http://www.montera.com.au/spec/DMN/1158-noname-imports-002-B"
+            name=""
+            importType="https://www.omg.org/spec/DMN/20230324/MODEL/"
+    />
+
+    <!-- references an informationRequirement from an import -->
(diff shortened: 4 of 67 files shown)


gitgabrio:
Hi @StrayAlien 
Thanks for the PR.
I tried to test it locally, but for some reason no "1158" get executed.
At the same time, there seems to be some issue in the validation.
Could you please check or provide suggestion ? Thanks!

StrayAlien:
HI @gitgabrio .. hmm, that could have something to do with the test runner - whichever one it is you're using (Java?). I don't use the Java one.  My understanding was that the Java test runner would run each test xml in a folder and use the 'modelName' referenced in the xml  as the root model for tests.  That root model may import other models in the same folder.

I've just checked the TCK and, how about that, this PR has the first test folder to contain more than one xml test file.  It has sometimes been spoken about to write tests like this, but, perhaps the Java test runner is not respecting multiple test files in a folder.

gitgabrio:
Hi @StrayAlien 
not sure, anyway, my doubts are the following:
1. the **TCK Validation** job fails - and IINW this should not happen
2. IINW, in all other tests the file name format is
`*-test-(\\d\\d).xml`
but in this folder it is
`*-(\\d\\d\\d)-test.xml`

I'm not sure if there is a "rule" for that

@baldimir wdyt ?

falko:
@saig0 FYI as we just talked about imports yesterday.

StrayAlien:
I can change the test files name. Though, that seems a pretty odd restriction.  It could be changed to something like "\\d+" and be more flexible and backwards compatible ....

gitgabrio:
@StrayAlien 
I personally have no strong opinion, but IINW until now everything here worked on that "convention", on top of which runners have been implemented.
I don't know if that convention has been discussed in the past (TBH, I guess it had, since otherwise there would not have been any reason to implement such filter 🤔 ), but if we want to change it, maybe we have to make an explicit agreement.
Beside that, there is also the TCK Validation job failure, that, again, IMO is a smell of some other problem
(please keep in mind I am very new to this environment, so I may be completely wrong)

StrayAlien:
@gitgabrio - test names have been changed.

gitgabrio:
@baldimir @yesamer 
Could you please double check ?

yesamer:
It seems the tests are still not consumed by Java-based engines, at least in my local.

StrayAlien:
Any more information that that?  I don't use Java for DMN - fancy debugging it?

StrayAlien:
I just tried the drools runner on TCK master - I assume that is the one your using (?).  I get 121 errors.  I see the import tests are failing.  

I can run the validation profile locally and it works.

Here are some sample drools runner errors:

```
[ERROR]   [Message [id=1, kieBase=defaultKieBase, level=ERROR, path=/Users/gregm/workspace/personal/dmn/tck-fork/runners/dmn-tck-runner-drools/../../TestCases/compliance-level-3/0086-import/0086-import.dmn, line=0, column=0
   text=DMN: Import type unknown: 'https://www.omg.org/spec/DMN/20230324/MODEL/'. (Invalid FEEL syntax on the referenced expression) ], Message [id=2, kieBase=defaultKieBase, level=ERROR, path=/Users/gregm/workspace/personal/dmn/tck-fork/runners/dmn-tck-runner-drools/../../TestCases/compliance-level-3/0086-import/0086-import.dmn, line=20, column=-1
   text=DMN: Unable to resolve type reference 'myimport.tPerson' on node 'A Person' (DMN id: _9e1f6cbc-584f-41f6-8748-97f579a3df43, The listed type definition was not found) ], Message [id=3, kieBase=defaultKieBase, level=ERROR, path=/Users/gregm/workspace/personal/dmn/tck-fork/runners/dmn-tck-runner-drools/../../TestCases/compliance-level-3/0086-import/0086-import.dmn, line=28, column=-1
   text=DMN: Required Business Knowledge Model 'http://www.trisotech.com/definitions/_f27bb64b-6fc7-4e1f-9848-11ba35e0df36#_32543811-b499-4608-b784-6c6f294b1c58' not found on node 'A Decision Ctx with DT' (The referenced node was not found) ], Message [id=4, kieBase=defaultKieBase, level=ERROR, path=/Users/gregm/workspace/personal/dmn/tck-fork/runners/dmn-tck-runner-drools/../../TestCases/compliance-level-3/0086-import/0086-import.dmn, line=33, column=-1
   text=DMN: Error compiling FEEL expression 'myimport.Say Hello(A Person)' for name 'normal greeting' on node 'A Decision Ctx with DT': Unknown variable 'myimport' (DMN id: _c6977a3f-f3e0-411f-a1fc-590db1b97958, Error compiling the referenced FEEL expression) ]]
```

As an FYI, I think the way others may have done the imports functionality with a test runner is kind of like this:

* each test xml file in a folder is a test suite
* the test suite file references a "modelName" DMN file - which is the "entry point" model
* the entry point DMN is in the same folder
* every DMN file in the same folder could be imported directly or indirectly by the entry point model by namespace.  That is, if a model has an "imports" element, the import namespace will be the namespace of one of the DMN files in the same folder.

That's it I think.  I can see the Java DmnTckRunner class does collect all dmn files in the same folder as the test suite and passes them to the the "beforeTestCases" function of the vendor runner impl.  The drools runner takes those and adds them as extra resources to the drools DMN runtime .... 

Given that it seems to fail on master ... perhaps a drools crew member could could check it out?

baldimir:
I guess the problem is that the common runner, that all the runners can extend, uses this pattern to match files. https://github.com/dmn-tck/tck/blob/7c85cec528fef2f82b26a846e234c398060d05fe/runners/dmn-tck-runner/src/main/java/org/omg/dmn/tck/runner/junit4/DmnTckSuite.java#L75

So either you doublecheck, that test xml files in this PR satisfy this pattern, please, or I could change that filter, if we all agree, to just take all xml files from a test folder. What do you think please? @dmn-tck/contributors

StrayAlien:
Hi @baldimir - just pushed another bounce at renaming things.  Hopefully that works now.

opatrascoiu:
@StrayAlien Thank you for the test.

I believe there is an issue with the `hrefs `when DRG elements are imported from a different file. According to the spec the namespace is mandatory, regardless if the import prefix is empty or not (see 13.3.2 References within the DMN XSD).

Here is an example of an incorrect href https://github.com/StrayAlien/tck/blob/c7a65b8af372cb92cd37a4688d4ad5d0a67fae37/TestCases/compliance-level-3/1158-noname-imports/1158-noname-imports-01-A.dmn#L16

gitgabrio:
@opatrascoiu 
If you are referring to a noname import
(6.3.3 Import metamodel)

> When the import name attribute is an empty string, the elements are imported in the default namespace of the
model. 

"nonamed import" should have the same namespace as the importing model itself, and then I have the impression the "namespace" will be redundant (since it is the same of the importing model).
At a slightly more abstract layer, I always had the impression that the idea is to consider the imported model "exactly as" if defined inside the importing one (as much as possible), and so href with namespace seems to contradict that.

All the above, in my opinion, overrule the 12.3.2 References within the DMN XSD you pointed at.

Please correct me if I'm misunderstanding your observation or missing something else of the context

@baldimir @yesamer

opatrascoiu:
@gitgabrio I disagree. The paragraph that you mentioned does not say explicitly that the syntactic structure of the hrefs should change. It defines the operational semantics of the imports when no prefix is needed in the feel expressions. 

The section that I mentioned is about the syntax of the hrefs.

gitgabrio:
@opatrascoiu 
IINW, when an hrefs refers to element of the same model/file, the namespace is not required (since it is the same as the model itself)
Then, the noname import define that nonamed imports should have same namespace as the importing model.
It seems we had a general agreement that the idea behind noname import is what I mention above (treat noname import elements as if written in the importing model itself)
And it seems that the href spec simply was not updated to explicitly deal with that situation ("_if an hrer refers to an elment in a different file but with a no-name import....._").
Beside, following your approach we would have that:
1. some href does not have the namespace of the model where they are
2. some href have the same namespace of the model in which they are (because they comes from noname imports)
3. other elements does not requires explicit reference to imported model because- being a noname import - are treated as if it is in the same model.

So, I think that in this case it should be better to apply a consistent interpretation, since spec - as whole - sometime are not very well described or "coherent"

opatrascoiu:
@gitgabrio 

DMN borrowed the reference mechanism via HREFs from XML. They refer to the **physical location** of the referred element. 

The physical location (in a file) and adding the definition via import in the current DMN context are two different things to me. 

```
When this Decision is referenced, e.g., by an InformationRequirement in a Decision that is defined in another file, the reference could take the following form:
<requiredDecision href=”http://www.example.org/Definitions01.xml#prebureauriskDec01”/> where
220 Decision Model and Notation (DMN), v1.6 Beta 1
“http://www.example.org/Definitions01.xml” is an URI reference to the XML document in which the “PreBureau Risk Category” Decision is defined (e.g., the value of the locationURI attribute in the corresponding Import element), and “prebureauriskDec01” is the value of the id attribute for the Decision.

```

StrayAlien:
Hi @opatrascoiu / @gitgabrio .  Thanks for the comments.  Very interesting.  It's great there is some discussion on these.

Having those imported hrefs as 'local' is intentional.  According to the spec:

> When the import name attribute is an empty string, the elements are imported in the default namespace of the model.

IMO, they become available under the default namespace of the imported model.  In a sense, the existence of the import namespace is 'erased'.  That also means the those imported inputs are available in the default namespace. 

If I were calling this model via an API I would not expect to have to qualify that input data name with a namespace - because, it has become 'part' of the importing model.

That's my take on it.  If it is not the case, then I'm not too sure what " imported in the default namespace of the model" might mean in practice.

So, that TCK test passing an imported input as if it was part of the importing model is intentional and there to assert same.

opatrascoiu:
@StrayAlien 

Regarding the paragraph mentioned here https://github.com/dmn-tck/tck/pull/674#issuecomment-2423240850

The HREFs contain the physical location (e.g. file that contains the definition) not the namespace of the model described in that file. These are different things. DMN 1.5. states the physical location should always be present if the DRG element is not described in the same file.

gitgabrio:
@opatrascoiu 
Your original comment was

> I believe there is an issue with the hrefs when DRG elements are imported from a different file. According to the spec the namespace is mandatory, regardless if the import prefix is empty or not (see 13.3.2 References within the DMN XSD).

But with that:

> The HREFs contain the physical location (e.g. file that contains the definition) not the namespace of the model described in that file.

it seems you were talking of physical location.
If a real, physical location is required, then all TCK models using href with physical path ("http...") are broken, because they refer the "namespace" and not an actually existing location (and this is true for lot of other models).

But that "References within the DMN XSD" paragraph has not been modified since 1.4, and the need of physical path seems to derive from

> however it can only reference an element within the same file.

that was true then, but not with no-named import.

As @StrayAlien mentioned,  it seems 

> the intention here is really the equivalent of an old-fashioned "include" 

If you think there is a _real_ need, beside the specs that maybe are simply outdated, could you please explain ? Thanks

References

https://datatracker.ietf.org/doc/html/rfc3986
> Uniform Resource Identifier (URI) is a compact sequence of
   characters that identifies an abstract or physical resource.

opatrascoiu:
@gitgabrio I checked all the DMN files in the master of the TCK project. The only ones that use the namespace instead of location (see 12.3.2 References within the DMN XSD) are the ones in 0086-import and the DMN files in this PR. All the other test cases have only one DMN model hence the HREFs don't contain a location.

We need to discuss in the next meeting how are we going to move forward on the references, both in DMN files and TCK files (references of DRG elements and inputNode and resultNode).

StrayAlien:
This spec change was driven by Trisotech?  Perhaps @SimonRinguette could chime in?

opatrascoiu:
Just to add some context here. 

A while back @StrayAlien added a test where the namespace is always specified in href, including the DRG elements defined in the same file. 

This use case was addressed in DMN 1.6 https://issues.omg.org/browse/DMN16-64: the namespace for hrefs to DRG elements defined in the same DMN is optional, is not an error to have it.

baldimir:
Based on the discussion on the meeting: 
- The main difference when referencing with href is that href needs to contain a physical location and not namespace. 
- We have tests in TCK repository that use namespace to reference also with hrefs. E.g. here (1). We may need to review such tests. 
- The no-name import may need to be clarified or discussied in RTF meetings. Looks like the intention of no-name imports is to import only to the default namespace in the DMN context, not in the XML contex, so hrefs shouldn't be affected by this. 

(1) https://github.com/dmn-tck/tck/blob/a967d577aafa2dd0e456084bf964d9e8fb6da0b5/TestCases/compliance-level-3/0091-local-hrefs/0091-local-hrefs.dmn

baldimir:
Similar FEEL context: 10.3.2.11 Scope and context stack.

opatrascoiu:
I believe that the ambiguity in this thread is caused  by 

> When the import name attribute is an empty string, the elements are imported in the default namespace of the
> model.

Following our discussion in the meeting, I believe this means

> When the import name attribute is an empty string, the elements are imported directly in the current DMN _context_ (see 7.1 Introduction metamodel and 10.3.2.11 Scope and context stack).

gitgabrio:
I think we all agree on ambiguity of explanation, but any attempt to disambiguate it, seems to me just a matter of personal opinion.
Saying that "namespace" has been used instead of "context" is a possible solution, but it also seems strange to me, since "namespace" and "context" have completely different semantics.
Just to reinforce, I'm not saying any of the given interpretation is right or wrong: I'm simply saying that with the information we currently have, it is not possible (IMO) to clearly infer the behavior.

gitgabrio:
@opatrascoiu 
What concern me is mostly the kind of confusion this kind of interpretation is going to create, from user POV.
I mean, on one side, for some elements, the "same namespace rule" apply - i.e. as in the same model. In other cases, the different file rule (i.e. reuiqred namespace) apply. 
In my experience, we, as developer, are able to keep this distinction in minds (but not even always), but from user POV, that often are not so "technical" it only is a useless nuisance.
There is not any real technical need for that:
1. that "namespace" in href does not point to a real file, but to a label
2. the "imported" file has to be solved anyway during compilation, to validate all the conditions where the imported namespace can be avoided.
So, it is just our own "convention", and probably when the propose has been made no-one thought of such consequence.
So, the ambiguity will be hard-coded in the model semantic itself, IMO

opatrascoiu:
@StrayAlien I looked at tests  1158-noname-imports-test-14.xml and 1158-noname-imports-test-15.xml. I like the intention behind them!

However, I disagree with the expected result. IMO it should be:

     <testCase id="001">
        <description>requirement of imported decision will be resolved in the context of the imported decision</description>
        <resultNode name="decision001" type="decision">
            <expected>
                <value xsi:type="xsd:string">model_b_decision003</value>
            </expected>
        </resultNode>
    </testCase>

and

    <testCase id="001">
        <description>fully qualified local requirement href of imported decision will be resolved in the context of the imported decision</description>
        <resultNode name="decision001" type="decision">
            <expected>
                <value xsi:type="xsd:string">model_b_decision003</value>
            </expected>
        </resultNode>
    </testCase>


I think that according to sections `13.3.2 References within the DMN XSD` and  `10.3.2.11 Scope and context stack` the hrefs are pointing to the DRG elements in the file of the imported decision. 

Once we address the issue with the hrefs in the DNN files and https://github.com/dmn-tck/tck/discussions/686 we are good to merge :)

baldimir:
We need to review based on the latest agreement from RTF. Based on the discussion in RTF, the href reference contains namespace, not location. It may be possible the namespace is missing in some test cases in the hrefs. This needs to be worked on.

opatrascoiu:
This PR is replaced by https://github.com/dmn-tck/tck/pull/692 with fixes agreed as part of the discussions in RTF and this group.

PR #677, opened 2024-10-09
Add IBM BAMOE 9.1.1 results. 

baldimir:
- This PR adds results for a new vendor - IBM. IBM BAMOE contains a downstream build of Drools, therefore I used the Drools runner to produce the results and there is not a separate runner added. 
- This PR also refactors and updates the Drools runner a bit. It was not properly maintained for some time, so I tried to fix as many "code smells" as possible. I also encapsulated some parts of the runner to separate classes, so it is not just one big class. 
- I also updated the Drools version to 10.0.999-SNAPSHOT as the previous Drools version is more than a year old. The Drools 10 final version will be hopefully released soon in the Apache KIE community. It is already being voted on. When the 10 release is out, I will update the Drools runner to the final version. The current snapshot should provide us with better results for Drools in the optional Drools PR check. 
- This PR also updates the Drools runner to point to Apache as a vendor, instead of Red Hat. I discussed this with @tarilabs, he is ok with this change. So for the future, Red Hat in results will be updated to Apache, when there are the first Drools 10 final results.  

The commits should be squashed when merging.


runners/dmn-tck-runner-drools/src/test/java/org/omg/dmn/tck/runner/drools/DroolsTCKTest.java
```diff
@@ -516,7 +320,7 @@ private Object parseType(ValueType value, DMNType dmnType) {
                                 } catch (DateTimeParseException e) {
                                     return ComparablePeriod.parse(dateTimeOrDurationValue.toString());
                                 }
-                            } else if (dateTimeOrDurationValue instanceof XMLGregorianCalendar) {
```
gitgabrio, line 519: Does this change the behavior, somehow ? 
Why the ` if (dateTimeOrDurationValue instanceof XMLGregorianCalendar)` is not needed anymore ?

runners/dmn-tck-runner-drools/src/test/java/org/omg/dmn/tck/runner/drools/DroolsTCKTest.java
```diff
@@ -530,25 +334,23 @@ private Object parseType(ValueType value, DMNType dmnType) {
                                     return LocalTime.parse(xmlCal.toXMLFormat());
                                 }
                                 return xmlCal.toGregorianCalendar();
-                            } else {
```
gitgabrio, line 533: see https://github.com/dmn-tck/tck/pull/677/files#r1793266442

runners/dmn-tck-runner-drools/src/test/java/org/omg/dmn/tck/runner/drools/DroolsTCKTest.java
```diff
@@ -530,25 +334,23 @@ private Object parseType(ValueType value, DMNType dmnType) {
...
-                                return dateTimeOrDurationValue;
                             }
                         }
                     }
                 } catch (Exception e) {
-                    logger.error("Error trying to coerce JAXB type " + val.getClass().getName() + " with value '" + val.toString() + "': " + e.getMessage());
+                    LOGGER.error("Error trying to coerce JAXB type {} with value '{}': {}", val.getClass().getName(), val, e.getMessage());
```
gitgabrio, line 341: IINW, at line 390
`Object dateTimeOrDurationValue = (val != null) ? val.getValue() : null;`
`val` could be `null`
but here it is used for message creation, potentially throwing a NPE during error logging.

runners/dmn-tck-runner-drools/src/test/java/org/omg/dmn/tck/runner/drools/DroolsTCKTest.java
baldimir: The instanceof check is not needed, as there is this or here https://github.com/dmn-tck/tck/pull/677/files#diff-9b2a1d8823619346c392f6da61429acdbb6a4fbf31673bcba671639a492fa564R310

runners/dmn-tck-runner-drools/src/test/java/org/omg/dmn/tck/runner/drools/DroolsTCKTest.java
baldimir: Actually, I found out the whole try-catch block was unnecessary there. The whole code wrapped doesn't throw any checked exception. I personally don't like catching exceptions with the general Exception class, so I removed it.

yesamer (approved):
Welcome, BAMOE! :D__

tarilabs:
Congratulations for having fresh results of the Drools open source project in the TCK !

baldimir:
@tarilabs not yet :) This is only adding for IBM BAMOE. I will update Drools results for when 10 is released. However I wanted to have the runner ready for that, so I will update as soon as possible.

baldimir:
Looks like I need to do some fixes for the Drools PR check now. I am working on it. I will push updates soon.

baldimir:
Done.

gitgabrio (approved):

Issue #681, opened 2024-10-30
For loop tests in 0084-feel-for-loops.dmn


opatrascoiu:
We have recently merged a PR with new tests in the above file. 

I believe that some of the tests are incorrect. For example,

>     <decision name="decision_023" id="_decision_023">
>         
>         <variable name="decision_023"/>
>         <literalExpression>
>             <text>for i in [1..2] return i</text>
>         </literalExpression>
>     </decision>
> 
>     <decision name="decision_024" id="_decision_024">
>         
>         <variable name="decision_024"/>
>         <literalExpression>
>             <text>for i in [@"1980-01-01"..@"1980-01-03"] return i</text>
>         </literalExpression>
>     </decision>
> 

According to the spec (10.3.2.14 For loop expression):

> An iteration context may either be an expression that returns a list of elements, or two expressions that return integers connected by “..”.

The above tests are incorrect (the expected value should be null) as the for loops have an iteration context made of one single expression, that is not a list, is a range. The correct syntax is 1..2 or @"1980-01-01"..@"1980-01-03" (see examples in the spec). 

I am in favor of extending the spec to support the above, as they make sense, but we need to specify the semantics for open ends  (e.g. (1, 2] ).

StrayAlien:
Hi @opatrascoiu - I was basing it on the spec text:

> The for loop expression iterates over lists of elements or ranges of numbers or dates

I guess "range" is ambiguous.  The tests following those do some checking for what might be invalid ranges (like the start being lesser than the end - and so on) but not open ended.

If they're contentious, I'm happy to comment them for now.  Btw, if "proper" ranges are to be included there, then modifications to the spec should also consider <=10, !=10,=10 etc comparitive ranges as well.  Like to say they're not permitted.

Also, worth noting is that spec text you referenced seems not to have been updated when dates where added in 1.5.

opatrascoiu:
> Hi @opatrascoiu - I was basing it on the spec text:
> 
> > The for loop expression iterates over lists of elements or ranges of numbers or dates
> 

@StrayAlien Yes, the `range` concept introduced in the paragraph that you mentioned (section 10.3.2.14 For loop expression) is ambiguous. IMO the rest of the section clarifies the meaning of range: 1..2 and not [1..2]. There are also some examples that support this approach.

I suggest we remove the tests and raise an RTF ticket to clarify.

StrayAlien:
[...]

PR #684, opened 2024-11-25
Change 'for loop' tests to match semantics in Section 10.3.2.14

opatrascoiu:
This is related to https://github.com/dmn-tck/tck/issues/681

According to Section 10.3.2.14, range expressions are not allowed as iteration context. For example, 1..2 is supported but not (1..2]

> An iteration context may either be an expression that returns a list of elements, or two expressions that return integers connected by “..”.

### DIFF
--- a/TestCases/compliance-level-3/0084-feel-for-loops/0084-feel-for-loops-test-01.xml
+++ b/TestCases/compliance-level-3/0084-feel-for-loops/0084-feel-for-loops-test-01.xml
@@ -291,26 +291,19 @@
     </testCase>
 
     <testCase id="decision_023">
-        <description>a valid numeric range is permitted as iteration context</description>
+        <description>numeric range is not permitted as iteration domain</description>
         <resultNode name="decision_023" type="decision">
             <expected>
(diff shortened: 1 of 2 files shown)


baldimir:
I am personally fine with the changes. What about others please? @dmn-tck/contributors

falko:
Didn't we have an RTF issue around this?

opatrascoiu:
@falko I do not remember having an RTF issue about loops over ranges such as [1..3). There is one in DMN 1.5 about loops over dates but using the syntax 1..3 (without parenthesis).

baldimir:
@SimonRinguette please check.

baldimir:
Notes from the meeting: 
- Based on discussion in the RTF group, this PR should be fine in the limits of DMN 1.5. 
- It may be interesting to support ranges as an interation context, however that would need clarifications in the specification, e.g. related to the boundaries, e.g. when there is infinite boundary etc. 

opatrascoiu:
Agreed on: Comment out the test cases and wait for the RTF outcome.

PR #690, opened 2025-02-06
Change model names to match file names for consistency

opatrascoiu:


baldimir:
This is a good idea. It may need clarification in the readmes/docs so contributors follow this convention.

PR #692, opened 2025-03-25
Add tests for imports with no prefix

opatrascoiu:
This PR contains fixes for https://github.com/dmn-tck/tck/pull/674 as per discussion in the DMN RTF:
- hrefs contain DMN namespaces not locations
- href must contain a namespace when the imported DRG element is not  defined in the current DMN file
-

PR #693, opened 2025-03-25
Add tests for imports with prefix

opatrascoiu:
Test cases are similar with the ones in https://github.com/dmn-tck/tck/pull/692 but with a prefix


baldimir:
@dmn-tck/contributors please review. This is an important topic, good to be merged as soon as possible so we can move focus on 1.6.

PR #694, opened 2025-03-25
Add tests for list boxed expressions

opatrascoiu:


opatrascoiu:
@SimonRinguette You were right about the expected values for string(null). I made the changes.

Please review

opatrascoiu:
@StrayAlien Well spotted! The editor that I used has some issues. I have fixed the typeRefs.

PR #696, opened 2025-03-28
Single parameter number function-TCK test cases

ChinchuAjith:
Closes https://github.com/apache/incubator-kie-issues/issues/1889


opatrascoiu:
Thank you for contributing.

Please move the test cases in 0058-feel-number-function anc check that there are no duplicates. Better if we have all the tests for number() in one place.

PR #702, opened 2025-06-04
[DMN 1.6] Change test case to match DMN16-50

opatrascoiu:
The semantics of bindings for multiple imports has been clarified:

`An InformationItem element contained in an InputData is assigned a value by an external data source that is attached at runtime. When an InputData is imported several times via transitive imports, the contained InformationItem is assigned only once and holds the same value.`


baldimir:
This is fine from my side. @dmn-tck/contributors please review.

PR #726, opened 2026-02-19
Add new name attribute for TestCases

opatrascoiu:
We have a use case where the TestCases need to have a name, it cannot be inferred from the name of the TCK file.


SimonRinguette (approved):
I'm good with this proposed change

yesamer (approved):

PR #729, opened 2026-02-27
Add new test cases for decision services

opatrascoiu:
The purpose of the PR is:
- add test cases for DSs with multiple output decisions
- add test cases for DSs expressed in a much simpler way, without using the invocableName and type. IMO it is more user-friendly.


SimonRinguette (changes requested):
I'm good with the new 015 test case.

However the 001a and 015a variant without the invocableName/type do not work for us because I'm expecting those values to invoke the decision service. We would need to modify (and I believe other vendors as well) their test runner to support omitting these values.

I would highly prefer not going in that direction (without the invocableName/type)

yesamer (approved):

SimonRinguette (approved):

PR #738, opened 2026-05-19
Add test cases for unary tests in decision tables

opatrascoiu:
As discussed in the last meeting, this PR replaces https://github.com/dmn-tck/tck/pull/672 without constraints. For constraints is going to be a new one.

### DIFF
--- a/TestCases/compliance-level-3/1163-unarytests-decisiontable/1163-unarytests-decisiontable-test-01.xml
+++ b/TestCases/compliance-level-3/1163-unarytests-decisiontable/1163-unarytests-decisiontable-test-01.xml
@@ -1,0 +1,172 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<!-- Contributed to DMN TCK by StrayAlien -->
+<testCases xmlns="http://www.omg.org/spec/DMN/20160719/testcase"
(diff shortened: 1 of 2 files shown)


yesamer:
@dmn-tck/contributors Please review

SimonRinguette (approved):
Works for me

SimonRinguette (changes requested):
Second thoughts here.

TC `009_d` and `009_b` should be commented out for now for a larger discussion on how a failure in a DT row should affect the result of the complete table.

opatrascoiu:
I think there are few more DTs with errors:

`007_a`, because the decision `booleanValueTest ` compares string with boolean in rule 1
`007_d`, because the decision `listValueTest ` compares list with boolean in rule 2
`009_a`, because the decision `expressionTest_a ` compares string with boolean in rule 1
`009_b`, because in decision `expressionTest_b ` the condition is not a boolean expression in rule 1
`009_d`, because the decision `expressionTest_d ` compares number with boolean in rule 1 - the condition is [345, true, 456] after evaluation

I think we should comment them all out until we clarify the issue in DMN RTF.

@yesamer @SimonRinguette  Any comments?

opatrascoiu:
The RTF issue is here https://issues.omg.org/browse/INBOX-2519

PR #746, opened 2026-06-30
Update decisionservice result evaluation logic to account for implicit conversion

AthiraHari77:
This PR intends to address the issue while evaluating decisionservice results when implicit conversion is present.

 When a decision body returns a date, the engine is permitted to implicitly convert it to date and time if the surrounding context (e.g., the Decision Service's declared output type) requires it.
Only the date → date and time implicit conversion changes the declared output type of the Decision Service (from date to date and time), which forces the TCK runner to parse the expected XML value differently. All other DMN implicit conversions (singleton list unwrapping, number coercion, etc.)  leave the declared type unchanged, so the existing parsing logic already handles them correctly.


yesamer:
@gitgabrio FYI

